PR #1, opened 2014-10-30
lift goes up to 11!

ivanov:
I got a good laugh out of "lift should go up to 11" at the hackathon in
SF, so decided to submit a PR with the feature

### DIFF
--- a/src/Native/Function.js
+++ b/src/Native/Function.js
@@ -73,6 +73,38 @@ function F9(fun) {
   wrapper.func = fun;
   return wrapper;
 }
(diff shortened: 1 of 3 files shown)


salutis:
Do we even need `Signal.mapX` function? Wouldn't `map`, `(<~)`, and  `(~)` be enough?

evancz:
Having talked to lots of beginners, many people struggle with `(<~)` and `(~)` for a while at first. I think it's valuable as an education tool *at least*, but I personally use `mapN` all the time.

PR #8, opened 2014-11-19
Fixed a broken link to another library page.

WillSewell:


evancz:
This should be fixed by [this](https://github.com/elm-lang/core/blob/master/src/Graphics/Collage.elm#L137), thanks for the fix though!

PR #23, opened 2014-11-29
Add withCredentials to XHR object

df3n5:
- This enables us to use with cookies across domains
 - Maybe we want to set this optionally, but I see no downside in having it enabled all the time.


evancz:
This kind of thing now lives on evancz/elm-http

PR #46, opened 2014-12-13
Change foldr's recursive call from foldl to foldr

pseudonom:
I believe this fixes https://github.com/elm-lang/core/issues/45.


evancz:
Cool, thank you!

PR #52, opened 2014-12-14
add Dict.size and Set.size

jameshfisher:


evancz:
Makes sense, thank you!

PR #63, opened 2014-12-23
Add Graphics.Element.scrollContainer

avh4:
This is a proposal to add the capability to create scrolling regions in UIs built using the `Graphics.Element` API.  

With this change, here is how you would make a 200x300 region containing a vertically-scrolling block of text:

```elm
main = plainText "my text..." |> width 200 |> scrollContainer 200 300 
```

`Graphics.Element.scrollContainer` as proposed here can also be used to make viewports that scroll horizontally or that scroll both vertically and horizontally.

As far as I know, there is currently no way to achieve this type of effect with the core API.

### DIFF
--- a/src/Graphics/Element.elm
+++ b/src/Graphics/Element.elm
@@ -188,6 +188,7 @@ newElement w h e =
 type ElementPrim
     = Image ImageStyle Int Int String
     | Container Position Element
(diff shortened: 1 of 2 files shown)


evancz:
I guess we'd just do it in HTML these days.

PR #72, opened 2014-12-30
Make scale documentation explicit about the scaling

arsatiki:
The scaling documentation was somewhat unclear about which aspect will get scaled.
Now it is specified explicitly.


evancz:
Cool, thank you!

Issue #75, opened 2015-01-01
Time.timestamp does not seem to satisfy the semantics implied in documentation


jvoigtlaender:
See https://groups.google.com/d/msg/elm-discuss/AWx51F_WzUE/iKBdMM94JJoJ

jvoigtlaender:
Also, I think the current implementation makes it possible that in:

```
test : Signal (Time,Time)
test = timestamp (every second)
```
pairs occur as value on the signal whose first and second component differ, contrary to what one would expect from the documentations of `Time.every` and `Time.timestamp`.

See this thread: https://groups.google.com/d/msg/elm-discuss/KzevSWc0gfU/UO5-BQozwEUJ

jvoigtlaender:
The PR #76 also fixes the above issue concerning `timestamp (every second)` as well as a similar issue concerning interaction between `timestamp` and `fps`/`fpsWhen`.

PR #76, opened 2015-01-02
Fix timestamp function and its interaction with every/fps/fpsWhen

jvoigtlaender:
making sure that events occurring in the same update cycle all get the
same timestamp;

also, making sure that timestamp (every t) will not carry events with different pair components;

this fixes issue #75

### DIFF
--- a/src/Native/Signal.js
+++ b/src/Native/Signal.js
@@ -118,10 +118,18 @@ Elm.Native.Signal.make = function(localRuntime) {
     input.kids.push(this);
   }
 
-  function timestamp(a) {
-    function update() { return Utils.Tuple2(localRuntime.timer.now(), a.value); }
-    return new LiftN(update, [a]);
+  function Timestamp(input) {
+    this.id = Utils.guid();
+    this.value = Utils.Tuple2(localRuntime.timer.now(), input.value);
+    this.kids = [];
+    this.recv = function(timestep, changed, parentID) {
+      if (changed) { this.value = Utils.Tuple2(timestep, input.value); }
+      broadcastToKids(this, timestep, changed);
+    };
+    input.kids.push(this);
   }
+
+  function timestamp(input) { return new Timestamp(input); }
 
   function SampleOn(s1,s2) {
     this.id = Utils.guid();
--- a/src/Native/Signal.js
+++ b/src/Native/Signal.js
@@ -120,8 +120,7 @@ Elm.Native.Signal.make = function(localRuntime) {
 
   function Timestamp(input) {
     this.id = Utils.guid();
-    var programStart = localRuntime.timer.now(); // it would be better to set programStart once as a global variable of the runtime [jvoigtlaender]
-    this.value = Utils.Tuple2(programStart, input.value);
+    this.value = Utils.Tuple2(localRuntime.timer.programStart, input.value);
     this.kids = [];
     this.recv = function(timestep, changed, parentID) {
       if (changed) { this.value = Utils.Tuple2(timestep, input.value); }
--- a/src/Native/Signal.js
+++ b/src/Native/Signal.js
@@ -120,7 +120,8 @@ Elm.Native.Signal.make = function(localRuntime) {
 
   function Timestamp(input) {
     this.id = Utils.guid();
-    this.value = Utils.Tuple2(localRuntime.timer.now(), input.value);
+    var programStart = localRuntime.timer.now(); // it would be better to set programStart once as a global variable of the runtime [jvoigtlaender]
+    this.value = Utils.Tuple2(programStart, input.value);
     this.kids = [];
     this.recv = function(timestep, changed, parentID) {
       if (changed) { this.value = Utils.Tuple2(timestep, input.value); }
--- a/src/Native/Signal.js
+++ b/src/Native/Signal.js
@@ -123,13 +123,17 @@ Elm.Native.Signal.make = function(localRuntime) {
     this.value = Utils.Tuple2(localRuntime.timer.programStart, input.value);
     this.kids = [];
     this.recv = function(timestep, changed, parentID) {
-      if (changed) { this.value = Utils.Tuple2(timestep, input.value); }
+      if (changed) {
+        this.value = Utils.Tuple2(timestep, input.value);
+      }
       broadcastToKids(this, timestep, changed);
     };
     input.kids.push(this);
   }
 
-  function timestamp(input) { return new Timestamp(input); }
+  function timestamp(input) {
+    return new Timestamp(input);
+  }
 
   function SampleOn(s1,s2) {
     this.id = Utils.guid();
--- a/src/Native/Time.js
+++ b/src/Native/Time.js
@@ -20,9 +20,8 @@ Elm.Native.Time.make = function(elm) {
 
   function fpsWhen(desiredFPS, isOn) {
     var msPerFrame = 1000 / desiredFPS;
-    var prev, curr, diff, wasOn = true;
-    var zero = true;
-    var ticker = NS.input(zero);
+    var wasOn = true;
+    var ticker = NS.input(true);
     var timeoutID = 0;
     function f(isOn, t) {
       if (isOn) {
@@ -34,10 +33,10 @@ Elm.Native.Time.make = function(elm) {
       return t;
     }
     function g(event, old) {
-      prev = snd(old);
-      curr = fst(event);
-      zero = snd(event);
-      diff = zero ? 0 : curr - prev;
+      var prev = snd(old);
+      var curr = fst(event);
+      var zero = snd(event);
+      var diff = zero ? 0 : curr - prev;
       return Utils.Tuple2(diff, curr);
     }
     var deltas = A2( Signal.map, fst, A3( Signal.foldp, F2(g), Utils.Tuple2(0, elm.timer.programStart), NS.timestamp(ticker) ) );
--- a/src/Native/Time.js
+++ b/src/Native/Time.js
@@ -10,20 +10,12 @@ Elm.Native.Time.make = function(elm) {
   var Maybe = Elm.Maybe.make(elm);
   var Utils = Elm.Native.Utils.make(elm);
 
-  function fst(pair) {
-      return pair._0;
-  }
-
-  function snd(pair) {
-      return pair._1;
-  }
-
   function fpsWhen(desiredFPS, isOn) {
     var msPerFrame = 1000 / desiredFPS;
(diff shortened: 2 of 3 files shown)


evancz:
I agree that it should be a global thing. Do you mind making that change?

Have you tested that this works in the cases you mentioned? I am messing with elm-reactor at the moment, and I am not *certain* that the timestamp argument is getting passed in properly in all cases.

I was playing around with initializing the signal graph separately from specifying it. So instead of calling `map` and actually creating a node at that moment, it is just some data to be dealt with later. This is more in the spirit of solving the recursive-values / looping problem described [here](https://gist.github.com/evancz/bdb3459af071d69bc4e8#signal-loops), but the point is that in that world, having a "global" start time that gets threaded through the graph initialization makes total sense.

jvoigtlaender:
Yes, I can introduce the global `programStart`. I'll add that change to this PR.

Yes, I've tested the PR. Specifically, I've tested that:
* `Signal.map2 (\(a,_) (b,_) -> (a==b)) (timestamp (Signal.map f s)) (timestamp (Signal.map g s))` only produces `True`s
* `uncurry (==) <~ timestamp (every second)` only produces `True`s
* `(\(_,d1,d2) -> d1 == d2) <~ foldp (\(ns,d) (os,_,_) -> (ns,ns-os,d)) (0,0,0) (timestamp (fps 1))` only produces `True`s (except in the first event, where this is simply not possible, because the first delta computed in the `foldp` will compare the current time against initial value 0 instead of against a reasonable previous time)

jvoigtlaender:
Okay, `programStart` is now set globally.

src/Native/Time.js
```diff
@@ -10,39 +10,46 @@ Elm.Native.Time.make = function(elm) {
...
       wasOn = isOn;
       return t;
     }
-    return A3( Signal.map2, F2(f), isOn, ticker );
+    function g(event, old) {
+      prev = snd(old);
+      curr = fst(event);
+      zero = snd(event);
+      diff = zero ? 0 : curr - prev;
+      return Utils.Tuple2(diff, curr);
```
laszlopandy: These variables should all be local to this function because they are not used anywhere else now.

src/Native/Time.js
```diff
@@ -10,39 +10,46 @@ Elm.Native.Time.make = function(elm) {
...
   }
 
   function every(t) {
-    var clock = NS.input(elm.timer.now());
+    var ticker = NS.input(Utils.Tuple0);
     function tellTime() {
-        elm.notify(clock.id, elm.timer.now());
+        elm.notify(ticker.id, Utils.Tuple0);
     }
+    var clock = A2( Signal.map, fst, NS.timestamp(ticker) );
```
laszlopandy: I think there should be a function `Time.timeOf : Signal a -> Signal Time`, which samples the current time of each signal update. It seems like there are many cases where you call `timestamp` but actually only want the first argument.

src/Native/Time.js
jvoigtlaender: Yes, I had not made them local purely because I wanted to create as small as possible a conceptual diff against the old version of this function, where these variables were declared at the top.

But yes, I'll refactor now as you suggest.

src/Native/Time.js
jvoigtlaender: The question is, will this be a function defined on the Elm side, or something added as a native signal on the JS side.

laszlopandy: I would write `Time.timeOf` in native and then define `Time.timestamp` in terms of it.
```
timestamp sig = Signal.map (,) (timeOf sig) sig
```

jvoigtlaender: So this would be a change in the standard library. Probably something that should be considered independently of this PR.

laszlopandy: Good point. Let's merge this one first.

laszlopandy:
Please merge with master, so you get the new run-test.sh script. Otherwise this branch will keep failing on Travis.

jvoigtlaender:
Merged, and Travis build pass looks fine.

src/Native/Time.js
```diff
@@ -10,39 +10,42 @@ Elm.Native.Time.make = function(elm) {
...
       return t;
     }
-    return A3( Signal.map2, F2(f), isOn, ticker );
+    function g(event, old) {
+      var curr = fst(event);
+      return Utils.Tuple2(snd(event) ? 0 : curr - snd(old), curr);
+    }
+    var deltas = A2( Signal.map, fst, A3( Signal.foldp, F2(g), Utils.Tuple2(0, elm.timer.programStart), NS.timestamp(ticker) ) );
+    return A3( Signal.map2, F2(f), isOn, deltas );
   }
```
laszlopandy: These last few lines are really hard to read for me.
 * `f` and `g` should have proper names. (I thought of `startStopTimer` and `calcDelta`, but you can choose).
 * The tuple coming out of `NS.timestamp` and the tuple from `g` are not the same, but I thought they were. Since this will not be exposed to Elm, you can use an object, it doesn't have to be an Elm type: `{ delta: 0, timestamp: elm.timer.programStart }`

jvoigtlaender:
Bumping this, now that the dust has settled around the 0.14.1 release.

evancz:
Do we have any certainty at all that this will work with Elm Reactor? I did a review just now, and it's not clear to me at all. If it does not work, are you responsible for making the changes there?

I have some style concerns that I'll add inline.

src/Native/Signal.js
```diff
@@ -118,11 +118,19 @@ Elm.Native.Signal.make = function(localRuntime) {
...
 
-  function timestamp(a) {
-    function update() { return Utils.Tuple2(localRuntime.timer.now(), a.value); }
-    return new LiftN(update, [a]);
+  function Timestamp(input) {
+    this.id = Utils.guid();
+    this.value = Utils.Tuple2(localRuntime.timer.programStart, input.value);
+    this.kids = [];
+    this.recv = function(timestep, changed, parentID) {
+      if (changed) { this.value = Utils.Tuple2(timestep, input.value); }
```
evancz: Use more modern style. I did a bunch of messy things when I was writing this code.

```javascript
if (changed) {
  this.value = Utils.Tuple2(timestep, input.value);
}
```

src/Native/Signal.js
```diff
@@ -118,11 +118,19 @@ Elm.Native.Signal.make = function(localRuntime) {
...
+      broadcastToKids(this, timestep, changed);
+    };
+    input.kids.push(this);
   }
 
+  function timestamp(input) { return new Timestamp(input); }
```
evancz:
```javascript
function timestamp(input) {
  return new Timestamp(input);
}
```

src/Native/Time.js
```diff
@@ -12,37 +12,36 @@ Elm.Native.Time.make = function(elm) {
 
   function fpsWhen(desiredFPS, isOn) {
```
evancz: I find the new version of this function very very confusing. At least it should not have 170+ character lines, but the ideal version might actually move `wasOn` and `timeoutID` into a signal so that it all gets picked up by Elm Reactor?

The same style stuff with `{}` applies here as well with the function and object. Prefer to bring things down a line.

jvoigtlaender:
Re: "Do we have any certainty at all that this will work with Elm Reactor? I did a review just now, and it's not clear to me at all. If it does not work, are you responsible for making the changes there?"

Do you have specific concerns about what might potentially break? I did think about how this might interact with the reactor while writing the PR. Specifically, I was considering hot-swapping and convinced myself that that will be okay (basing that judgement on an analysis of what `hotSwap` in `Runtime.js` does). Also, I had taken into account in my first version of the PR how the timer indirection stuff worked prior to https://github.com/elm-lang/core/commit/d4a6d9350d5817912a88e7902cfe5f09413fbae1 (with `inducedDelay` and all), and then when that commit came out, thought about how that would affect my PR and adapted it accordingly. What guarantees beyond those considerations can I give? Will I feel responsible for making changes in the reactor if they become necessary because of what I introduced here? To the best of my abilities and understanding, yes.

Aside: the PR is divided into small-scale commits. Some of these can easily be identified as mere refactorings which certainly have no potential to break anything. I could talk you through the remaining commits in order, to explain why each of them is okay (not changing semantics in an adverse way).

src/Native/Signal.js
jvoigtlaender: Sure, I can do these style changes. I had refrained from doing any of that, both to make the syntactic diff as small as possible and since you somewhere expressed that contributors should as much as possible preserve syntactic style they encounter in the files they edit.

src/Native/Time.js
jvoigtlaender: Do you find the old version of this function less confusing? That would be https://github.com/elm-lang/core/blob/master/src/Native/Time.js#L13-39. It has the same basic structure with the non-exposed signals (though nothin for Elm Reactor to pick up).

The crucial commit concerning this function is this one: https://github.com/jvoigtlaender/core/commit/274cd7eb9a05c1b32dab563a871c557097bc3392. All subsequent commits are just renamings or moving around of variables, prompted by @laszlopandy's comments.

Should I talk you through the mentioned commit? I'd say it very much *simplifies* this function compared to how it was previously.

evancz:
Thanks for the fix! I did some style changes that were needed in the files, but I'll trust that things work in a good way.

One simple way to see if things work well with Elm Reactor is to use Elm Reactor with the new code. If you run through some programs that use `fps` and pause, rewind, and swap, that's a pretty good indicator.

jvoigtlaender:
Cool that you merged this.

Also, I agree that the commit https://github.com/elm-lang/core/commit/79cd80e68d0fe448713b4aefdae6be80ec23675a made the working of the `fpsWhen` function clearer. In fact, given that clearer form I now think a further change is warranted, in this line: https://github.com/elm-lang/core/blob/master/src/Native/Time.js#L37. Namely, it would be better to have `wasOn = isOn.value` there. The reason is that if the initial value of the passed `isOn` signal is `false` (otherwise the proposed new line is anyway equivalent to the current line), then the current implementation unnecessarily does a `clearTimeout(0)` at signal initialization time, whereas the proposed change would prevent that.

jvoigtlaender:
Re: "One simple way to see if things work well with Elm Reactor is to use Elm Reactor with the new code", am I correct that I cannot do this by testing with the `elm-reactor` executable that was installed by the Elm-0.14.1 installer? Because its debugger mode will run with some "compiled-in" version of the standard library, that is, it will still run the implementations of `timestamp`, `fps` etc. that were current when that `elm-reactor` executable was built?

PR #78, opened 2015-01-05
run-test.sh script

laszlopandy:
Makes it easy to run the tests locally.


laszlopandy:
This will require a configuration change on Travis CI, so that it calls the run-test.sh script instead of manually building and running everything.

PR #79, opened 2015-01-05
Fix disappearing HTML elements, Fixes #862

### LINKED ISSUE #862: NaN argument to the mod operator (%) causes unbounded recursion
Running this simple program causes a JavaScript exception:
```
module Main exposing (..)

import Html exposing (Html, text, div)


main : Html a
main =
    let
        nanFloat : Float
        nanFloat = sqrt -1
        
        nanInt : Int
        nanInt = ceiling nanFloat
        
        recursionHere : Int
        recursionHere = 6 % nanInt

    in
        div [] 
            [ div [] [ text (toString recursionHere) ]
            ]
```

Resulting error:
```
Uncaught RangeError: Maximum call stack size exceeded
    at mod (VM98 javascript:1117)
    at mod (VM98 javascript:1124)
    at mod (VM98 javascript:1124)
    at mod (VM98 javascript:1124)
    ...
```

On Ellie: https://ellie-app.com/36bpc9T7v5ba1/5

The implementation should check for NaN in the arguments, and likely return NaN. In addition, it should not be possible to get an Int containing a value like Nan or Infinity. 

Browser: Chrome 58.0.3029.96 (64-bit)
OS: macOS 10.12.4
Elm: 0.18.0

jwmerrill:
See discussion on #862, especially https://github.com/elm-lang/elm-compiler/issues/862#issuecomment-68155239.

This merits careful review because I don't know why the `kid.getContext` check that I'm removing was there in the first place. @etaque reports that this change fixes the issue.

### DIFF
--- a/src/Native/Graphics/Collage.js
+++ b/src/Native/Graphics/Collage.js
@@ -358,7 +358,7 @@ Elm.Native.Graphics.Collage.make = function(localRuntime) {
             ++i;
             if (!kid) {
                 div.appendChild(node);
-            } else if (kid.getContext) {
+            } else {
                 div.insertBefore(node, kid);
             }
         }


evancz:
Reading the code and knowing what my intent was with that code, I am pretty sure this is just a bug. I have not touched this code in ages, so it is somewhat mysterious if this arose with 0.14. In any case, this looks good to me!

Thank you so much for figuring this out, I really really appreciate it!

PR #85, opened 2015-01-09
Add more Basics tests

deadfoxygrandpa:
I went through and added at least one test for every function defined in Basics. I used the example cases written in the documentation where available. All of these tests currently pass.

Some of these are pretty trivial tests, but I was thinking why not just add one anyway?

If this is something that you think is acceptable or worth doing, I can continue adding tests like this to other files in core.

### DIFF
--- a/tests/Test/Basics.elm
+++ b/tests/Test/Basics.elm
@@ -50,10 +50,34 @@ tests =
             , test "atan2" <| assert (abs (1.55 - atan2 36 0.65) < 0.01)
             , test "pi" <| assert (abs (3.14 - pi) < 0.01)
             ]
-
+        basicMathTests = suite "Basic Math Tests"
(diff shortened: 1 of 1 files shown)


evancz:
Awesome, this looks great! I think it'd be great to have more comprehensive test coverage on other modules, but you should coordinate with @laszlopandy because he is working on some changes to make a bunch of functions total. I guess there won't be merge conflicts with the kinds of changes going on?

evancz:
I think the `Json.Decode` module is a great candidate for tests at the moment. I think @laszlopandy tried to think of some tests for `Result` and couldn't think of anything super meaningful. In any case, yeah, this is awesome!

laszlopandy:
Wow awesome work. Don't worry about merge conflicts, I can fix them and I'm not really touching Basics.

PR #86, opened 2015-01-09
Core 2.0 - remove runtime exceptions

laszlopandy:
Core 2.0:
 * Remove runtime exceptions:
   * Changed type `List.head : List a -> Maybe a`
   * Deleted `List.tail` in favour of `List.drop 1`
   * Deleted `List.foldl1`, `List.foldr1`, `List.scanl1`
   * Added `List.uncons`
   * Changed `List.maximum` and `List.minimum` to take default values
 * Added `Graphics.Model`
   * Including `Debug` will not include `Collage` or `Element`
   * Public types are re-exported from the same module as before.
   * Private types which were exposed before are now hidden: `BasicForm`, `LineCap`, `LineJoin`, `ShapeStyle`, `FillStyle`, `ElementPrim`, `ImageStyle`, `Three`. 
   * Since no one should have been using the private types, I don't expect any breaking changes from the `Graphics.Model` refactor.
 * Fixed compiling while doing breaking changes in core: `run-test.sh` will now delete the elm-package version of core and symlink it to the local git version.

### DIFF
--- a/src/Graphics/Element.elm
+++ b/src/Graphics/Element.elm
@@ -43,7 +43,6 @@ import List as List
 import Color (..)
 import Maybe ( Maybe(..) )
 
-
 type alias Properties = {
   id      : Int,
   width   : Int,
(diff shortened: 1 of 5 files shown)


evancz:
Is it possible to do this in three batches? I think there are problems in the Graphics part, but not the others.

I believe some of the functions you hid are actually public. See [this page](http://package.elm-lang.org/packages/elm-lang/core/1.0.0/Graphics-Collage#LineStyle) for LineStyle, LineCap, LineStyle, and FillStyle.

evancz:
Also, don't bump the version number. I prefer to do it with `elm-package bump` right before a release. This way you always know your reference point when you are changing stuff.

deadfoxygrandpa:
Basics has a possible runtime exception:
```js
  function mod(a, b) {
        if (b === 0) {
            throw new Error("Cannot perform mod 0. Division by zero error.");
        }
        var r = a % b;
        var m = a === 0 ? 0 : (b > 0 ? (a >= 0 ? r : r+b) : -mod(-a,-b));

        return m === b ? 0 : m;
  }
```
Could this return NaN instead?

evancz:
@deadfoxygrandpa, I don't really know what is the right thing here, but an error sort of feels more correct. I'd be interested if there is a better/faster implementation in general though.

PR #90, opened 2015-01-10
List does not throw runtime exceptions

laszlopandy:
* Changed type `List.head : List a -> Maybe a`
   * Deleted `List.tail` in favour of `List.drop 1`
   * Deleted `List.foldl1`, `List.foldr1`, `List.scanl1`
   * Added `List.uncons`
   * Changed `List.maximum` and `List.minimum` to take default values

### DIFF
--- a/src/Graphics/Element.elm
+++ b/src/Graphics/Element.elm
@@ -41,8 +41,7 @@ import Basics (..)
 import Native.Graphics.Element
 import List as List
 import Color (..)
-import Maybe ( Maybe(..) )
-
+import Maybe ( Maybe(..), withDefault )
 
 type alias Properties = {
   id      : Int,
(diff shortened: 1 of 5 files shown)


src/Graphics/Element.elm
```diff
@@ -274,16 +273,18 @@ flow : Direction -> List Element -> Element
...
-    DLeft  -> newFlow (List.sum ws) (List.maximum hs)
-    DRight -> newFlow (List.sum ws) (List.maximum hs)
-    DIn    -> newFlow (List.maximum ws) (List.maximum hs)
-    DOut   -> newFlow (List.maximum ws) (List.maximum hs)
+    DUp    -> newFlow maxW (List.sum hs)
+    DDown  -> newFlow maxW (List.sum hs)
+    DLeft  -> newFlow (List.sum ws) maxH
+    DRight -> newFlow (List.sum ws) maxH
+    DIn    -> newFlow maxW maxH
+    DOut   -> newFlow maxW maxH
```
evancz: This code does extra traversals compared to the old way.

src/Signal.elm
```diff
@@ -148,7 +149,9 @@ update wins, just like with `merge`.
 -}
 mergeMany : List (Signal a) -> Signal a
 mergeMany signals =
-    List.foldr1 merge signals
+    case signals of
+        s :: rest -> List.foldr merge s rest
```
evancz: I think this is buggy. `s` is the leftmost thing, but it is used as the rightmost. I think `foldr` needs the last element.

src/List.elm
```diff
@@ -244,18 +238,20 @@ product numbers =
...
 {-| Find the minimum element in a non-empty list.
 
-    minimum [3,2,1] == 1
+    minimum 10 [3,2,1] == Just 1
+    minimum 10 []      == 10
 -}
-minimum : List comparable -> comparable
-minimum = foldl1 min
+minimum : comparable -> List comparable -> comparable
+minimum def list = foldl min def list
```
evancz: I think it is often hard to pick a minimum value. I'm imagining people writing `minimum (maximum 0 xs) xs`.

Why did you go for this instead of returning a maybe?

evancz: I am having a hard time evaluating this change without any examples as data.

laszlopandy: Yeah I just realized today that there is a difference between a default
value (which will be ignored if the list is non-empty) and a base value for
foldl.

So I will change it back to maybe.

src/Signal.elm
laszlopandy: I did a reverse and foldl. I'm pretty sure this is equivalent to foldr, but please check if over again.

PR #96, opened 2015-01-14
Add `import Text` to fix runtime error.

### LINKED ISSUE #93: Graphics.Input.hoverable causes runtime error
Minimum example:

````elm
import Graphics.Element as E
import Graphics.Input as Input
import Signal as S

chan = S.channel False

main = E.spacer 100 200 |> Input.hoverable (S.send chan)
````

Produces:
````
Cannot read property 'make' of undefined

Open the developer console for more details.
````

It seems to error on `var Text = Elm.Native.Text.make(localRuntime);` because `Elm.Native.Text` is undefined. Seems like a dependency ordering issue.

mgold:
Fixes #93.

I threw in a comment in hopes of guarding against a regression.

### DIFF
--- a/src/Graphics/Input.elm
+++ b/src/Graphics/Input.elm
@@ -17,6 +17,7 @@ To learn about text fields, see the
 import Signal
 import Graphics.Element (Element)
 import Native.Graphics.Input
+import Text -- used internally
 
 
 {-| Create a standard button. The following example begins making a basic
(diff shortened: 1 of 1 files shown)


evancz:
Does it fix things to import `Native.Text` as well? I don't think it's a *huge* deal, but might be something to consider.

mgold:
That also works. Updated. Native communicates that it's internal without a comment, so that sounds good to me.

evancz:
Cool, thank you!

PR #97, opened 2015-01-15
added explicit export list to Array.elm

jvoigtlaender:
meant to address part of https://github.com/elm-lang/elm-lang.org/issues/173


evancz:
Great, thank you! According to `elm-package diff` we are no longer exporting the `Array` constructor, which definitely should be hidden.

So looks good to me!

PR #100, opened 2015-01-15
added explicit export list to Char.elm

jvoigtlaender:
meant to address part of https://github.com/elm-lang/elm-lang.org/issues/173

### DIFF
--- a/src/Char.elm
+++ b/src/Char.elm
@@ -1,5 +1,7 @@
-
-module Char where
+module Char
+    ( isUpper, isLower, isDigit, isOctDigit, isHexDigit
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
When exporting a `type alias`, like `KeyCode` here, will the importing module "know" that `KeyCode` is  `Int`, or will this be opaque from the outside?

evancz:
As things stand, the outside module will know. I think this is the right way to go in the sense that you can do hiding with `type KeyCode = KeyCode Int` and we'll eventually do the same unboxing trick that OCaml and Haskell do (the `newtype` trick in Haskell)

PR #112, opened 2015-01-15
added explicit export list to Text.elm

jvoigtlaender:
meant to address part of elm-lang/elm-lang.org#173


evancz:
Cool, I just made the change such that `markdown` can be removed from the public API as well. I'll do that real quick.

PR #114, opened 2015-01-15
added explicit export list to Touch.elm

jvoigtlaender:
meant to address part of elm-lang/elm-lang.org#173


evancz:
I think this removes the `Touch` constructor from the public API, which is more in line with the design guidelines about how to expose record types. Cool :)

PR #121, opened 2015-01-20
Bump Elm-Test dependency version to 1.0.3.

laszlopandy:


evancz:
Thank you @laszlopandy and @deadfoxygrandpa!

PR #122, opened 2015-01-20
Removed mention of (?) operator in documentation

TheSeamau5:
The documentation for `Array.get` says that the  `?` operator from the `Maybe` library makes `Array.get` easier to use when there isn't any such operator in the `Maybe` library


evancz:
Thanks, that got renamed to `withDefault` to follow the guidelines about avoiding infix ops :)

PR #123, opened 2015-01-21
Fix Regex.split, and add a couple tests

avh4:
See the added `Test/Regex.elm` for the example cases.  `Regex.split` seems to have been completely broken at some point.


evancz:
Awesome, thank you!

PR #124, opened 2015-01-21
Replace "elm" by "localRuntime" in JS files

jvoigtlaender:


jvoigtlaender:
Like this?

evancz:
Yeah, added some comments on the merge though.

Some modules do this trick where `values` get attached to `Elm` instead of just `localRuntime`. I generally think this is not a great idea, and I'm not sure of the benefits really. I'd be curious if there's some decisive way to know the right thing on that.

jvoigtlaender:
Concerning the latter point, I have no idea when to prefer attaching values to `Elm`. Is the default attaching them to `localRuntime`, and whenever the alternative is better in your judgement, you will do it yourself?

evancz:
When a module needs to call `setTimeout` or `notify` or attach event listeners, then it definitely needs to use `localRuntime`. You don't want to be able to accidentally `notify` different Elm modules in crazy ways.

When the module is stateless, it *can* be attached to `Elm` in theory. That should work for stuff like Array and List, but I have been bitten before by a module that used a `guid` in some indirect and surprising way that led to bugs. So it works, but it can lead to crazy bugs. Overall, it's not clear that this is a worthwhile thing to ever do.

PR #126, opened 2015-01-21
Remove incorrect upcase from Char.toCode (fix #50)

### LINKED ISSUE #50: `Char.toCode` gives the same result for lowercase and uppercase letters
The offending line is line 20 of [/src/Native/Char.js](https://github.com/elm-lang/core/blob/master/src/Native/Char.js)
>  toCode : function(c) { return c.toUpperCase().charCodeAt(0); },

jonathanhefner:
This is a straight-forward fix for #50.  Simply removed the erroneous `toUpperCase` call.

Test coverage is included.


evancz:
Awesome, thank you for the fix!

PR #128, opened 2015-01-22
Signal.constant -> Native.Signal.input in Runtime.js

jvoigtlaender:
For consistency with https://github.com/elm-lang/core/pull/125.


jvoigtlaender:
I take this back.

PR #129, opened 2015-01-22
Add more Char test coverage

jonathanhefner:
Added more test coverage of Char functions.  I would like these tests to be more DRY, but I was on the fence about the best (or most idiomatic) way to do that.

### DIFF
--- a/tests/Test/Char.elm
+++ b/tests/Test/Char.elm
@@ -17,4 +17,25 @@ tests =
           , test "toCode 'z'" <| assertEqual 122 <| Char.toCode 'z'
           , test "toCode 'A'" <| assertEqual 65 <| Char.toCode 'A'
           , test "toCode 'Z'" <| assertEqual 90 <| Char.toCode 'Z'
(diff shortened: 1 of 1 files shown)


evancz:
Thank you! I know there's a decent amount of interest in something like QuickCheck (which is essentially fuzz testing in which you say "this relationship should hold" and then throw a bunch of randomly generated values at it). Would that make this feel nicer?

jonathanhefner:
Fuzz testing is pretty cool, but because the relationship here is piecemeal and somewhat arbitrary (probably best defined by a small lookup table), I don't know if it would help.

I'm considering creating a package of test helper functions (on top of ElmTest) that might make such things easier.

evancz:
Cool, improving testing in general is a goal for us. [This thread](https://groups.google.com/forum/#!topic/elm-dev/-oC1b4KuELA) gives some context about how we can make progress, so I'd be curious if that aligns with what you want to do. My worry is committing too much to any particular testing strategy, so I'm curious to see what is needed by different approaches and how to support that in a minimal way.

PR #130, opened 2015-01-22
Port simple Char functions from JavaScript to Elm

jonathanhefner:
I'm not sure if this is wanted or not, but I figure it doesn't hurt to offer.

I ported the Native.Char functions that do not use the JavaScript standard lib to pure Elm.  My reasons for doing so:

  * Inspecting / understanding the implementation is easier.
  * Any tooling (static analysis, optimization, test coverage) that Elm has or will have can now target these.
  * If Elm itself is ever ported to another runtime, these will now be freebies.

### DIFF
--- a/src/Char.elm
+++ b/src/Char.elm
@@ -16,25 +16,31 @@ module Char
 
 import Native.Char
 
+isBetween : Char -> Char -> Char -> Bool
+isBetween low high char = 
+    let code = toCode char
+    in  (code >= toCode low) && (code <= toCode high)
+
 {-| True for upper case letters. -}
 isUpper : Char -> Bool
-isUpper = Native.Char.isUpper
+isUpper = isBetween 'A' 'Z'
 
 {-| True for lower case letters. -}
 isLower : Char -> Bool
-isLower = Native.Char.isLower
+isLower = isBetween 'a' 'z'
 
 {-| True for ASCII digits `[0-9]`. -}
 isDigit : Char -> Bool
-isDigit = Native.Char.isDigit
+isDigit = isBetween '0' '9'
 
 {-| True for ASCII octal digits `[0-7]`. -}
 isOctDigit : Char -> Bool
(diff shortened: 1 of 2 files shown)


src/Char.elm
```diff
@@ -16,25 +16,31 @@ module Char
...
 {-| True for ASCII octal digits `[0-7]`. -}
 isOctDigit : Char -> Bool
-isOctDigit = Native.Char.isOctDigit
+isOctDigit = isBetween '0' '7'
 
 {-| True for ASCII hexadecimal digits `[0-9a-fA-F]`. -}
 isHexDigit : Char -> Bool
-isHexDigit = Native.Char.isHexDigit
+isHexDigit char = 
+    isDigit char || isBetween 'a' 'f' char || isBetween 'A' 'Z' char
```
evancz: Should the Z be an F?

jonathanhefner: Absolutely.  I ran the tests... on master (FML).  Fixing now.

src/Char.elm
```diff
@@ -16,25 +16,31 @@ module Char
 
 import Native.Char
 
+isBetween : Char -> Char -> Char -> Bool
+isBetween low high char = 
+    let code = toCode char
+    in  (code >= toCode low) && (code <= toCode high)
```
jonathanhefner: The Basics module isn't auto-imported here, so (&&), (>=), and (<=) are undefined.

Is it safe to `import Basics (..)`?  I assume the reason to not do so is to avoid circular dependencies...?  If that's the case, is it uncouth to use `Native.Basics.and`, etc?

evancz: I'd just import the functions in question: `import Basics ((&&), (>=), (<=))`

It is fine as long as the tests run. If there's a circular dependency, it'll just endlessly initialize modules until some piece of technology tells you to stop.

evancz:
Looks great, thank you!

PR #131, opened 2015-01-23
Correcting the type annotation for Json-Encode.object3's docs.

krisajenkins:
I think this was a copy & paste error from the docs for object2.


evancz:
Yep, thank you!

PR #133, opened 2015-01-23
Cherry-pick for 1.1.1 release

kasbah:
So far, I cherry-picked 723aab74e2a1c for the toCode fix and 633a3acb8db529f2 to try and make the tests work.

Other candidates might be among [these](https://gist.github.com/kasbah/b4e880f2df87b72d859b). Let me know your suggestions if you want me to add them to this PR.

I couldn't `run-test.sh` because I got the error below. Any ideas? Also, should I do any other testing?
```
Error in package elm-lang/core 1.1.0 in module List:

Parse error:
Overlapping definitions for infix operators: ::
```

PR #134, opened 2015-01-23
Cherry-pick for 1.1.1 attempt2

kasbah:
Proposing this as an alternative to #133 as there were a lot less commits mentioning the word "fix" than I initially thought (got my git revision range wrong). So here they are cherry picked. Let's see how the CI likes it.

Happy to prune out any you don't like or add any as well, or go back to doing it more manually with #133.

### DIFF
--- a/src/Graphics/Input.elm
+++ b/src/Graphics/Input.elm
@@ -17,6 +17,7 @@ To learn about text fields, see the
 import Signal
 import Graphics.Element (Element)
 import Native.Graphics.Input
+import Text -- used internally
 
 
 {-| Create a standard button. The following example begins making a basic
--- a/src/Basics.elm
+++ b/src/Basics.elm
@@ -1,5 +1,17 @@
-
-module Basics where
+module Basics
+    ( (==), (/=)
+    , (<), (>), (<=), (>=), max, min, Order (..), compare
+    , not, (&&), (||), xor, otherwise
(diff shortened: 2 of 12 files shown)


kasbah:
Made some amendments after the CI build failed.

evancz:
Sorry for the delay, this looks good to me! I'll push it to the public repo soon.

src/Graphics/Input.elm
```diff
@@ -17,6 +17,7 @@ To learn about text fields, see the
 import Signal
 import Graphics.Element (Element)
 import Native.Graphics.Input
+import Text -- used internally
```
mgold: Uh-uh - this should be `Native.Text` (with no comment). It will still fix the bug but isn't quite as nice on the source code level.

kasbah:
I should have kept this up-to-date and included 57981c2fef3374410fbdf6dfbab48ed41a566cfa as well. @evancz what would you say to me keeping a WIP PR for stable and you can merge it whenever you want to do a stable release?

The rule would be no API function name changes or additions whatsoever: only bug fixes.

evancz:
Let me see if I understand exactly. We have a PR to stable that just has patch changes. It can be merged in and released whenever.

Does this have any implications for merging in all of master at some later date? Otherwise, this sounds great to me, and I'm happy to do 1.1.2 to get the other fixes in there.

PR #136, opened 2015-01-24
more descriptive local function names in fpsWhen

jvoigtlaender:
After a good night's sleep, thought of these more descriptive function names, hopefully making clear to the reader of this function what the role of the Boolean value used there is.

@jwmerrill, what do you think of these names?


jwmerrill:
I think this is a good change. Nice thought.

evancz:
I like it to, thanks again!

PR #138, opened 2015-01-24
added a missing `return`

jvoigtlaender:
Without that `return`, the `localRuntime.setTimeout` in that line: https://github.com/elm-lang/core/blob/master/src/Native/Time.js#L50 doesn't give a meaningful return value, which means that the `clearTimeout` in this line: https://github.com/elm-lang/core/blob/master/src/Native/Time.js#L58 has no effect whatsoever.

This has led to buggy behavior of `fpsWhen` since at least Elm-0.13 (checked with http://share-elm.com).


evancz:
Ahh, wow, thank you!

Issue #139, opened 2015-01-24
Semantics of `fpsWhen` is quite off.


jvoigtlaender:
So, in https://github.com/jvoigtlaender/core/commit/274cd7eb9a05c1b32dab563a871c557097bc3392 I had been trying (as part of the effort to make `timestamp` work in a consistent manner over all signals) to refactor `fpsWhen` while preserving its semantics. I failed in that, which luckily was caught by @jwmerrill, and fixed in https://github.com/elm-lang/core/commit/6a36cd7244420b1a4cde6af870da76bb2b447361. Throughout all this, I hadn't been checking what the semantics of `fpsWhen` was exactly, only trying to preserve it. Have now looked into the semantics of `fpsWhen` and found serious problems, I think.

Let's consider a signal `isOn` with the following behavior:
* initial value `False`
* at 0.5 seconds, has a `True` event
* at 2.0 seconds, has a `False` event
* at 3.5 seconds, has a `True` event
* at 4.0 seconds, has a `False` event
* no further events

For example,
```elm
isOn = flip member [1,7] <~ keepIf (flip member [1,4,7,8]) 0 (foldp (\_ c -> c+1) 0 (every 500))
```

Let's now consider the signal obtained by `fpsWhen 1 isOn`. Surprisingly, it shows the following behavior (since at least Elm-0.13):
* at 0.5 seconds, has an event with value 0
* at around 1.5 seconds, has an event with value 0
* at 2.0 seconds, has an event with value 0
* at around 2.5 seconds, has an event with value around 1000
* at 3.5 seconds, has an event with (same) value around 1000
* at 4.0 seconds, has an event with (same) value around 1000
* at around 4.5 seconds, has an event with value 0
* no further events

One way to check these claims is to run
```elm
main = asText << snd <~ foldp (\(t',d) (t,l) -> (t',(t'-t,d)::l)) (0,[]) (timestamp (fpsWhen 1 isOn))
```
(for example with different language versions at http://share-elm.com)

Many of the values we see appearing are dubious (for example, `isOn` was `True` only for 2 seconds overall, and yet we see three time deltas of 1 second each). Also, the events at 2.5 and 4.5 seconds shouldn't be happening at all. It turns out that the `clearTimeout` in the implementation of `fpsWhen` has no effect whatsoever. PR https://github.com/elm-lang/core/pull/138 fixes that specific "bogus events" aspect. However, even after that fix, the semantics of `fpsWhen` is still off. For the above situation it now behaves as follows:
[...]

PR #141, opened 2015-01-25
Fix fpsWhen

jvoigtlaender:
Fixes https://github.com/elm-lang/core/issues/139.


jwmerrill:
I think the semantics here are now correct.

I took a stab in #142 at implementing the same semantics in a different way. The main difference is to timestamp the combination of isOn and ticker, instead of timestamping them separately, which allows a little less branching inside the update function. @jvoigtlaender, what do you think about that as an alternative?

jvoigtlaender:
Closing this in favour of https://github.com/elm-lang/core/pull/142.

PR #143, opened 2015-01-26
Fix fpsWhen, including consideration for hotswapping

### LINKED ISSUE #139: Semantics of `fpsWhen` is quite off.

jvoigtlaender:
fixes #139
yet another alternative to #141 and #142 
based on @jwmerrill's version of `fpsWhen`

All local state now packaged in a signal, to allow Elm Reactor to pick it up during hotswapping.

### DIFF
--- a/src/Native/Time.js
+++ b/src/Native/Time.js
@@ -34,9 +34,9 @@ Elm.Native.Time.make = function(localRuntime) {
             delta: 0
         };
 
-        function update(obj,state) {
-            var timestamp = obj._0;
-            var isOn = obj._1;
+        function update(input,state) {
+            var timestamp = input._0;
+            var isOn = input._1;
             var wasOn = state.wasOn;
             var timeoutID = state.timeoutID;
             var previousTime = state.previousTime;
--- a/src/Native/Time.js
+++ b/src/Native/Time.js
@@ -27,13 +27,25 @@ Elm.Native.Time.make = function(localRuntime) {
         // Its value is a tuple with the current timestamp, and the state of isOn
         var input = NS.timestamp(A3(Signal.map2, F2(firstArg), Signal.dropRepeats(isOn), ticker));
 
-        var wasOn = false;
-        var timeoutID;
-        var previousTime = localRuntime.timer.programStart;
+        var timeoutID = 0;
+        if (isOn.value)
+        {
+            timeoutID = localRuntime.setTimeout(notifyTicker, msPerFrame);
+        }
 
-        function update(obj) {
+        var initialState = {
+            wasOn: isOn.value,
+            timeoutID: timeoutID,
+            previousTime: localRuntime.timer.programStart,
+            delta: 0
+        };
+
+        function update(obj,state) {
             var timestamp = obj._0;
             var isOn = obj._1;
+            var wasOn = state.wasOn;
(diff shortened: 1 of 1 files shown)


jwmerrill:
Great. This looks good to me.

evancz:
Great work guys, this looks really good!

jvoigtlaender:
Looks like I pushed this commit: https://github.com/jvoigtlaender/core/commit/f11f7ca3ed973ca9b222dea82b65750c5518095e too late, it occurs in my branch but not in the merge.

Think it is relevant?

jvoigtlaender:
I followed up my previous question with a new pull request (#160).

PR #144, opened 2015-01-26
Make collages resolution independent

salutis:
This should make Elm usable on modern systems with high-resolution screens. Tested with Retina MacBook Pro 15".
Related discussion: https://groups.google.com/forum/#!topic/elm-discuss/pq1KS8XgFiQ

### DIFF
--- a/src/Native/Graphics/Collage.js
+++ b/src/Native/Graphics/Collage.js
@@ -297,8 +297,8 @@ Elm.Native.Graphics.Collage.make = function(localRuntime) {
         canvas.style.height = h + 'px';
         canvas.style.display = "block";
         canvas.style.position = "absolute";
-        canvas.width  = w;
-        canvas.height = h;
+        canvas.width  = w * window.devicePixelRatio;
+        canvas.height = h * window.devicePixelRatio;
         return canvas;
     }
 
@@ -314,8 +314,14 @@ Elm.Native.Graphics.Collage.make = function(localRuntime) {
         var kids = div.childNodes;
         var i = 0;
         function transform(transforms, ctx) {
-            ctx.translate(w/2, h/2);
-            ctx.scale(1,-1);
+            ctx.translate(
+              w / 2 * window.devicePixelRatio,
+              h / 2 * window.devicePixelRatio
+            );
+            ctx.scale(
+              0 + window.devicePixelRatio,
+              0 - window.devicePixelRatio
+            );
             var len = transforms.length;
             for (var i = 0; i < len; ++i) {
                 var m = transforms[i];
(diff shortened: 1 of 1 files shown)


Apanatshka:
I took a look at the support for `devicePixelRatio` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window.devicePixelRatio) which in turn refers to [PKK](http://www.quirksmode.org/blog/archives/2012/06/devicepixelrati.html). I would suggest that you use a local variable and default the value to `1` if `window.devicePixelRatio` doesn't exist:
```
devicePixelRatio = window.devicePixelRatio || 1;
```
And please note that MDN says this property doesn't get updated if the browser is moved between screens with different device pixel ratios.

evancz:
Great, thank you so much! :D

I think there is one place that needs to know about the new initial scale though. Specifically when dealing with an `Element` inserted with `toForm`. I believe there's something like `createTransform` that needs its initial configuration updated.

evancz:
Is there an event that triggers when `windew.devicePixelRatio` changes?

There is already a mechanism to do a redraw asynchronously (for images), so I assume that could be reused?

salutis:
MDN says: There are **no callbacks or events** to listen to when this value changes (for example, if you drag the window between displays with different densities).

evancz:
Oh, I guess the `toForm` stuff should work the same, maybe I was wrong about that! Do you mind testing `toForm` thing to see if it is the right size?

Hmm, is there a recommended behavior for apps that need to cross from screen to screen? I assume going low to high is not a big deal, but is the reverse true?

Also, if the dimensions of the canvas are set initially, but then the scaling is done on a per-render bases, is it possible for them to get out of sync with each other?

salutis:
@evancz, I've just tried to embed _Clock_ using `toForm` like this:
    ```collage 400 400 [(toForm (collage 400 400 ...```
... and it works as expected. Is that what you saying?

evancz:
Yeah, I was worried it might show up half size or something. Another example would be adding `toForm (image 20 20 "http://elm-lang.org/yogi.jpg")` and see if it comes out 20x20.

evancz:
Does the question about crossing from a normal monitor to a hi-DPI monitor make sense?

salutis:
@evancz No issues. This:
```collage 400 400 [toForm (image 20 20 "http://elm-lang.org/yogi.jpg")]```
... draws exactly the same output as the official version at http://elm-lang.org/try.

evancz:
Ah, I have read through the code all in one go and my concern is not realized there. Transferring from lo to hi should be no problem. Thanks a ton for working on this @salutis!

salutis:
My pleasure, @evancz!

Issue #148, opened 2015-01-28
Initial value in a channel ignored in foldp


rehno-lindeque:
The initial value in a `Channel` doesn't appear in a foldp over the channel converted to a Signal.

```elm
type Action = Init
            | ...

actions : Signal.Channel Action
actions = Signal.channel Init

update : Action -> Model -> Model
update action state =
  case Debug.log "action" <| action of      -- Debug.log never prints "action: Init"
    Init -> ...

model : Model -> Signal Action -> Signal Model
model initialState action =
  Signal.foldp update initialState action  -- Seems like the initialState here should be
                                           -- combined with the initial action, either
                                           -- to create the first value or to create the
                                           -- first two values in the signal

main = run emptyModel (Signal.subscribe actions)

```

Initializing a signal with a good `initialState` is not always easy because initialization may depend on a signal coming in from a port in which case you need to do something like this:

```elm
port userId : Signal Int

main = run emptyModel (Signal.zip (Signal.subscribe actions) userId)
```
...in this case creating an initial model based on the incoming `userId` seems near impossible unless some event can be triggered (perhaps by explicitly calling `send` on the `userId` port after the widget has been initialized in order to force an update).

This might assist with issues like this one as well: https://github.com/elm-lang/core/issues/91. I'm not taking the work on promises into account, I don't know if this may be handled already in the upcoming implementation.

jvoigtlaender:
Maybe you want `foldp'` from http://package.elm-lang.org/packages/Apanatshka/elm-signal-extra/3.2.1/Signal-Extra?

rehno-lindeque:
Oh thanks! That one totally slipped my mind, I gave it a try:

This did indeed do the trick...
```
Signal.foldp' update (flip update emptyModel) (Signal.zip action ownerId)
```

This still feels like some pretty strange behavior though? It completely blind-sided me that the first value in the signal was ignored by the fold.

evancz:
The "initialization" phase is different from the "running" phase. Imagine writing this code `foldp (\_ c -> c + 1) 0 Mouse.clicks` and having the initial value be 1!

In the TodoMVC app, I [bring the initial value in as a value, not a signal](https://github.com/evancz/elm-todomvc/blob/master/Todo.elm#L307-L309), and that works great. Is that not plausible for your case?

rehno-lindeque:
[...]

PR #149, opened 2015-01-29
Improved documentation of Signal.foldp

jvoigtlaender:
Meant to keep issues/surprises/questions like https://github.com/elm-lang/core/issues/148 and https://groups.google.com/d/msg/elm-discuss/m4JB7CPXbLI/Dbac1AaAmd0J from arising.

### DIFF
--- a/src/Signal.elm
+++ b/src/Signal.elm
@@ -104,7 +104,7 @@ map5 =
 
 
 
-{-| Create a past-dependent signal. Each update from the incoming signals will
+{-| Create a past-dependent signal. Each update from the incoming signal will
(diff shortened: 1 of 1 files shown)


mietek:
What happened to this PR? I don’t see the commit in the [history of `src/Signal`](https://github.com/elm-lang/core/commits/master/src/Signal.elm).

jvoigtlaender:
This is a very good question. @evancz, you merged this PR at some time, but nevertheless it, and its effect, are not visible. Did you mess with the git history at some point?

The change is still a useful one, I think. Should I simply make a new PR with the same content?

PR #150, opened 2015-01-29
Comprehensive rewrite of List tests

jonathanhefner:
I rewrote the List tests to be more comprehensive.  The tests now cover all exposed API functions, with the exception of `map3..5` (which were also previously not covered).  I can add coverage for `map3..5` as well as non-exposed functions, if desired.

The tests are also now parameterized, so that empty list, 1-element list, and N-element list use-cases are all covered equally.  This also enables easy stress testing, either for stack overflow checking or for benchmarking.

I tried to keep the code as clean and clear as possible.  If there are any parts that need improvement, please let me know.

### DIFF
--- a/tests/Test/List.elm
+++ b/tests/Test/List.elm
@@ -166,4 +166,6 @@ testListOfN n =
             [ test "sorted" <| assertEqual (reverse xs) (sortWith (flip compare) (reverse xs))
             , test "unsorted" <| assertEqual (reverse xs) (sortWith (flip compare) xs)
             ]
+            
+        , test "scanl" <| assertEqual (0 :: (map (\x -> sum [1..x]) xs)) (scanl (+) 0 xs)
         ]
--- a/tests/Test/List.elm
+++ b/tests/Test/List.elm
@@ -1,90 +1,169 @@
 module Test.List (tests) where
 
-import Basics (..)
-
-import List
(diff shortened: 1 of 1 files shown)


evancz:
Very cool, thank you!

My main concern is that if `reverse` is broken, lots of tests will behave in odd ways. That feels weird to me.

jvoigtlaender:
Regarding the `reverse` concern, wouldn't the obvious solution be to replace the `reverse` imported from `List` by a local, guaranteed-to-be-correct implementation of `reverse` in these lines:

https://github.com/elm-lang/core/blob/master/tests/Test/List.elm#L111
https://github.com/elm-lang/core/blob/master/tests/Test/List.elm#L157
https://github.com/elm-lang/core/blob/master/tests/Test/List.elm#L161-162
https://github.com/elm-lang/core/blob/master/tests/Test/List.elm#L166-167

?

jonathanhefner:
I agree, it was a concern of mine.  The problem is (without TCO) you have to rely on either `reverse` or `map` (or their underpinnings: `foldl` and `foldr`, respectively).  Basically, there is always some non-primitive function you are relying on.  But it may be worthwhile to remove as many layers as possible, e.g. limit yourself to just `foldl`, so I'll make another pass at it soon.

PR #152, opened 2015-01-30
 Port List functions from JavaScript to Elm

jonathanhefner:
This builds off of #150, so that should be merged first.

In the same spirit as #130, I ported several List API functions from JavaScript to pure Elm.  Some notes:

* I kept `scanl` in a separate commit because it is a tiny performance regression (1.55 -> 1.65 seconds on an old laptop for a 100k element list).  The regression is due to the requisite extra `reverse` call.
* I did not port `sortBy` in terms of `sortWith` because that resulted in a larger performance regression (2.5 -> 3.0 seconds on an old laptop for a 100k element list).  The regression is (likely) due to the extra string comparison required by the JavaScript `sortWith` implementation.  If we decide the performance penalty is worth it, I can port it also.
* There are several native functions (`any`, `drop`, `take`, `repeat`) which were not converted, but could be by introducing an additional 1 or 2 native functions to drive looping.  Either a function like `foldlWhile` that handles early termination, and / or a function like `foldn` (or `Range.foldl`) that handles counting without creating intermediary lists (i.e. without `[1..n]`).  I can implement these in a separate experimental branch if anyone just wants to see what they would look like.

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -143,7 +143,13 @@ foldr = Native.List.foldr
     scanl (+) 0 [1,2,3,4] == [0,1,3,6,10]
 -}
 scanl : (a -> b -> b) -> b -> List a -> List b
-scanl = Native.List.scanl
+scanl f b xs =
+  let scan1 x accAcc =
+        case accAcc of
+          acc::_ -> (f x acc) :: accAcc
+          [] -> [] -- impossible
+  in
+      foldl scan1 [b] xs |> reverse
 
 {-| Keep only elements that satisfy the predicate.
 
--- a/src/List.elm
+++ b/src/List.elm
@@ -101,8 +101,8 @@ isEmpty xs =
 
 
 member : a -> List a -> Bool
-member =
-  Native.List.member
+member x xs =
+  any (\a -> a == x) xs
 
 
 {-| Apply a function to every element of a list.
@@ -112,7 +112,8 @@ member =
     map not [True,False,True] == [False,True,False]
 -}
 map : (a -> b) -> List a -> List b
-map = Native.List.map
+map f xs =
+  foldr (\x acc -> (f x) :: acc) [] xs
 
 {-| Same as `map` but the function is also applied to the index of each
 element (starting at zero).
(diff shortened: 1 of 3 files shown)


evancz:
This looks great, nice work!

I think the last point about `any`, `drop`, etc. is interesting. It'd be good to move more code into Elm, so I'd be curious to see what this looks like and if it has performance implications.

jvoigtlaender:
Concerning the "implement some functions using native loops" aspect, there are already some candidate implementations for such loops in https://github.com/TheSeamau5/elm-loop.

jonathanhefner:
I created three branches to show what looping could be like for `any`, `repeat`, `take`, and `drop`:

* [Using foldn](https://github.com/jonathanhefner/elm-core/tree/experiment-foldn)
  * `any` cannot be implement via `foldn`, and thus is not ported
* [Using foldlWhile](https://github.com/jonathanhefner/elm-core/tree/experiment-foldlWhile)
  * `repeat` cannot be implemented via `foldlWhile`, but a safe, less-efficient, pure-Elm version is included for consideration (note the ** in the benchmark data)
* [Using loopUntil](https://github.com/jonathanhefner/elm-core/tree/experiment-loopUntil)
  * all 4 functions are possible, however the iteration logic was quite confusing to write (I say this as someone who grew up writing C, C#, and Java)


When benchmarking the branches, I first ran the combined tests for just the 4 functions on a 100k element list (that's the first column).  That didn't provide a clear enough picture, so then I isolated the tests for each function (that's the remaining columns).  For `take` and `drop`, there were multiple tests doing similar work, so I pared it down to just one test taking or dropping 99,999 elements from a 100k element list.

branch     | All 4  | `any` | `repeat`  | `take` n-1  | `drop` n-1
-----------|--------|-------|-----------|-------------|-----------
master     | 3.15s  | 1.0s  | 1.4s      | 1.4s        | 1.1s
foldn      | 4.0s   | n/a   | 1.35s     | 1.65s       | 1.15s
foldlWhile | 4.15s  | 1.05s | 1.45s**   | 1.6s        | 1.3s
loopUntil  | 4.65s  | 1.4s  | 1.55s     | 1.65s       | 1.35s

PR #153, opened 2015-02-02
Fixing JSON decoding for Ints.

krisajenkins:
One way to check if a number is an integer in JavaScript is (value|0).
Sadly, this fails if the value is greater than 2^32, but still less than
Number.MAX_SAFE_INT.

This patch switches the Json decoder to use a polyfill of the proposed
ES6 Number.isInteger function.

### DIFF
--- a/src/Native/Json.js
+++ b/src/Native/Json.js
@@ -13,6 +13,9 @@ Elm.Native.Json.make = function(localRuntime) {
 	var Result = Elm.Result.make(localRuntime);
 	var Utils = Elm.Native.Utils.make(localRuntime);
 
+	function isInteger (nVal) {
+		return typeof nVal === "number" && isFinite(nVal) && nVal > -9007199254740992 && nVal < 9007199254740992 && Math.floor(nVal) === nVal;
+	}
 
 	function crash(expected, actual) {
 		throw new Error(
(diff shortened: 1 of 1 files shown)


ajhager:
I have benchmarked and unit tested the original code, your proposed change, and a new method of my own design (original, proposed, and blend respectively below.) I used the tests from https://github.com/parshap/js-is-integer/blob/master/test.js and the benchmarks are at http://jsperf.com/isinteger-elm.

The original does indeed fail for values over and under 2147483647. proposed and blend are both correct, but blend is only slightly slower than the original in the general case and nearly equivalent in the typical case.

```js
function original(value) {
  return typeof value === 'number' && (value|0) === value
}

function proposed(value) {
  return typeof value === "number" && isFinite(value) && value > -9007199254740992 && value < 9007199254740992 && Math.floor(value) === value;
}

function blend(value) {
  if (typeof value !== 'number') {
    return false;
  }
    
  if (value < 2147483647 && value > -2147483647) {
    return (value|0) === value;
  }
       
  return isFinite(value) && !(value % 1);
}    
```

evancz:
@krisajenkins, thanks for providing a fix for this! I was not sure whether the behavior you expect is the behavior everyone would expect. If @ajhager agrees, I guess there is some sort of consensus :)

I'm trying this new thing out where I close issues more aggressively, even if they are a good idea that should happen. In this case, it seems like there are multiple folks who can do the updated fix, so to avoid a chance of blocking, can we do a new PR with the faster version and some tests?

ajhager:
I will do the new PR with tests.

evancz:
Awesome, thank you!

Also, sorry for the huge delay @krisajenkins, and thank you again @ajhager for unsticking things! :)

PR #158, opened 2015-02-03
Comprehensive rewrite of Char tests

jonathanhefner:
Like #150, I rewrote the Char tests to be more comprehensive.  All exposed API functions are covered and tested with all alpha-decimal characters.  The tests are also more structured, so it is easier to add tests for future corner cases (e.g. `toLocaleLower` in a Turkish locale).

### DIFF
--- a/tests/Test/Char.elm
+++ b/tests/Test/Char.elm
@@ -1,45 +1,94 @@
 module Test.Char (tests) where
 
 import Basics (..)
-import Char
+import Char (..)
+import List
 
 import ElmTest.Assertion (..)
 import ElmTest.Test (..)
 
+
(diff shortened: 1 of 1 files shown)


jonathanhefner:
I'm... not sure what happened to the build.  Travis CI says the build timed out, but the tests run swiftly (and pass) locally.  Is it possibly a glitch in the build system?  Is there a way to trigger another attempt?

jonathanhefner:
Looks like it was just a glitch.

In case anyone is curious, I was able to force another build by doing an empty `git commit --amend` and then a `git push -f` to my branch.

evancz:
Why do you put single tokens in parentheses?

jonathanhefner:
Just for consistency.  I guess it also feels a little easier to eyeball with the parens acting as delimiters, but that's subjective.

If it's bad practice (or just ugly), I can strip them out.

evancz:
I'd generally consider it bad style to have extra parens, but I only care to the extent that people will use these tests as a guide for writing their own. I'm not sure what the right thing is here, but it feels weird.

In any case, things look good, thanks!

PR #161, opened 2015-02-06
Text module Refactoring *WIP*

jazmit:
WIP on Text refactoring.  Anyone interested is welcome to comment and advise.
The idea is to normalize the internal representation of `Text` objects ~~to `{Array.<{text: string, style: Object.<string, string}>}`~~.  The keys and values of the style object will still be valid CSS keys and values, but this will make it easier to manipalate text objects without resorting to using the DOM API.
I've put a TODO by every function I think needs to change.

TODO
- [ ] Add tests for Text module
- [x] Change Text representation
- [ ] Move text-related functions from Util to Text
- [ ] Add comments for Text module

### DIFF
--- a/tests/Native/Test/Text.js
+++ b/tests/Native/Test/Text.js
@@ -1,12 +1,13 @@
 Elm.Native.Test      = Elm.Native.Test || {};
 Elm.Native.Test.Text = {};
 Elm.Native.Test.Text.make = function(localRuntime) {
-    localRuntime.Native           = localRuntime.Native           || {};
-    localRuntime.Native.Test      = localRuntime.Native.Test      || {};
+    localRuntime.Native = localRuntime.Native || {};
+    localRuntime.Native.Test = localRuntime.Native.Test || {};
     localRuntime.Native.Test.Text = localRuntime.Native.Test.Text || {};
 
-    if (localRuntime.Native.Test.Text.values)
+    if (localRuntime.Native.Test.Text.values) {
         return localRuntime.Native.Test.Text.values;
+    }
 
     var Utils = Elm.Native.Utils.make(localRuntime);
 
--- a/tests/Native/Test/Text.js
+++ b/tests/Native/Test/Text.js
@@ -1,0 +1,23 @@
+Elm.Native.Test      = Elm.Native.Test || {};
+Elm.Native.Test.Text = {};
+Elm.Native.Test.Text.make = function(localRuntime) {
+    localRuntime.Native           = localRuntime.Native           || {};
+    localRuntime.Native.Test      = localRuntime.Native.Test      || {};
+    localRuntime.Native.Test.Text = localRuntime.Native.Test.Text || {};
+
+    if (localRuntime.Native.Test.Text.values)
+        return localRuntime.Native.Test.Text.values;
+
+    var Utils = Elm.Native.Utils.make(localRuntime);
+
+    function textToHtmlString(text) {
+        // makeText returns a String object
+        // use toString() to convert to a string primitive
+        // which can be interpreted as an Elm String
+        return Utils.makeText(text).toString();
+    }
+
+    return localRuntime.Native.Test.Text.values = {
+        textToHtmlString: textToHtmlString
+    };
+};
--- a/tests/Test/Text.elm
+++ b/tests/Test/Text.elm
@@ -8,14 +8,17 @@ import Native.Test.Text
 
 
 tests : Test
-tests = suite "Text Tests"
-    [ testFromString
-    ]
+tests =
+    suite "Text Tests"
+        [ testFromString
+        ]
 
 toHtmlString : Text -> String
-toHtmlString = Native.Test.Text.textToHtmlString
+toHtmlString =
+    Native.Test.Text.textToHtmlString
 
 testFromString : Test
 testFromString =
     let result = toHtmlString (fromString "Hello")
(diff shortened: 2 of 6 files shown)


evancz:
I'm not sure what you have proposed as the data structure, but I think it'd be best to model it like any other union type in elm:

```elm
type Text = Literal String | Styled String String Text | Append Text Text
```

You can do all that in JS, but I'd *think* of like a union type.

jazmit:
Ah.. lightbulb.. I was wondering why everything was implemented as linked lists.. so maybe we should keep the current representation but just replace the current CSS string with [key, value] pairs?

jazmit:
toString on a Text object returns "<internal structure>" for me.. is this intended behaviour?

tests/Native/Test/Text.js
```diff
@@ -0,0 +1,23 @@
+Elm.Native.Test      = Elm.Native.Test || {};
+Elm.Native.Test.Text = {};
+Elm.Native.Test.Text.make = function(localRuntime) {
+    localRuntime.Native           = localRuntime.Native           || {};
+    localRuntime.Native.Test      = localRuntime.Native.Test      || {};
+    localRuntime.Native.Test.Text = localRuntime.Native.Test.Text || {};
```
evancz: Don't do this aligning stuff. It's really frail. When any line changes, lots of things need to change. Not worth it in the long run!

tests/Native/Test/Text.js
```diff
@@ -0,0 +1,23 @@
...
+
+    if (localRuntime.Native.Test.Text.values)
+        return localRuntime.Native.Test.Text.values;
```
evancz: Always use curly braces!

tests/Test/Text.elm
```diff
@@ -0,0 +1,21 @@
...
+
+import ElmTest.Assertion (..)
+import ElmTest.Test (..)
+import Basics (..)
+import Text (..)
+import Native.Test.Text
+
+
+tests : Test
+tests = suite "Text Tests"
```
evancz: Always bring the definition down a line:

```elm
tests =
  suite "Text Tests"
    [ ...
    ]
```

tests/Test/Text.elm
```diff
@@ -0,0 +1,21 @@
...
+    [ testFromString
+    ]
+
+toHtmlString : Text -> String
+toHtmlString = Native.Test.Text.textToHtmlString
```
evancz: Same here, always bring it down.

tests/Test/Text.elm
```diff
@@ -0,0 +1,21 @@
...
+
+testFromString : Test
+testFromString =
+    let result = toHtmlString (fromString "Hello")
+     in test "simple" <| assertEqual "Hello" result
```
evancz:
```elm
testFromString =
    let result = toHtmlString (fromString "Hello")
    in
        test "simple" <| assertEqual "Hello" result 
```

evancz:
I am unsure about how to ideally test this. If it is represented directly as a Union Type in Elm, we can run tests by making sure it's the value we expect. Perhaps we can do it this way and add a "secret" tag that only `toString` looks for. That'd mean making a special Union Type in JS:

```elm
function append(text1, text2) {
    return {
        ctor: 'Append',
        _0: text1,
        _1: text2,
        isText: true
    };
}
```

So it'll look like a union type to all of Elm, but when `toString` looks at it, it can check for `isText`.

And yes, I made it `<internal structure>` on purpose. It was not clear to me what the "right" way to visualize it given that it really is not a string and that `Text.fromString "hello"` is not the same as `Text.bold (Text.fromString "hello")`. I don't really know the right answer here, but I don't feel that we should expose the internal structure. What do you think?

Also, we were talking about [moving some `Text` functions around](https://groups.google.com/d/msg/elm-discuss/ydhKsvJtDZw/ZNfktG1HrLcJ). Might be something to consider here.

jazmit:
I think exposing `type Text = Literal String | Styled String String Text | Append Text Text` is probably not such a frightening idea.. after all I would think we'd want it to be possible to manipulate and process Text values from client code without depending on Native modules.  However, this can be left for another pull request, as it won't necessarily make for great tests - they'd turn out kinda tautological:
````elm
assertEqual (Text.append text1 text2) (Append text1 text2)
````
If we're testing conversion to and from another format I'd be more confident that code changes aren't breaking anything.  Since it's not outrageous that we might one day provide a `toHtmlString : Text -> String` function, I think that continuing with the current approach is probably OK.  If one day this function does exist it just requires changing one line of the test.

jazmit:
I'll have a think about moving the asText function to the Element module.

jazmit:
Any objections to pulling in the Elm QuickCheck library to help with the testing here?

evancz:
Okay, things should be refactored now. Not sure about speed, but at least we are no longer going through HTML in the Collage code!

I think we need to look into how `\n` behaves in `Text` rendered to the `collage`? My guess is badly.

I don't mind about `elm-check`, but I think do that kind of thing separate from the `Text` stuff.

evancz:
As an aside, this whole addition is really cool :) I was testing it out to make sure everything was in order, and it was very fun!

evancz:
All the `Graphics.*` modules have moved to `evancz/elm-graphics` so it makes sense to retarget stuff like this. That module also did get refactored for this change and I think it's a lot nicer.

PR #164, opened 2015-02-07
Re-introducing `Maybe.isNothing`

jvoigtlaender:
One motivating example:
---------------------------------
```elm
dblclicks : Time -> Signal ()
dblclicks d =
  let step t' s = case s of
                    Nothing -> Just t'
                    Just t  -> if t'-t < d then Nothing else Just t'
  in map (always ())
         (keepIf isNothing Nothing
                 (foldp step Nothing
                        (map fst (timestamp Mouse.clicks))))
```
It's a pain to have to write something like `(\m -> case m of {Nothing -> True; _ -> False})` instead of `isNothing` above.

One possible rebuttal:
-----------------------------
"But you could write `((==) Nothing)` instead of the above explicit lambda-expression with pattern-matching."

My answer: That's not really good in general. It pretends that `(==)` works okay on all types (note that `isNothing : Maybe a -> Bool` is polymorphic in `a`). But in the future, to prevent runtime errors like that arising from `identity == (\x -> x)`, it might come to pass that the compiler restricts uses of `(==)`. It would have to restrict them not only on function types, but also somehow on polymorphism (since any `a` could turn into a `b->c`). Then suddenly `((==) Nothing)` would be less generally applicable than `isNothing`.

Another possible rebuttal:
----------------------------------
"Uses of `fromJust` and `isNothing` smell bad. Pattern-matching should be used instead."

My answer: I'm *not* proposing to re-inroduce `fromJust`. I yet have to see evidence of a piece of code that uses `isNothing` (but *not* `fromJust`) and seems bad style because of that use of `isNothing`. Also my example above uses only `isNothing`, not `fromJust`. I don't know what there could be to dislike about the use of `isNothing` in there. (See also [this post](https://groups.google.com/d/msg/elm-discuss/4PSA48ws5JU/E1aLBWL__rgJ).)

A cheap argument from my side:
-------------------------------------------
We do have `List.isEmpty` (but not `isNotEmpty`, `head : [a] -> a`, and `tail : [a] -> [a]`) in the standard lib. The consistent thing to do w.r.t. `Maybe` is to have `Maybe.isNothing` (but not `isJust` and `fromJust : Maybe a -> a`) in the standard lib.

Evidence from other people's projects:
---------------------------------------------------
Searching GitHub gives the following places where Elm users have felt a need to introduce `isNothing` in their own code.
* https://github.com/etaque/tacks/blob/b0348de9a79b15d83a46847613f4c0f86e04ee7e/game/Core.elm#L53-L57
* https://github.com/dreuter/elm-duotris/blob/a6b80816b96b077eb8b681cd0927a3fbb31c4c95/duotris.elm#L92-L94
* https://github.com/drenninghoff/bachelorthesis/blob/2622cb9a0d6e011f99bf39f23b145b6330a931d3/code/PrologTree.elm#L202-L204

(Disclosure: Two of the above are students of mine. I don't think that is a relevant bias though.)

In addition, there are various GitHub hits where Elm users have used `isNothing` without defining it themselves. Apparently these projects are not yet upgraded from 0.13 to 0.14. If they were, maybe not all of these uses could be replaced by pattern-matching without pain, so at least in some of these cases it would be likely that these people would also start introducing their own implementation of `isNothing`:
* https://github.com/osnr/Birdhouse/blob/08f2366866109ddd67ecdf02726d9185e6b873a4/src/elm/Utils.elm#L24
* https://github.com/kpjjpk/tp-final-prog-funcional/blob/800605827ab9b0672e39e0ee9b23789a163e5543/AntColony/Logic/Pheromone.elm#L14
* https://github.com/Apanatshka/phybas-graviz/blob/523f9454cd7ebf01ae85f20ac7f78580cf967492/TGF.elm#L76
* https://github.com/JetBrains/CodingGame/blob/87a2f92b613e51856f766145367ad7cdcc9de56c/Game.elm#L183
* https://github.com/Timoha/whiteboard-tpa/blob/a0af6347ac2090caa6cff4836ea7dc13b3eb6fbf/public-dev/src/Editor.elm#L269
* https://github.com/grzegorzbalcerek/chess-elm/blob/c029f4e3ea1432fd4310099ace91aa37b5c46aa8/Chess/Game.elm#L239
* https://github.com/adamw/traffic/blob/ab8bd6e3957190baed8b7efe7c78ae91bcb1ab3f/Physics/TrafficLights.elm#L101
* https://github.com/SimSaladin/hajong/blob/1b1b8173ed87d7f6bcc618f3137e9b8eafa0ad35/hajong-client-web/Game.elm#L141
* https://github.com/Sheeo/Klan-Bracket/blob/c0afdcf1857b423c65d8f83733b3016fa3102e0d/Bracket.elm#L120

I don't think it is possible to dismiss all the uses above with "`isNothing` is bad style, use pattern-matching instead".

Note that it is also well possible that additional uses exist where people have simply chosen a different name but implemented the same functionality.

About the process here
-------------------------------
I know that the proper process to suggest an addition to the standard lib now is to first create a new package like `elm-maybe-extras`, put the new function there, wait a while to see whether it gets used. But let's be realistic. For a simple function like `Maybe.isNothing`, almost nobody will make the effort to add the required line to `elm-package.json` and add the required line to import the extra module into their code, just to be able to use that simple function which can just as easily be defined in one's own code. So what would happen is that the extra package would not get used but *still* we would see the proliferation of repeated definitions of `isNothing` in different people's projects. The only way I see to prevent such redundancy is to add `isNothing` right to the standard lib.


jvoigtlaender:
Can I bump this again? I'm still lacking any clue about what is considered harmful about `isNothing`. And as expressed above, pushing people into either redundantly re-defining it again and again, or into using something like `((==) Nothing)`, will most likely be regretted down the road. (Also, there seems to be no recommendation to use `((==) [])` instead of `isEmpty` in the list case.)

rtfeldman:
This seems useful and harmless. +1 in my book.

mgold:
Seems harmless. `filterMap` handles a lot of common cases but the double clicks example shows that it's not a silver bullet.

jvoigtlaender:
Can this be added to https://github.com/elm-lang/core/issues/322?

evancz:
The process is to go through *-extra, which it looks like has happened here.

I'd like to see the arguments for why things are useful be folded into that process. So when it comes time to review the list and maybe and result and whatever else libraries, we have all that stuff easily accessible. I am not sure if that happened in this case, so I'll add it to the meta issue. In the future, try to put it in the extra process though.

mgold:
I think that it only makes sense to have either both `isJust` and `isNothing`, or neither. They do come in handy when you don't want to do a pattern match.

jvoigtlaender:
And yet we have only `List.isEmpty`, not also `List.containsSomething`.

I'm not sure `Maybe.isNothing` and `Maybe.isJust` are equally useful.

What is an example where one would prefer `isJust` over a pattern match (and over `isNothing`)?

mgold:
Okay, that's an interesting point. Another function that is sometimes useful is `dropNothings : List (Maybe a) -> List a` but you can usually work around that with `List.concatMap`.

PR #168, opened 2015-02-11
Make sure Http requests for file:// scheme would not stall.

mblsha:
When XHR is finished for file:// resource, its status code is set to 0 both in cases of success and failure.

Guess this solution is better than stalling the XHR forever.


evancz:
Thanks! We're currently working on a big rewrite of the `Http` library on the `promises` branch, so I think we shouldn't continue doing stuff to the old `Http` library. There are still some decisions to be made about whether the library or the user is in charge of dealing with these kinds of details, so this is good timing :) In any case, this information will be valuable over there.

PR #171, opened 2015-02-13
Fixed bug in Array.map and Array.indexedMap

brown-dragon:
Array.map and Array.indexedMap have a bug that can most easily be seen
with the following use case:

EXPECTED: ```Array.empty == Array.map (\e -> e) Array.empty```
ACTUAL: ```Array.empty != Array.map (\e -> e) Array.empty```

(the same for Array.indexedMap)

This is because Array.map/Array.indexedMap initializes itself
incorrectly using:

        table: new Array(a.table)  // creates a new array with an array element

instead of

        table: new Array(a.table.length) // a correctly-sized array

This bug has been fixed in this commit

### DIFF
--- a/src/Native/Array.js
+++ b/src/Native/Array.js
@@ -320,7 +320,7 @@ Elm.Native.Array.make = function(localRuntime) {
         var newA = {
             ctor: "_Array",
             height: a.height,
-            table: new Array(a.table)
+            table: new Array(a.table.length)
         };
         if (a.height > 0)
         {
@@ -345,7 +345,7 @@ Elm.Native.Array.make = function(localRuntime) {
         var newA = {
             ctor: "_Array",
             height: a.height,
-            table: new Array(a.table)
+            table: new Array(a.table.length)
         };
         if (a.height > 0)
         {


brown-dragon:
I just realized this bug has been reported in https://github.com/elm-lang/core/pull/169

However this is a better fix, especially because the bug exists in all `Array.map` calls
(not just the empty case). It is most visible in the empty case but the initialization is
incorrect in all cases.

evancz:
Awesome, thanks for the fix!

TheSeamau5:
Yay! Awesome! 

By the way, I just ran elm-check on the old code and it has found the bug.

```
Map transforms lists and arrays equally has failed with the following input: (<function>,Array.fromList [])
```

I know it's a bit of a shameless plug, but I think we could look into elm-check integration with the current testing of the standard library. I'm pretty surprised by how well elm-check is currently doing.

pchiusano:
Any chance of a 0.14.x patch release for this?

kasbah:
Prompted by you @pchiusano, I proposed it for next stable with #192.

PR #180, opened 2015-02-21
Fix a tiny typo in date module toTime function doc (1990 to 1970)

audriussagadinas:
Change 1990 to 1970 where 1990 was incorrect. Also specify that time is returned in milliseconds, because, by (common?) definition, UNIX time is number of SECONDS that have elapsed since midnight (UTC) of Thursday, 1 January 1970.

Remove extra empty line at file end.

### DIFF
--- a/src/Date.elm
+++ b/src/Date.elm
@@ -44,9 +44,9 @@ fromString =
   Native.Date.read
 
 
-{-| Convert a date into a time since midnight (UTC) of 1 January 1990 (i.e.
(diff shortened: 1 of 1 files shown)


evancz:
Ah, thanks! Do you mind retargeting this change to the `promises` branch though? That's where all the latest work is happening.

evancz:
Go for #275 which also fixes this. Sorry for the big delay!

PR #182, opened 2015-02-24
Fix runtime error in native toString

ThomasWeiser:
`toString` needs special case for `null` value, otherwise it throws a type error.

To reproduce the bug: `toString Json.Encode.null`
*TypeError: Cannot use 'in' operator to search for '_' in null*


ThomasWeiser:
I am unsure whether this bug-fix PR should go to stable branch or to master branch.

kasbah:
If it is merged into master I'll cherry pick it for next stable.

ThomasWeiser:
Ok thanks. I have made a new PR (#183) against master branch.

PR #183, opened 2015-02-24
Fix runtime error in native toString (handling null values)

ThomasWeiser:
`toString` needs special case for `null` value, otherwise it throws a type error.

To reproduce the bug: `toString Json.Encode.null`
*TypeError: Cannot use 'in' operator to search for '_' in null*

### DIFF
--- a/src/Native/Utils.js
+++ b/src/Native/Utils.js
@@ -357,6 +357,10 @@ Elm.Native.Utils.make = function(localRuntime) {
 
 	var toString = function(v)
 	{
+		if (v === null)
+			{
+				return '<internal structure>';
+			}
 		var type = typeof v;
 		if (type === 'function')
 		{
--- a/src/Native/Utils.js
+++ b/src/Native/Utils.js
@@ -358,9 +358,9 @@ Elm.Native.Utils.make = function(localRuntime) {
 	var toString = function(v)
 	{
 		if (v === null)
-			{
-				return '<internal structure>';
-			}
+		{
+			return '<internal structure>';
+		}
 		var type = typeof v;
 		if (type === 'function')
 		{


kasbah:
Why was this closed?

ThomasWeiser:
First I made the PR #182 against the stable branch. You wrote a comment 
that made me think I should make the PR against the master branch. So I 
did (#183, still open) and closed the first one. Should I reopen #182?

Thanks!

kasbah:
Yes, no it's fine. I deleted my comment a second after I made it actually. I erroneously thought this PR was closed without comment.

kasbah:
Though you probably want to rebase this on current master so it can be merged more easily.

ThomasWeiser:
Ok, I rebased the commit onto current master

ThomasWeiser:
@evancz: Would be nice to have the fix in the upcoming 0.15 release.

evancz:
I don't think this is really the right fix. I would expect a `Json.Value` to appear as an `<internal value>` or something. I don't think raw JS objects should be printed out the same as records or whatever. I'm not sure how to do this at the moment though.

ThomasWeiser:
I am not sure, whether `Json.Value` should be printable or not. I find it very useful to be able to `Debug.log` these values. Otherwise one would have to patch the generated JavaScript code to insert the logging, which is much more awkward.

Anyway, the current code for `toString` is incomplete in the sense, that it throws a confusing runtime error for the JavaScript `null` value. IMHO this should be fixed anyway.

ThomasWeiser:
> I don't think this is really the right fix. I would expect a Json.Value to appear as an <internal value> or something.

Ok, I have modified the the code to return `"<internal structure>"` for `null` values, as @evancz proposed.

ThomasWeiser:
Please consider fixing for the upcoming 3.0.0 release (I've just rebased the PR).

This unnecessary runtime error did bite me multiple times.

src/Native/Utils.js
```diff
@@ -357,6 +357,10 @@ Elm.Native.Utils.make = function(localRuntime) {
 
 	var toString = function(v)
 	{
+		if (v === null)
+			{
+				return '<internal structure>';
+			}
```
shamrin: The block above seems to have wrong indentation. Tabs vs spaces?

In addition, should this code also handle `undefined`? E.g. [with `== null`][1].

[1]: http://marijnhaverbeke.nl/blog/null-and-undefined.html

ThomasWeiser: As far as I can see the source and my fix are both tab indented.

`undefined` is handled well with the existing code, so it seems sufficient just to test for `null`.

ThomasWeiser:
Rebased again on current master

src/Native/Utils.js
shamrin: You are right, there are only tabs in the code. However, you still have unnecessary indent for `if` body. Brace `{` should be below `i`.

ThomasWeiser: ooops

eeue56:
FWIW it is very useful to be able to log `Json.Value` and see the actual JS object. E.g, when working with Http requests, it's a lot nicer if I can debug the data coming down right in my Elm code where it comes down, rather than make a port which I have to send stuff to in order to debug.

evancz:
https://github.com/elm-lang/core/blob/master/src/Native/Utils.js#L340-L343

PR #186, opened 2015-03-05
Documentation: Graphics.Element (absolute, relative)

justinmanley:
It was unclear to me how different types of `Position`s (`Absolute Int` vs `Relative Float`), so I added a note to the documentation.

It's a bit confusing. As it turns out, the important difference is not int vs. float, but how the `number` is interpreted. `Absolute` positions are interpreted as pixel values, while `Relative` positions are interpreted as percentage values.

See https://github.com/elm-lang/core/blob/abcb93e93a263cdd8fda739735f87974dc63729a/src/Native/Graphics/Element.js#L221 to understand how `Absolute` / `Relative` values are used in the native code.

### DIFF
--- a/src/Graphics/Element.elm
+++ b/src/Graphics/Element.elm
@@ -334,9 +334,11 @@ layers es =
 
 
 -- Repetitive things --
-
+{-| A pixel value. -}
 absolute : Int -> Pos
 absolute = Absolute
+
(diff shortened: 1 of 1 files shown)


evancz:
I added in something along these lines, hopefully it is better. Thanks for pointing this out!

PR #187, opened 2015-03-05
Fix List.head doc comment with new Maybe return type

etaque:


evancz:
Thanks for the PR, these docs got fixed at some point though!

PR #191, opened 2015-03-10
Caching of the Image object in FImage's form drawImage (native)

silverio:
The drawing of the "FImage" form (used by e.g. the currently undocumented Graphics.Collage.sprite) currently creates the Image object every time the form is drawn (which effectively loads the image again), making it practically unusable.

A proposal here is to cache the image object as a property of the corresponding form object, first time it is attempted to be drawn. 

Another possibility would be to create and cache it earlier, at the stage of the form object creation.

In my opinion, even better long term alternative would be to include something like jcollard's sprite library (https://github.com/jcollard/elm-util) as part of the core (one caveat there being that it's relying on an  element for drawing, as opposed to the "sprite" one that does direct canvas drawing, without DOM intervention).

### DIFF
--- a/src/Native/Graphics/Collage.js
+++ b/src/Native/Graphics/Collage.js
@@ -299,11 +299,13 @@ Elm.Native.Graphics.Collage.make = function(localRuntime) {
 
 	// IMAGES
 
-	function drawImage(redo, ctx, form)
-	{
-		var img = new Image();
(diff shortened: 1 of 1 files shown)


MrMormon:
@silverio Is there a more relevant PR elsewhere?

evancz:

PR #193, opened 2015-03-10
fixing Stream.sample in Elm-0.15

jvoigtlaender:
The current 0.15 version is broken for situations like this:
```elm
main = Varying.map Graphics.Element.show (Varying.fromStream (0,0) s)

s = let v = Mouse.position
    in Stream.sample always v (Stream.fromVarying v)
```
The output is never updating! That cannot be what is supposed to happen, right?

The commit here fixes this in the way that changes the existing code of `Stream.sample` the least.

### DIFF
--- a/src/Stream.elm
+++ b/src/Stream.elm
@@ -203,7 +203,7 @@ sample f varying events =
           (Native.Signal.initialValue varying, fromVarying varying)
 
       sampleEvents =
-          merge
+          Native.Signal.genericMerge (\(Sample s) (Update u) -> SampleAndUpdate s u)
             (map Sample events)
             (map Update varyingUpdates)
   in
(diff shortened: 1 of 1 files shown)


evancz:
Nice find, thank you for the fix!

I wonder if this pattern of A, B, or both will show up in other places.

jvoigtlaender:
In such situtations, what I'd actually do is to model the thing as `(Maybe A, Maybe B)` rather than via `type SomeOfAB a b = OnlyA a | OnlyB b | Both a b`. For the concrete case, doing it like this: https://github.com/jvoigtlaender/core/commit/f7262c0c846f64ab8db5e6eaaa6edf68ddf2fee4. There, it simplifies the definition of the `sampleUpdate` function quite a bit.

The only reason I didn't right away do this in the pull request here is that I wanted it to be the minimal fix (in terms of changed lines and cognitive load).

evancz:
That looks great to me, do you mind making a PR for that?

I also changed `fps` to a `Stream` which I think simplified stuff there, but I may have messed things up.

jvoigtlaender:
Have added a PR here: https://github.com/elm-lang/core/pull/195.

I'll also try to take a look at the new version of `fps` separately.

PR #195, opened 2015-03-12
simplified sampleUpdate in Stream.sample

jvoigtlaender:
... and renamed some type variables so that `a` and `b` in `SampleEvent` and `SampleState` correspond to the roles of `a` and `b` in the type of `sample`.

### DIFF
--- a/src/Stream.elm
+++ b/src/Stream.elm
@@ -203,16 +203,16 @@ sample f varying events =
           (Native.Signal.initialValue varying, fromVarying varying)
 
       sampleEvents =
-          Native.Signal.genericMerge (\s u -> (fst s, snd u))
-            (map (\s -> (Just s, Nothing)) events)
-            (map (\u -> (Nothing, Just u)) varyingUpdates)
+          Native.Signal.genericMerge (\u s -> (fst u, snd s))
+            (map (\u -> (Just u, Nothing)) varyingUpdates)
+            (map (\s -> (Nothing, Just s)) events)
   in
       fold sampleUpdate { state = initialValue, trigger = Nothing } sampleEvents
         |> fromVarying
         |> filterMap (\state -> Maybe.map (f state.state) state.trigger)
 
 
-type alias SampleEvent a b = (Maybe b, Maybe a)
+type alias SampleEvent a b = (Maybe a, Maybe b)
 
 
 type alias SampleState a b =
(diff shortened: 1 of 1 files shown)


src/Stream.elm
```diff
@@ -203,40 +203,28 @@ sample f varying events =
...
-    Update b ->
-        { state = b
-        , trigger = Nothing
-        }
-
-    SampleAndUpdate a b ->
-        { state = b
-        , trigger = Just a
+sampleUpdate : SampleEvent b a -> SampleState b a -> SampleState b a
+sampleUpdate (ms,mu) state =
```
evancz: Let's do better names than `mu` and `ms`. I don't know what they mean, and saving characters is not a good goal.

src/Stream.elm
```diff
@@ -203,40 +203,28 @@ sample f varying events =
...
+            (map (\s -> (Just s, Nothing)) events)
+            (map (\u -> (Nothing, Just u)) varyingUpdates)
   in
       fold sampleUpdate { state = initialValue, trigger = Nothing } sampleEvents
         |> fromVarying
         |> filterMap (\state -> Maybe.map (f state.state) state.trigger)
 
 
-type SampleEvent a b = Sample a | Update b | SampleAndUpdate a b
+type alias SampleEvent b a = (Maybe b, Maybe a)
```
evancz: What's the reason for switching the order here? Maybe we can give helpful names to these type variables instead?

evancz: Oh, I guess to match the function given to `sample`?

jvoigtlaender: Yes, exactly.

src/Stream.elm
jvoigtlaender: `maybeUpdate` and `maybeSample`?

evancz: `maybeNewState` and `trigger` seem clearest to me.

src/Stream.elm
evancz: I think this is far enough from convention to be pretty confusing. I'd either change the order of things such that it just matches (switching the args in `genericMerge`) or not bother with trying to get `a` and `b` to match.

evancz: I also think the type alias is not making things clearer at this point.

jvoigtlaender:
How about now? The record type alias with named components makes it clearer?

evancz:
I did some last bits of cleanup, but I this is overall a lot nicer!

PR #207, opened 2015-04-04
Fix: Call Element.newElement as a native function

ThomasWeiser:
newElement is now a native function in Native.Graphics.Element.js

It is not exported by Element.elm.
So calling it as Element.newElement throws a runtime error


ThomasWeiser:
It's fixed in the meantime

evancz:
Wait, so this got fixed since it was opened?

ThomasWeiser:
Yes, it got fixed with your commit 5b0bc59987f67f7d39c09c623fc198790324ac0e

evancz:
Cool, sorry for the confusion and thanks for the PR!

PR #209, opened 2015-04-09
fixing fpsWhen in a post-0.15 future

jvoigtlaender:
This is retargeting PR https://github.com/elm-lang/core/pull/198/ towards the `future` branch.


jvoigtlaender:
Does this mean we are not going to *have* a future with a separate stream concept?

evancz:
Oops, I cleaned out a bunch of branches that were zero commits in front of master not realizing that `future` was purposely behind. Do you happen to know which commit corresponds to the future branch so we can tag it?

jvoigtlaender:
Oh. I'm not even sure I know how git is working, so "a commit corresponding to a branch"? Well, my best guessing would go like that: My [branch](https://github.com/jvoigtlaender/core/tree/typing-in-fpsWhen) in which I was working for this PR here, and which was diffed against the `future` branch to make this PR, currently ends with [this commit](https://github.com/elm-lang/core/commit/ec22bcc4818d2e806d80107d94060394392856fe). So maybe that is the one you are looking for. It seems that would make sense, because it happened one day before the current `master` branch has a commit titled "Get rid of Stream module, revert to 0.14 style Signals".

jvoigtlaender:
Yes, [this](https://github.com/elm-lang/core/commit/ec22bcc4818d2e806d80107d94060394392856fe) must be the commit you want. It has the exact same diff against my "fixing branch" as is shown in this PR here.

PR #210, opened 2015-04-10
cleanup in Time.js

jvoigtlaender:
Done in the aftermath of https://github.com/elm-lang/core/commit/9d52af2e1c651be2474d6df997ad7ecd6765058d.

Main difference: `fps` is already defined from `fpsWhen` in `Time.elm`, so there is no need anymore to have it in `Time.js`.

Something I noticed while going through this: somewhere along the way, `Time.delay` and `Time.since` seem to have vanished. Purposefully?

### DIFF
--- a/src/Native/Time.js
+++ b/src/Native/Time.js
@@ -12,7 +12,6 @@ Elm.Native.Time.make = function(localRuntime)
 	var Signal = Elm.Signal.make(localRuntime);
 	var NS = Elm.Native.Signal.make(localRuntime);
 	var Maybe = Elm.Maybe.make(localRuntime);
-	var Utils = Elm.Native.Utils.make(localRuntime);
 
 
 	// FRAMES PER SECOND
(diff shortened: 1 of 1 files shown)


evancz:
Great, thank you! I removed `delay` and `since` because I think it'll be better to recommend doing that kind of thing with `Task.sleep`. I'm not sure if that's true given that things go out ports. Are there things that definitely need those things right now?

jvoigtlaender:
Rate-limiting a signal, as useful e.g. for certain UI behavior. Like explained [here](http://underscorejs.org/#debounce). There was a relevant [thread](https://groups.google.com/d/msg/elm-discuss/2CHk-bE0Gpw/7Jw76C8LCwUJ) on the mailing list where we ended up with an Elm implementation based on `Time.since`. The need for such rate-limiting has come up more than once on the mailing list, so I'm pretty sure people would not be pleased to lose it. I don't immediately see how to do it with `Task.sleep`.

PR #214, opened 2015-04-12
Create a new date from a year, a month, a day, an hour, a minute, a second and a millisecond.

guyonvarch:
In a project for my company, I need to get the time corresponding to the begin of the day from a given time. Using Date.fromTime and the string representation to get the begin of the day fail because of locals, and passing from the string representation is a poor choice to get what I want I think.

Firefox and chrome behave differently with the following expression. Firefox take the local time while chrome take the UTC time, look at the output on my machine:

new Date("2015-04-10T00:00:00.000")
Firefox: Date 2015-04-09T22:00:00.000Z
Chrome: Fri Apr 10 2015 02:00:00 GMT+0200 (CEST)

I can't rely on Date.fromTime for my problem.

However, I found a solution with the following function :

  function beginOfDay(date) {
    var msPerDay = 24 * 60 * 60 * 1000;
    var offsetMs = new Date().getTimezoneOffset() * 60 * 1000;

    var ms = date.valueOf() - offsetMs;
    var beginMs = ms - (ms % msPerDay) + offsetMs;

    return new Date(beginMs);
  }

But it have to rely on getTimezoneOffset(), it is simpler with a newDate function that works with locals: 

new Date(2015, 04, 10, 0, 0, 0, 0)
Firefox: Date 2015-05-09T22:00:00.000Z
Chrome: Sun May 10 2015 22:00:00 GMT+0200 (CEST)

### DIFF
--- a/src/Date.elm
+++ b/src/Date.elm
@@ -9,7 +9,7 @@ module Date
 issues with internationalization or locale formatting.
 
 # Conversions
-@docs fromString, toTime, fromTime
+@docs fromString, newDate, toTime, fromTime
 
 # Extractions
 @docs year, month, Month, day, dayOfWeek, Day, hour, minute, second, millisecond
@@ -42,6 +42,13 @@ type Month
 fromString : String -> Result String Date
 fromString =
   Native.Date.read
+
+
+{-| Create a new date from a year, a month, a day, an hour, a minute, a second
+and a millisecond.
+-}
+newDate : Int -> Month -> Int -> Int -> Int -> Int -> Int -> Date
+newDate = Native.Date.newDate
 
 
 {-| Convert a date into a time since midnight (UTC) of 1 January 1990 (i.e.
--- a/src/Native/Date.js
+++ b/src/Native/Date.js
@@ -22,6 +22,12 @@ Elm.Native.Date.make = function(localRuntime) {
 			: Result.Ok(date);
 	}
 
+	function newDate(year, monthName, day, hour, minute, second, millisecond)
+	{
+		var month = monthTable.indexOf(monthName.ctor);
+		return new window.Date(year, month, day, hour, minute, second, millisecond);
+	}
+
 	var dayTable = ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"];
 	var monthTable =
 		["Jan", "Feb", "Mar", "Apr", "May", "Jun",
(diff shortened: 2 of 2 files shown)


jvoigtlaender:
Is still still relevant/needed? If so, maybe you can use functionality from http://package.elm-lang.org/packages/rluiten/elm-date-extra/latest/? Generally, the way towards adding new functionality to `core` is via such `...-extra` packages. So even if your needs are not currently addressed by the linked package, the best strategy would be to target it with a pull request (instead of `core`).

jvoigtlaender:
... except if you absolutely need this to be done on the native/JavaScript side. In that case, getting it realized in a non-`core` package would be (more) problematic.

guyonvarch:
What I need is to get the date corresponding to the begin or the end of the day.

This can be done with `new Date(…)` and number parameters, or with the `setHours` method, or finally with `new Date().getTimezoneOffset()` like I mentioned in the first post.

The only core function giving us the date is in the Date module: `fromString : String -> Result String Date`.

I found a way to get the same result with `new Date(str: String)`, but I still have to give the time zone, which I have not ?

Same result in chrome and firefox:

``` javascript
new Date("Fri May 10 2015 00:00:00 GMT+0200 (CEST)").getTime()
1431208800000

new Date(2015, 04, 10, 0, 0, 0, 0).getTime()
1431208800000
```

jvoigtlaender:
I don't get from your response whether or not what you want can be done with http://package.elm-lang.org/packages/rluiten/elm-date-extra/latest/ or as a pull request against that package. You describe what you would do in JavaScript, but the package I mention does already provide a wrapper around at least some of JavaScript's date API. I suggest you do look at that package and see how far it, or its approach, can get you. Without clear indication that that package can't possibly give you what you want, there is no strong case to make for adding functionality to `core`.

For context on this "policy", see https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions.

jvoigtlaender:
@guyonvarch, any response to what I said in my last comment above?

guyonvarch:
The function of this PR was recently added in elm-date-extra :
https://github.com/rluiten/elm-date-extra/commit/2e2c489f414f6ad6f888bdb6ded1723898c86d6d#diff-b171d4b828e1da35cd5803d9eb0ec412R178

I tested it on a day on firefox and chrome and got the same result as with `new Date(…)`. We can close this PR.

PR #217, opened 2015-04-18
Update Signal.elm

Tzbob:
Add docs for creating mailbox with an explicit note that it's an impure function.


evancz:
Nice, I did not realize there were no docs at all for this function! Added a bit more exposition and made the change!

PR #222, opened 2015-04-21
Fixed #221 - Image src URLs with spaces in them don't work for fittedImage and croppedImage

### LINKED ISSUE #221: Image src URLs with spaces in them don't work for fittedImage and croppedImage
In the Graphics.Element package, `fittedImage` and `croppedImage` don't work for all image URLs. 

For example, if an image URL is "somewhere/my cool photo.jpg", the image won't show up with `fittedImage` and `croppedImage`. It will, however, show up when you use "image" or "tiledImage", or if the image was named "somewhere/my-cool-photo.jpg" (without spaces).

srikumarks:
The problem was that the native implementations of `fittedImage` and
`croppedImage` used CSS `background-image`, but were not encoding the
URLs before generating the property. This commit encodes the `src`
parameter using `encodeURIComponent`, which makes them work correctly.

### DIFF
--- a/src/Native/Basics.js
+++ b/src/Native/Basics.js
@@ -132,6 +132,11 @@ Elm.Native.Basics.make = function(localRuntime) {
 		round: Math.round,
 		toFloat: function(x) { return x; },
 		isNaN: isNaN,
-		isInfinite: isInfinite
+		isInfinite: isInfinite,
+
+		encodeURI: encodeURI,
+		encodeURIComponent: encodeURIComponent,
+		decodeURI: decodeURI,
+		decodeURIComponent: decodeURIComponent
 	};
 };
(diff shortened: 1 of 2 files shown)


srikumarks:
The best way to deal with such string safety is via the type system. A `URL` type should be used instead of `String` for these image source arguments.

srikumarks:
The Travis failure, btw, is a failure to install elm-compiler-0.15.

evancz:
Can you show an example of a "good string" that can be given to these functions that causes an unexpected result?

srikumarks:
I'd noted a few in [my issue comment here](https://github.com/elm-lang/core/issues/221#issuecomment-96365683). For example, passing "hello).png" would generate "url(hello).png)" for the "background" property which fails to parse.

evancz:
I don't really understand what is going on here or what would be good, but the `Graphics.*` modules have moved to `evancz/elm-graphics` so it makes sense to retarget stuff like this.

evancz:
Oh, I get it from the last comment. The change in the one file makes sense, but the other one is weird.

Issue #223, opened 2015-04-22
Add Task.parallel


TheSeamau5:
This feature has been discussed previously in the mailing list, but the following thread brings a compelling case for this function: https://groups.google.com/forum/#!topic/elm-discuss/n-6Mh5KhF18

The idea behind a `parallel` function is to be able to simultaneously perform multiple requests (usually Http requests) and then collect a successful value only if all requests have succeeded or else collect an error value.

```elm
parallel : List (Task error value) -> Task error (List value)
-- alternatively, parallel : List (Task error value) -> Task (List error) (List value)
```
In essence, `parallel` would be analogous to `sequence`.

The use case that has come up from the mailing list is that you want to get a list of data (json objects) from the interwebs that you then wish to display. Except that you wish to display them progressively, as they come (as opposed to all at once or one by one). 

`parallel` would allow to batch requests and thus, rather that display items one by one, you could display them 20 by 20, or 5 by 5, etc... These requests would be sent in parallel, thus theoretically making things faster (in theory you could get 5 responses as fast as you get 1 response).

TheSeamau5:
See example [`parallel`](https://github.com/douglascrockford/RQ/blob/master/rq.js#L279) function in Douglas Crockford's [`RQ`](https://github.com/douglascrockford/RQ) library for async programming in JS.

Link to library: https://github.com/douglascrockford/RQ
Link to parallel function : https://github.com/douglascrockford/RQ/blob/master/rq.js#L279

TheSeamau5:
In working on the mailing list problem, I've stumbled on an implementation of `parallel` that is very useful and is entirely in Elm, using `spawn`. The difference is that it requires an `Address` to be passed to it so that it may do its thing.

```elm
parallel : Address a -> List (Task error a) -> Task error (List ThreadID)
parallel address tasks = 
  let
      sendToAddress task = spawn (task `andThen` send address) 
  in
      sequence (List.map sendToAddress tasks)
```

Raynos:
I've recently written a very efficient `parallelCollect()` function for doing N asynchronous things in parallel and collecting all the errors and results.

This implementation might help you ( https://gist.github.com/Raynos/ae3ac8aeb95c3b5ae956 ).

I paid special attention to allocating the minimum amount of closures necessary to be able to do this.

johnpmayer:
Hmm. I'd rather be able to write this as a regular task without Mailboxes
or itself as native code. Sounds like we need an actually Promise
[...]

PR #224, opened 2015-04-22
Added Task.parallel

TheSeamau5:
Added simple implementation for `Task.parallel`. This function will run
a lists of tasks in parallel and will return a task that computes a
list of ThreadIDs.

Solves this issue: https://github.com/elm-lang/core/issues/223

As @johnpmayer pointed out [here](https://groups.google.com/d/msg/elm-discuss/n-6Mh5KhF18/wDaFb0GR-2MJ), further down the road, we may want to get an `MVar` implementation a la Haskell in Elm. This would give more fine grain control and allow communication between "threads". Currently, those `ThreadID` values are completely useless (aside from debugging) but they could be put to good use.

But, in the meantime, this is not a bad option and it's worth adding to the core library due to how common this use case is.

### DIFF
--- a/src/Task.elm
+++ b/src/Task.elm
@@ -2,7 +2,7 @@ module Task
     ( Task
     , succeed, fail
     , map, map2, map3, map4, map5, andMap
-    , sequence
+    , sequence, parallel
     , andThen
     , onError, mapError
     , toMaybe, fromMaybe, toResult, fromResult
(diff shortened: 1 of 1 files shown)


evancz:
For the reasons, described in that thread [here](https://groups.google.com/d/msg/elm-discuss/n-6Mh5KhF18/fz58yFRqf8gJ), I'd like to wait on a proper implementation of this. For now, people can spawn a bunch of stuff or make a library. I think adding the wrong semantics to a core library is not a good thing to do.

TheSeamau5:
Cool. Anyways, if people need this sort of thing, there's a library:
http://package.elm-lang.org/packages/TheSeamau5/elm-task-extra/2.0.0
It has `parallel` and `optional` and a few others.

PR #229, opened 2015-04-28
Make Keyboard derived signals mergeable

jvoigtlaender:
E.g., in Elm-0.14, `merge Keyboard.arrows Keyboard.wasd` made sence, but
it doesn't currently in Elm-0.15, because `Keyboard.wasd` will *never* be
propagated. The "culprit" was [this commit](https://github.com/elm-lang/core/commit/46fb36ecb40d563a0cc877d073405bfd689ba3ca), which profoundly changed the semantics of these signals by omitting a `dropRepeats` call that previously happened on the JS side.

### DIFF
--- a/src/Keyboard.elm
+++ b/src/Keyboard.elm
@@ -141,6 +141,7 @@ toXY {up,down,left,right} keyCodes =
 -}
 arrows : Signal { x:Int, y:Int }
 arrows =
+  Signal.dropRepeats <|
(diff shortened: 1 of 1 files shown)


evancz:
Okay, I merged this in. Not sure what's going on with github about it, but  d5c1f49d0ed3757a5bea76fd2a2876cb145890d7 should do it.

Thanks for figuring out the issue here! I'll do a patch release for this.

Issue #234, opened 2015-05-01
Separate Path and Shape types?


jvoigtlaender:
The documentation gives the impression that

```elm
outlined : LineStyle -> Shape -> Form
```

and

```elm
traced : LineStyle -> Path -> Form
```

have different types. But they don't. Which can lead to "wrong applications", like for [this user](https://groups.google.com/d/msg/elm-discuss/Dk_pb0l_ns8/QnMAjNs5_Y4J).

hoosierEE:
Yeah, I think you're right about this.  The compiler should complain if someone gives `traced` a `circle`, or `outlined` a `segment`.  I experienced this myself, twice.  Once because I made the mistake, and twice because I forgot about the first time!

evancz:
I agree, I implemented this in a long long ago time when it wasn't clear whether type aliases would be distinct in some way. In any case, I think it makes sense to hide all the internal details by making both of them a `type`

jvoigtlaender:
Cool, I'll send a PR.

jvoigtlaender:
Closed via https://github.com/elm-lang/core/pull/235.

hoosierEE:
Nice, thanks!

PR #235, opened 2015-05-01
Turned Path and Shape into separate types

jvoigtlaender:
addresses https://github.com/elm-lang/core/issues/234

### DIFF
--- a/src/Graphics/Collage.elm
+++ b/src/Graphics/Collage.elm
@@ -129,8 +129,8 @@ dotted clr =
 
 
 type BasicForm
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Merge this? You (@evancz) agreed in #234 that this separation should be enforced.

jvoigtlaender:
Merged current master to make sure this still compiles.

evancz:
Thanks! This finally became a requirement with the new documentation tool.

Also, thanks for all the great PRs! The are always reasonable improvements with well-done PRs :)

PR #237, opened 2015-05-02
Added docu links between Char and Keyboard modules

jvoigtlaender:
Also, I wonder whether one of the two modules should import the other, so that the `KeyCode` type alias would be defined only in one place. (Imagine you change what type to use for `KeyCode` at some point for some reason. If you forget about one of the two places, the types of some functions suddenly get out of sync (so that library users cannot really use them together), without the compiler being able to warn you as the implementer of the library.)

### DIFF
--- a/src/Char.elm
+++ b/src/Char.elm
@@ -82,8 +82,8 @@ toLocaleLower =
 type alias KeyCode = Int
 
 
-{-| Convert to unicode. Used with the `Keyboard` library, which expects the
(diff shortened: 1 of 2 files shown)


jvoigtlaender:
Closing this, replaced by cleaner PRs https://github.com/elm-lang/core/pull/382 and https://github.com/elm-lang/core/pull/383 (that can be merged independently from each other).

PR #239, opened 2015-05-05
Fix error handling at jsToElm()

jinjor:
Hi,
I tried `port` and got an error when passing object from JavaScript-side to Elm-side.
```
Uncaught ReferenceError: arg is not defined
```
I modified that variable and now it seems to be working correctly.


jinjor:
I saw this request is already accepted by another PR, so close it.

PR #242, opened 2015-05-11
Update tests for 0.15

mgold:
This boils down to 3 things:
* Adding the `exposing` keyword multiple times per file
* Updating `elm-package.json` with the elm-version field
* Removing the tests for List.uncons which seems to have disappeared

I see that the travis build is in cabal hell, but at least you can manually run
the tests with `tests/run-test.sh` and they all pass.

### DIFF
--- a/tests/Test.elm
+++ b/tests/Test.elm
@@ -1,14 +1,14 @@
 module Main where
 
-import Basics (..)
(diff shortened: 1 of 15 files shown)


evancz:
Cool, thank you!

PR #243, opened 2015-05-12
Improve docs for Json.(En|De)code

mgold:
Improve docs for the Json libraries, most notably
* Rename a type alias Task that has nothing to do with 0.15 Tasks
* Give an example use for `Decode.tuple1`
* Fix assorted errors, clarify wording, add information
* Give a brief introduction to what a `Decoder a` is
* Note that Infinity and NaN serialize to null (which means encode and decode aren't quite inverses)

Hopefully an easy merge - if there's anything odd, please let me know.

### DIFF
--- a/src/Json/Decode.elm
+++ b/src/Json/Decode.elm
@@ -1,5 +1,6 @@
 module Json.Decode where
-{-| A way to turn Json values into Elm values.
+{-| A way to turn Json values into Elm values. A `Decoder a` represents a
+decoding operation that will either produce a value of type `a`, or fail.
 
 # Run a Decoder
 @docs decodeString, decodeValue
@@ -74,10 +75,11 @@ decodeString =
 really a helper function so you do not need to write `(:=)` so many times.
 
     -- object.target.value = 'hello'
-
     value : Decoder String
     value =
         at ["target", "value"] string
+
+It is defined as
 
     at fields decoder =
         List.foldr (:=) decoder fields
@@ -123,15 +125,15 @@ object2 =
     Native.Json.decodeObject2
 
 
-{-| Use two different decoders on a JS value. This is nice for extracting
+{-| Use three different decoders on a JS value. This is nice for extracting
 multiple fields from an object.
 
-    type alias Task = { task : String, id : Int, completed : Bool }
+    type alias Job = { name : String, id : Int, completed : Bool }
 
-    point : Decoder Task
+    point : Decoder Job
     point =
-        object3 Task
-          ("task" := string)
+        object3 Job
+          ("name" := string)
           ("id" := int)
           ("completed" := bool)
 -}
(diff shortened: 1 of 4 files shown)


mgold:
Rebased on to master and added a second commit fixing Graphic.Input's docs - previous they invoked `Signal.Mailbox` as a function when it should be `Signal.mailbox`, lowercase. CI should pass.

mgold:
Closing and reopening to force CI to rebuild; it seems to be a flaky download.

src/Json/Decode.elm
```diff
@@ -309,11 +306,16 @@ null =
     Native.Json.decodeNull
 
 
-{-| Great for handling optional fields. The following code decodes JSON
-objects that may not have a profession field.
+{-| Extract a Maybe value, wrapping successes with `Just` and turning any
+failure in `Nothing`. Great for handling fields that may be missing or null. The
```
evancz: This is actually not that great for things that may be `null` because it actually admits all sorts of errors. A decoder that can be null or something, is different than a decoder that is something or anything. If the given decoder messes up, it'll give `Nothing` (corresponding to `null` which it is not) rather than failing.

src/Json/Encode.elm
```diff
@@ -55,6 +55,8 @@ int =
     Native.Json.identity
 
 
+{-| Encode a Float. `Infinity` and `NaN` are encoded as `null`.
+-}
```
evancz: What?

evancz:
Thanks! Besides the two comments, this looks really great!

src/Json/Encode.elm
mgold:
````
> import Json.Encode as E
> E.encode 0 (E.float (1/0))
"null" : String
> E.encode 0 (E.float (0/0))
"null" : String
```
Personally I'm fine with this behavior. I like how encoding always succeeds.

src/Json/Decode.elm
mgold: That's a valid concern and I'll revisit this when I get a chance.

src/Json/Encode.elm
evancz: Shouldn't Elm's floats map on to the same floats though? Like, why would NaN not result in Nan? Is that not valid JSON or something?

mgold: > Numeric values that cannot be represented as sequences of digits (such as Infinity and NaN) are not permitted. [[page 9](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf)]

````
$ node
> JSON.stringify(NaN)
'null'
> JSON.stringify(Infinity)
'null'
```

mgold:
New commit should address the null issue.

mgold:
Thanks!

PR #247, opened 2015-05-15
Fix broken error message

rehno-lindeque:
The runtime reports `"arg is not defined"` instead of the proper error message due to the value being propagated differently after the release of 0.15.

### DIFF
--- a/src/Native/Port.js
+++ b/src/Native/Port.js
@@ -70,7 +70,7 @@ Elm.Native.Port.make = function(localRuntime) {
 				"Regarding the port named '" + name + "' with type:\n\n" +
 				"    " + type.split('\n').join('\n        ') + "\n\n" +
 				"You just sent the value:\n\n" +
-				"    " + JSON.stringify(arg.value) + "\n\n" +
+				"    " + JSON.stringify(value) + "\n\n" +
 				"but it cannot be converted to the necessary type.\n" +
 				e.message
 			);


evancz:
Great, thank you!

madsflensted:
Any chance this will be part of a patch release soon? Seems minor, but really blocks the helpful feedback from the port verification code, and turns a bad input type into a "say what?"-error.

rehno-lindeque:
It's a great big hack, but can apply this locally to the core elm package if you want by the way (I did this because we needed it badly). Would be nice to have bug fixes published in a rolling fashion though.

evancz:
It looks like I can do a MINOR release with a few rollbacks.

I think this is the second or third time that patches got held up on bigger changes, so I think it makes sense to start using the `stable` branch properly again. Any patch changes go directly there and get merged into master. We merge master into stable at more strategic times. Does that sound plausible?

If I get stable up to date with the latest public release, can someone cherry pick this over? Then we can do a PATCH no problem.

evancz:
Okay, 2.1.0 should be out with these changes.

madsflensted:
:+1:

rehno-lindeque:
Thanks @evancz! I only rarely submit PR's but I'm happy to PR against a stable branch after being accepted into master from now on if that sounds like a good idea to you.

evancz:
No problem, that sounds good to me. And please let me know if I should do a patch release on things! By default I'll sync things up with the larger release cycle.

Issue #248, opened 2015-05-20
Touch events triggered by mouse click+move are mutable


grumpyjames:
Hi,

I have been exploring the possibility of writing a little gestures library on top of `Touch`. Initial testing, by tapping using macbook pro touchpad, proved rather tricky - all the touch events I generated had the same co-ordinates.

The following program reproduces the issue: run it, 'click' (and hold) with the touchpad, and wiggle your finger around. As the pointer moves, all of the events from the past change to match the latest touch's (x,y).

<pre>
main = S.map show <| foldp (::) [] touches
</pre>

I think this is caused by lines 115/116 in Native/Touch.js

<pre>
                root.value[i].x = point._0;
                root.value[i].y = point._1;
</pre>

...I think that should probably use the nearby 'touch' function. I'll put together a pull request to fix, just wanted to drop this in here in case I am utterly wrong (please let me know if so!)

evancz:
Sorry we did not get to this earlier! The touch library isn't in core anymore though, so in a sense, this problem is solved. We should be getting touch support back in time!

PR #249, opened 2015-05-20
Avoid mutation of already notified touch events when mousemoving.

grumpyjames:
Instead, replace the appropriate touch event with a new one with updated XY.

This should fix issue #248.


grumpyjames:
Looks like the underlying code is in transit anyway; I'll close this.

Issue #251, opened 2015-05-23
There is no O(1) way of checking if a dict is nonempty


imeckler:
You could do `Dict.foldl (\_ _ _ -> True) False`, but this is O(n).

lachenmayer:
Looking at the [source](https://github.com/elm-lang/core/blob/2.0.1/src/Dict.elm), it seems like `isEmpty x = x == Dict.empty` should always work.

The reason why the documentation states that "Dictionary equality with (==) is unreliable and should not be used" is that a Dict is a self-balancing red-black tree, and the distribution of red and black nodes depends on the insertion order.
That being said:

`Dict.empty` is defined as `RBEmpty LBlack`.
From a non-empty tree, the only way to arrive at an empty Dict is to remove an element.
Every `remove` operation performs `blacken` ([here](https://github.com/elm-lang/core/blob/2.0.1/src/Dict.elm#L216)).
`blacken` will result in `RBEmpty LBlack` if the tree is empty ([here](https://github.com/elm-lang/core/blob/2.0.1/src/Dict.elm#L370)).

I've only glanced at the code briefly, do let me know if you find a counter-example.
You're right though, Dict should have an `isEmpty` function, since `List` and `String` both contain one.

evancz:
I think the reason this function does not exist is because I wanted a uniform API across all the containers, but wasn't sure what that name should be. I think `isEmpty` is a nice way to go, so I think it's cool to add that as long as we do it in `List`, `Dict`, `Set`, and `Array`.

jvoigtlaender:
I think this issue can be closed.

PR #253, opened 2015-05-26
O(1) methods for Array.isEmpty, Dict.isEmpty, and Set.isEmpty

AlexNisnevich:
See @evancz's comments on #251.


evancz:
Great, thanks @AlexNisnevich and @jvoigtlaender!

PR #254, opened 2015-05-27
Simplify Dict code

jvoigtlaender:
* removing redundant bindings
* inlining/simplifying case branches
* removing two unused, and unexported, functions


jvoigtlaender:
Ah, sorry, the last of these simplification steps were too quick. So I'll revise.

The test suite / Travis went green, though. So test coverage of `Dict` does not seem to be very deep.

PR #256, opened 2015-05-28
Infix operator for exponentiation should be right-associative

ThomasWeiser:
I don't think that anyone will ever use an expression like `3^3^3`,
but it's nicer to follow the general convention.
Otherwise it would look like an oversight.

### DIFF
--- a/src/Basics.elm
+++ b/src/Basics.elm
@@ -126,7 +126,7 @@ infixl 6 +
 infixl 6 -
 infixl 7 *
 infixl 7 /
-infixl 8 ^
+infixr 8 ^
 
 infixl 7 //
 infixl 7 %


evancz:
I saw you mentioned that this is the standard in other languages. Do you mind sharing a few links just for the record? Otherwise this looks great to me, thank you for finding this and sending a PR!

ThomasWeiser:
Haskell:
- [Precedences and fixities of prelude operators](https://www.haskell.org/onlinereport/decls.html#prelude-fixities) (scroll up a bit)
- Prelude: [`(**), logBase :: a -> a -> a infixr 8`](https://hackage.haskell.org/package/base-4.8.0.0/docs/Prelude.html)

[F# operator `**`](https://msdn.microsoft.com/en-us/library/dd233228.aspx)

[Python operator `**`](http://stackoverflow.com/questions/2383034/what-is-the-associativity-of-pythons-operator)

[R](http://cran.r-project.org/doc/manuals/r-release/R-lang.html#Infix-and-prefix-operators): The exponentiation operator `^` [...] group right to left

And a [proposal for Javascript/ES7](http://bocoup.com/weblog/designing-an-exponentiation-operator-for-javascript/)

evancz:
Great, thank you!

mgold:
I doubt anyone is relying on this, but isn't it a backwards-incompatible change?

evancz:
Yes, it is. I have an item on my todo list of "fixity changes are breaking changes in elm-package" but it is not implemented right now. Not sure what I'm going to do. It may be that the explicit exports that @jvoigtlaender added recently will "remove" some functions causing a major version change anyway.

Issue #236, opened 2015-05-02
Missing export lists


jvoigtlaender:
The following modules are currently missing export lists:
* https://github.com/elm-lang/core/blob/master/src/Graphics/Input.elm
* https://github.com/elm-lang/core/blob/master/src/Graphics/Input/Field.elm
* https://github.com/elm-lang/core/blob/master/src/Json/Decode.elm
* https://github.com/elm-lang/core/blob/master/src/Json/Encode.elm

Is it really intended that all internals of those libraries are part of the external API (and thus potentially problematic to change later on)?

jvoigtlaender:
I could have a go at adding these export lists. Should I?

PR #257, opened 2015-05-30
Adding missing export lists

jvoigtlaender:
Exporting all the entities from `Graphics.Input`, `Graphics.Input.Field`, `Json.Decode`, and `Json.Encode` that are part of the documented API, i.e., all the entities that are in those files except for the "internal representations" of `Value` und `Decoder`.

Addresses https://github.com/elm-lang/core/issues/236.


evancz:
Looks good to me, thank you!

Issue #175, opened 2015-02-16
Array.repeat causes a runtime error


TheSeamau5:
Calling `Array.repeat` with a negative size causes a runtime error.

```
Invalid array length

Open the developer console for more details.
```

This is the code that has caused this:

```
import Array 
import Text (asText)

main = 
  asText <|
    Array.repeat -5 4
```

This behavior is not the same as the behavior with lists as calling `List.repeat` with a negative size yields the empty list.

mgold:
@TheSeamau5 I think you can close this now.

sindikat:
This issue still exists in Elm 0.15.1, just paste this in http://elm-lang.org/try :

    import Array
    import Graphics.Element exposing (show)
    main = show <| Array.repeat (-5) 0

mgold:
The patch was merged after the latest version of core was released. You'll have to wait for 0.16, sorry.

PR #258, opened 2015-06-03
Fix runtime error creating arrays of negative length

mgold:
This one-character change (plus test) fixes a runtime error when creating arrays of negative length with `Array.initialize` or `Array.repeat`. This behavior is now consistent with `List.repeat`.


mgold:
This will fix #175.

PR #261, opened 2015-06-04
Small code improvement in Dict.elm

jvoigtlaender:
I guess this makes it  more obvious what that code is doing, and is also more in line with how other non-auxiliary functions in this module are written.

Of course, semantically there is no change, due to this definition: https://github.com/elm-lang/core/blob/2.0.1/src/Dict.elm#L82


evancz:
Makes sense to me!

Issue #260, opened 2015-06-04
Make `Set a` and `Dict a ()` be different types?


jvoigtlaender:
Currently they are not. Which means that expressions like `Dict.get 5 (Set.singleton 7)` etc. are type-correct, even though:
* they make no sense (the particular given one makes specifically little sense, since `Set` does not itself have a `get` function)
* the documented APIs suggest that these expressions should not be valid (types of `Dict.get` and `Set.singleton` are completely incompatible, as far as a reader of http://package.elm-lang.org/packages/elm-lang/core/latest/Dict and http://package.elm-lang.org/packages/elm-lang/core/latest/Set is concerned).

jvoigtlaender:
BTW, I conjecture `Set a` and `Dict a ()` are currently only accidentally not distinguished, similarly as in #234, where Evan indicated that at same point in the past it wasn't clear whether type aliases (which are at play here) would behave like Haskell's `type` or like Haskell's `newtype`.

jvoigtlaender:
Closed via https://github.com/elm-lang/core/pull/263.

evancz:
Oh, also, this means that the `toString` function is messed up for now. We may want to give the internal constructor a weirder name when we fix that!

jvoigtlaender:
So currently it will say something like `Set (Dict.fromList [(0,())])` instead of `Set.fromList [0]`? Well, probably better to fix that, yes. Though I must say looking at the implementation of `toString` in the native JS code right now, I'm a bit scared overall. I mean, it seems that if somebody decided to have a constructor named `RBNode` in their own union type, results could be quite catastrophic?

jvoigtlaender:
Indeed, this leads to a runtime error:
```elm
import Graphics.Element

type V = RBNode

v = RBNode

main = Graphics.Element.show v
```

PR #263, opened 2015-06-05
Make `Set t` a type different from `Dict t ()`

jvoigtlaender:
Addresses https://github.com/elm-lang/core/issues/260.

### DIFF
--- a/src/Set.elm
+++ b/src/Set.elm
@@ -1,4 +1,3 @@
-
 module Set
     ( Set
     , empty, singleton, insert, remove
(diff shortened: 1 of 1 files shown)


mgold:
This will force core to version 3.0.0. I'm not sure we're there yet. Semvar aside, looks good.

Incidental: maybe we need a `breaking` branch so we could merge this now and forget about it until the next major release?

jvoigtlaender:
This here is, of course, another case like the ones in #283.

evancz:
With the new docs format, I think we need to fix all of these cases. Thanks for the PR!

PR #264, opened 2015-06-06
Improve documentation of `Debug.watch`

jvoigtlaender:
Specifically, it is tempting to call `Debug.watch` directly on a signal, which is not useful though. That can lead to puzzlement, since it then seems as if `Debug.watch` serves no purpose. See [this question on the mailing list](https://groups.google.com/d/msg/elm-discuss/0fjuzWVQvOA/qxvar9TBKnAJ). The expanded documentation should make clear now that one needs to map `Debug.watch` *into* the signal if one wants to observe the value carried on the signal.

### DIFF
--- a/src/Debug.elm
+++ b/src/Debug.elm
@@ -56,12 +56,18 @@ crash = Native.Debug.crash
 
 {-| Watch a particular value in the debugger. Say we want to know the value of
 a variable called `velocity` because it may not be updated correctly. Adding
-`Debug.watch` allows us to name the value and show it with the debugger.
+`Debug.watch` allows us to name the value and show it with the debugger. The
+result of evaluating such an expression is exactly the same as if not having
+the call to `Debug.watch` at all.
 
 	  Debug.watch "velocity" velocity == velocity
 
-Notice that the result of evaluating this expression is exactly the same as
-not having the expression at all. That means it's easy to add to any value.
+That means it's easy to add `Debug.watch` to any value.
+
+Note that calling `Debug.watch` on a signal is not useful. Instead, it needs
+to be mapped into the signal (to act on the contained value). So if you want
+to watch a timer signal, instead of `Debug.watch "time" <| Time.every second`
+you need `Debug.watch "time" <~ Time.every second`.
 -}
 watch : String -> a -> a
 watch = Native.Debug.watch
--- a/src/Debug.elm
+++ b/src/Debug.elm
@@ -57,8 +57,7 @@ crash = Native.Debug.crash
 {-| Watch a particular value in the debugger. Say we want to know the value of
 a variable called `velocity` because it may not be updated correctly. Adding
 `Debug.watch` allows us to name the value and show it with the debugger. The
-result of evaluating such an expression is exactly the same as if not having
-the call to `Debug.watch` at all.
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
I wonder whether it would make sense to even make it illegal to call `Debug.watch` on something of `Signal`-type. Could be made a runtime error. Should be unproblematic since `Debug.watch` shouldn't be used in production anyway. And calling it on a signal is never what you want (because it has no effect). So the debugger would save you some trouble if it simply errored out (with a helpful message) if you accidentally call `Debug.watch` on a `Signal`-typed thing.

jvoigtlaender:
Same (concerning the proposed runtime error) for `Debug.watchSummary`, of course.

src/Debug.elm
```diff
@@ -56,12 +56,18 @@ crash = Native.Debug.crash
 
 {-| Watch a particular value in the debugger. Say we want to know the value of
 a variable called `velocity` because it may not be updated correctly. Adding
-`Debug.watch` allows us to name the value and show it with the debugger.
+`Debug.watch` allows us to name the value and show it with the debugger. The
+result of evaluating such an expression is exactly the same as if not having
+the call to `Debug.watch` at all.
```
mgold: Remove "if". Also, "the result of evaluating such an expression" is a bit technical. Maybe just "the value is returned unchanged"?

src/Debug.elm
```diff
@@ -56,12 +56,18 @@ crash = Native.Debug.crash
...
 
 	  Debug.watch "velocity" velocity == velocity
 
-Notice that the result of evaluating this expression is exactly the same as
-not having the expression at all. That means it's easy to add to any value.
+That means it's easy to add `Debug.watch` to any value.
+
+Note that calling `Debug.watch` on a signal is not useful. Instead, it needs
+to be mapped into the signal (to act on the contained value). So if you want
+to watch a timer signal, instead of `Debug.watch "time" <| Time.every second`
```
mgold: Both `every` and `second` are in the `Time` module so it doesn't make much sense to have one qualified and one not. For the sake of explanation I would drop the `Time.` part.

src/Debug.elm
```diff
@@ -56,12 +56,18 @@ crash = Native.Debug.crash
...
+the call to `Debug.watch` at all.
 
 	  Debug.watch "velocity" velocity == velocity
 
-Notice that the result of evaluating this expression is exactly the same as
-not having the expression at all. That means it's easy to add to any value.
+That means it's easy to add `Debug.watch` to any value.
+
+Note that calling `Debug.watch` on a signal is not useful. Instead, it needs
+to be mapped into the signal (to act on the contained value). So if you want
```
mgold: > mapped into
onto? Maybe that's just me.

mgold:
Comments inline; hopefully I'm not too nitpicky. I like where this is going.

jvoigtlaender:
I reformulated a few bits. I do think that "mapped into" is better here.

mgold:
I won't fight you on it then. Should be good for whenever Evan gets back from his world tour.

evancz:
Looks great to me, thank you @jvoigtlaender!

PR #267, opened 2015-06-06
Correct behavior of `toString`

jvoigtlaender:
... on signals, should say "\<Signal\>", not "\<Stream\>".

Much as I regret that we don't have separate signal and stream concepts: As long as we don't, messages to the programmer/user should name the concept correctly. Currently, if you run
```elm
import Graphics.Element

main = Graphics.Element.show (Signal.constant 42)
```
you get output "\<Stream\>". But it should be (and before 0.15, was): "\<Signal\>". Streams are not a thing, currently.


evancz:
Thank you for the fix!

PR #270, opened 2015-06-11
Format port errors for undefined values

rtfeldman:
**Before:** (note the very last line)
```
Uncaught Error: Port Error:
Regarding the port named 'setParagraphs' with type:

    List Quiz.Question.Paragraph

You just sent the value:

    undefined

but it cannot be converted to the necessary type.
Runtime error when sending values through a port.
Expecting an array but was given 
```

**After:** (note the difference in the very last line)

```
Uncaught Error: Port Error:
Regarding the port named 'setParagraphs' with type:

    List Quiz.Question.Paragraph

You just sent the value:

    undefined

but it cannot be converted to the necessary type.
Runtime error when sending values through a port.
Expecting an array but was given undefined
```

### DIFF
--- a/src/Native/Utils.js
+++ b/src/Native/Utils.js
@@ -325,7 +325,8 @@ Elm.Native.Utils.make = function(localRuntime) {
 		throw new Error('Runtime error in module ' + moduleName + ' (' + span + ')' + msg);
 	}
 
-	function formatValue(value) {
+	function formatValue(value)
+	{
 		// Explicity format undefined values as "undefined"
 		// because JSON.stringify(undefined) unhelpfully returns ""
 		return (value === undefined) ? "undefined" : JSON.stringify(value);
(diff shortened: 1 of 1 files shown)


src/Native/Utils.js
```diff
@@ -325,12 +325,17 @@ Elm.Native.Utils.make = function(localRuntime) {
 		throw new Error('Runtime error in module ' + moduleName + ' (' + span + ')' + msg);
 	}
 
+	function formatValue(value) {
```
ajhager: I would bring the bracket down to the next line to match the style of the file, or even just remove the `formatValue` function and inline the check. Otherwise :+1: LGTM!

evancz: Agree about this, one way or the other. After the change, @rtfeldman, please ping this thread for merge :)

rtfeldman: Fixed in 4aca9b3cddf71ed70fb911d7d6a6691943ac1124

rtfeldman:
@evancz this should be ready to merge now!

evancz:
Looks good, thanks!

PR #276, opened 2015-06-13
Add currentDate as a task in Date.elm

ThomasWeiser:
We have Date and we have Task, so IMHO we should have a task to get the current date.

This PR implements
```elm
Date.currentDate : Task () Date
```

Of course it could be named differently or could be put into another module. The chosen ones are just my preference.

I didn't write a test within the core library as there are no tests for tasks up to now.
But I did test the code for my own and it works like expected.

### DIFF
--- a/src/Native/Date.js
+++ b/src/Native/Date.js
@@ -28,7 +28,7 @@ Elm.Native.Date.make = function(localRuntime) {
 		["Jan", "Feb", "Mar", "Apr", "May", "Jun",
 		 "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
 
-	currentDate = Task .asyncFunction (
+	var currentDate = Task .asyncFunction (
 		function (callback) {
 			callback (Task.succeed (dateNow ()));
 		}
--- a/src/Native/Date.js
+++ b/src/Native/Date.js
@@ -8,6 +8,7 @@ Elm.Native.Date.make = function(localRuntime) {
 	}
 
 	var Result = Elm.Result.make(localRuntime);
+	var Task = Elm.Native.Task.make (localRuntime);
 
 	function dateNow()
 	{
@@ -27,6 +28,12 @@ Elm.Native.Date.make = function(localRuntime) {
 		["Jan", "Feb", "Mar", "Apr", "May", "Jun",
 		 "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
 
+	currentDate = Task .asyncFunction (
+		function (callback) {
+			callback (Task.succeed (dateNow ()));
+		}
+	);
+
 
 	return localRuntime.Native.Date.values = {
 		read    : readDate,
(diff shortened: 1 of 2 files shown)


src/Native/Date.js
```diff
@@ -27,6 +28,12 @@ Elm.Native.Date.make = function(localRuntime) {
 		["Jan", "Feb", "Mar", "Apr", "May", "Jun",
 		 "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
 
+	currentDate = Task .asyncFunction (
```
rtfeldman: seems unlikely that this compiles... :wink:

dnalot:
Why a Task?

src/Native/Date.js
ThomasWeiser: Thanks, this should read `var currentDate = ...`, so I missed the `var`.
Or do you have other concerns?

ThomasWeiser:
Here is a simple test app:

```elm
import Html
import Date exposing (Date)
import Time
import Task

port execute : Task.Task () ()
port execute =
  Date.currentDate
  `Task.andThen`
  \date -> Signal.send dateInput.address date

dateInput : Signal.Mailbox Date
dateInput = Signal.mailbox (Date.fromTime 0)

main = Signal.map
  (\date -> Html.text <| toString (Date.toTime date))
  dateInput.signal
```

ThomasWeiser:
@dnalot 

> Why a task?

How would you provide the ability to get the current date? Maybe I am missing something.

Granted, there is already `Time.timestamp`, which does the same in principal, but
- it's not documented that timestamps actually represent real time/date.
- it's cumbersome to define an extra signal when you need the current date in a task (e.g. send the local time to a server).

IMHO, the core libarary should provide both,
- `timestamp` as a signal function for needs within the pure Elm-Architecture
- `currentDate` as a task for needs in the imperative task-based world

dnalot:
Ha I forgot time isn't referentially transparent. Your implementation looks straightforward! It seems essential for debugging and countdown timers. What's your use case? Feedback: functions generally shouldn't restate their [package name](http://package.elm-lang.org/help/design-guidelines#naming). I like JS's `Date.now` but maybe there are alternatives.

ThomasWeiser:
Thanks @dnalot, I renamed the function to `Date.current` (but `Date.now` would be equally handy).

dnalot:
`current` is only slightly odd being an adjective. @evancz recommended in #303 vetting new functionality in separate libraries. I didn't see a good fit in [existing packages](http://package.elm-lang.org/packages) or [unpublished repos](https://github.com/search?q=date+language%3AElm&type=Repositories) but it sounds like a superb idea for a new `elm-date-extra`! @rehno-lindeque might have something to add.

evancz:
To get this done in a timely manor, I think it makes sense to make a package for time that has nice things in it. We will do a systematic review of these libraries at some point and bring them into core. (plan outlined [here](https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions))

evancz:
Sorry for the long delay on getting you a response @ThomasWeiser. It is hard to juggle all the stuff going on these days!

rehno-lindeque:
An `-extra` package sounds like a fine idea. @ThomasWeiser we have a bunch of them here https://github.com/circuithub?utf8=%E2%9C%93&query=elm if you're looking for any examples

PR #279, opened 2015-06-15
Add fromOk to extract value from Ok with default

jivagoalves:


jvoigtlaender:
The process toward adding functions like this to `core` is to first collect them in `*-extra` packages. These will later be revisited as a whole, and decisions be made about what gets into `core` proper.

So I am closing this here. You may want to make it a pull request targetted at https://github.com/circuithub/elm-result-extra (if that doesn't already have this function).

PR #284, opened 2015-07-01
Use console.error when logging error messages

rtfeldman:
Browsers use this to show the console message in red, which makes it easier to spot when you have a chatty console going.


rtfeldman:
Chrome also adds a dropdown to `console.error` messages so you can view a stack trace if you like.

evancz:
It seems that this works with quite a lot of browsers, and we'd have issues on node either way. So this is fine with me, thanks!

PR #289, opened 2015-07-01
Fix indentation errors in Random module

mgold:
The style changes in ddd74cbd8f2 broke the build with an indentation error on line 257.

Line 208 was syntactically valid before, but I've never seen let-bindings that way and it conflicts with line 257 as well as other places in the codebase.


evancz:
I fixed this the other day, but it didn't make it to master yet. Thanks for the fix!

mgold:
Okay, no problem.

PR #291, opened 2015-07-10
Logic issue in tests for Char

zeckalpha:
`assertEqual (False) (List.all ...)` -> `assertEqual (False) (List.any ...)`


evancz:
Makes sense, thanks!

PR #294, opened 2015-07-14
Fix runtime error when testing equality of Json.Encode.null values

ThomasWeiser:
Function `eq` in Utils.js is currently only suitable for normal Elm types, not for arbitrary (speak `Json.Encode`) values.

As a result, comparing `Json.Encode.null`with anything non-null throws a runtime error.

Example:
```elm
Json.Encode.null == Json.Encode.object []
```

This PR fixes these cases.

Unit-tests are added to tests/Test/Equality.elm

### DIFF
--- a/src/Native/Utils.js
+++ b/src/Native/Utils.js
@@ -23,8 +23,12 @@ Elm.Native.Utils.make = function(localRuntime) {
 			{
 				continue;
 			}
-			if (typeof x === 'object')
+			if (typeof x === 'object' && typeof y === 'object')
 			{
+				if (x === null || y === null)
+				{
+					return false;
+				}
 				var c = 0;
 				for (var i in x)
 				{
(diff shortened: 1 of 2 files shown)


ThomasWeiser:
Rebased to current master.
Ready for being merged for upcoming relase 3.0.0

evancz:
6f5fbb8bc1ba2ce0fd3a3acf41e347c447689022 does a check like this, and 1c0077728a24859baaceb6040e0d97229319922b makes a note that it will not be possible to check for equality on `Json.Encode.Value` once a proper fix is implemented. That scenario is fundamentally problematic because a `Json.Encode.Value` can come through ports, meaning it can have functions.

This issue blocked because it was not clear what the ultimate fix would be. Now that it is clearer, I am fine with doing the stopgap.

PR #297, opened 2015-07-15
Specify drawing order in `collage`, etc

maxsnew:


jvoigtlaender:
@evancz, this seems useful information to add to the documentation. Merge?

evancz:
Makes sense, thanks @jvoigtlaender for the ping and @maxsnew for the change!

Issue #216, opened 2015-04-15
Add (?) to Maybe?


evancz:
With lots of core functions becoming total, it makes sense to consider an infix version of `withDefault`.

```haskell
(?) : Maybe a -> a -> a

firstNumber =
    head numberList ? 0
```

If you want to get the head of a list of numbers *or* just go with zero if it is empty. This is really cool, but here are two main concerns:

  1. Are we adding too many infix operators?
  2. What precedence this operator should have?

Can you share some examples that make the case for adding this function? Do you have any examples that help address concerns 1 or 2?

TheSeamau5:
I'll try to help give context for 1) by listing all the non-mathy/non-logic-y infix operators currently in Elm:

- `::` (cons)
- `|>` (apply right)
- `<|` (apply left)
- `>>` (compose right)
- `<<` (compose left)
- `<~` (Signal.map)
- `~` (Signal.mapN)
- `++` (List.append or String.append)
- `:=` (Json.Decode.decodeField)

evancz:
I just saw that something similar seems to exists [in C#](http://volatileread.com/Wiki/Index?id=2104). What other languages have something like this?

kasbah:
[Coffeescript](http://coffeescript.org/#overview) (see `#Existence: `)

TheSeamau5:
There's a long list here: https://en.wikipedia.org/wiki/Null_coalescing_operator

Apparently it's called the "null coalescing operator". 

They don't all agree on the exact syntax.  But here's a list of the versions I got from just skimming through the wiki page:

- `?`
- `??`
- `:?`
- `?:`
- `//`

Usually, there's an avoidance of the first one I'm guessing because of the existence of the ternary  operator `x == null ? x : 0` in those languages. But Elm doesn't have the ternary operator, so it can just pick whatever. :D

kasbah:
Maybe it could be:
```elm
(?) :: Maybe a -> Bool

(?:) :: Maybe a -> a -> a
```

which would looke like this in use:

```elm
if (head xs)? then f xs else g 

a = head xs ?: x

```

VulumeCode:
To address point 2, if it's supposed to chain you should take that into account:

    head numberList1 ? head numberlList2 ? 0

And you could consider it for record access:

    task : Maybe Task
    task?.description : Maybe String

Works nicely together:

    task?.description ? "" : String

mitchmindtree:
[...]

PR #299, opened 2015-07-16
Added maybe function

JustusAdam:
I'd really like to get the maybe function back. In Haskell I tend to use it a ton.

### DIFF
--- a/src/Maybe.elm
+++ b/src/Maybe.elm
@@ -1,4 +1,4 @@
-module Maybe (Maybe(Just,Nothing), andThen, map, withDefault, oneOf) where
+module Maybe (Maybe(Just,Nothing), andThen, map, withDefault, oneOf, maybe) where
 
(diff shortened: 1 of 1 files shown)


dnalot:
I think, 
```elm
map (+1) (fromString x) `withDefault` 0
```
reads better than, 
```elm
maybe (+1) 0 (fromString x) -- flipped order
```
for Elm's target user coming from JS. I bet it already exists in [another package](http://package.elm-lang.org/packages) though?

dnalot:
Check out `Maybe.Extra`'s [default operator](http://package.elm-lang.org/packages/circuithub/elm-maybe-extra/1.3.0/Maybe-Extra): 
```elm
map (+1) (fromString x) ? 0
```

evancz:
I think `map ((+) 1) maybe ? 0` is preferable over this. We have #216 to track that, but as @dnalot says, you can use it in `Maybe.Extra` until we make a decision about moving it into core.

PR #301, opened 2015-07-19
Changed documentation from [Int] to List Int

irrwitz:
IIRC [Int] was the old syntax and has been changed to List Int


dnalot:
Future idea (a if the list weren't long enough): verify code in package docs would compile in the REPL

evancz:
Nice find @irrwitz, thanks for the fix!

PR #303, opened 2015-07-21
Add List.find

rtfeldman:
This adds `List.find`, a useful function I've been using in a personal `Util.elm` file, but which seems like a good candidate for inclusion in `core` given that it's standard in several languages, for example:

* [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)
* [Ruby](http://apidock.com/ruby/Enumerable/find)
* [Scala](http://www.scala-lang.org/api/2.11.5/index.html#scala.collection.immutable.List@find(p:A=>Boolean):Option[A])
* [Haskell](https://hackage.haskell.org/package/base-4.7.0.1/docs/src/Data-List.html#find)

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -40,7 +40,7 @@ The current sentiment is that it is already quite error prone once you get to
 @docs foldr, foldl
 
 # Special Folds
-@docs sum, product, maximum, minimum, all, any, scanl
+@docs sum, product, maximum, minimum, all, any, find, scanl
 
 # Sorting
 @docs sort, sortBy, sortWith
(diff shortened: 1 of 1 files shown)


dnalot:
This should proxy to Native right? I've been wondering whether there's criteria for Core inclusion.

rtfeldman:
Would proxying to Native improve this though? With the impending self-TCO this compiles to a `while` loop anyway, yeah?

dnalot:
Native JS uses C++ cheats.

dnalot:
PS I agree it should be included I just wish we had a criteria for completeness and rejection.

evancz:
For now the recommended way of extending the standard libraries is in a package, like [elm-list-extra](http://package.elm-lang.org/packages/circuithub/elm-list-extra/latest) or [elm-random-extra](http://package.elm-lang.org/packages/TheSeamau5/elm-random-extra/latest). There can be lots of experimentation with APIs there that does not block on me or the release cycle of `elm-lang/core`. At some point, we will go through and try to merge in all the best things about the `*-extra` world in a coherent way.

I suspect @rehno-lindeque would accept this function in elm-list-extra, try a PR there for now.

dnalot:
Thanks for the clarification @evancz makes sense to me :+1:

rtfeldman:
Fair enough! Moved the PR to `elm-list-extra`.

Issue #240, opened 2015-05-06
Asynchronous loops created with Tasks and Signal.mailbox trigger crash in stepTask


pchiusano:
Here is [a minimal gist demonstrating the problem and showing a workaround](https://gist.github.com/pchiusano/0c6697e0c5981e576f55). Clicking rapidly generates:

```
Uncaught TypeError: Cannot read property 'status' of undefined
```

pointing to [this line](https://github.com/elm-lang/core/blob/master/src/Native/Task.js#L163) in `Task.js` `stepTask`.

The problematic pattern seems to be signals with 'asynchronous loops', whose output generates asynchronous tasks that affect the future value of that signal. I worked around it with an intermediate 'relay' mailbox, but that seems pretty kludgy.

[Background discussion on the mailing list](https://groups.google.com/forum/#!topic/elm-discuss/0WWYFocT7T8)

srid:
FWIW, I hit this bug in my app and had to workaround it (see the two commits linked above).

Dobiasd:
fyi, I also bumped into this problem in one of my apps. Luckily it happens with a task, that does not harm if it fails now and then. :) So I guess I will just leave it and wait for this issue to be fixed.

evancz:
If you can take a swing at the fix, it would be very appreciated!

Dobiasd:
@evancz I'm note sure if this was for me or not. In case it was: I would love to, but I do not even know where to start. Despite using Elm a lot, I have no clue about how it works internally. And I don't even know how I could start learning about it. :D

evancz:
For anyone who can help really! You can mess with it locally by manually modifying `elm-stuff/packages/elm-lang/core/Native/Task.js` in your project. You can add `console.log` or whatever else into the JS and that'll get packaged up when you run elm-make. If you can find the error in your version of `Native/Task.js` that'll be the bug.

colinmccabe:
I'm seeing these exceptions with a Time.every that gets mapped to an Http request whose result is mailed to a view. There is no loop. Oddly, the pace of the exceptions quickens dramatically when I'm in a different browser tab.

I'm rather obsessed with this and want to fix it. Unfortunately I'm clueless about how Elm actually works. :)

colinmccabe:
This is what is causing the error in my case:

[...]

PR #304, opened 2015-07-21
Mitigate workQueue race condition (#240)

colinmccabe:
- Move workQueue length check out of `setTimeout`
- Quickly store a reference to `workQueue[0]` in a local variable in case `workQueue[0]` changes

Running Async tasks more frequently or increasing the `setTimeout` delay in `onComplete` increases the frequency of the exception described in #240, which suggests that the bug is indeed caused by `workQueue[0]` being overwritten with a proceeding Async task before `runTask()` can run.

I haven't observed the exception with this patch in place, even under extreme conditions with Async tasks being run one after another, milliseconds apart. Without the patch, there is a blizzard of exceptions.

Is this enough to fix the race condition? Not sure exactly how async JS works.

### DIFF
--- a/src/Native/Task.js
+++ b/src/Native/Task.js
@@ -72,12 +72,14 @@ Elm.Native.Task.make = function(localRuntime) {
 		{
 			workQueue.shift();
 
-			setTimeout(function() {
-				if (workQueue.length > 0)
-				{
-					runTask(workQueue[0], onComplete);
-				}
-			}, 0);
+			if (workQueue.length > 0)
+			{
+				var task = workQueue[0];
+
+				setTimeout(function() {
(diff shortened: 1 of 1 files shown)


evancz:
Great sleuthing!

Based on this, I think the current issue comes from a bad interaction between `onComplete` and `register`. Before, it was possible for a task to complete in `onComplete` bringing the number of active tasks down to zero. Then `register` would add a task and start running it. Then the `setTimeout` would run, checking if there's more work to do, which now there is, and then starts running the first task *again*. This repeats all the work until it gets to where the first run paused, causing the crash.

That's my understanding at least. And under that theory, your fix should avoid this.

PR #306, opened 2015-07-22
changes to support new debugger API

vilterp:
- Time signals notify Utils.Tuple0 instead of null
- output nodes save their value (so the debugger can get their values w/out having to look at their parents)
- saving whether a node was really updated each frame

Will let you know when this is ready for review.

### DIFF
--- a/src/Native/Runtime.js
+++ b/src/Native/Runtime.js
@@ -14,8 +14,9 @@ if (!Elm.fullscreen) {
 				connected = true;
 			}
 			var container = document.createElement('div');
-			if(connected) {
-				document.body.appendChild(container);
+			document.body.appendChild(container);
+			if(!connected) {
+				container.style.visibility = 'hidden';
 			}
 			return init(Display.FULLSCREEN, container, module, args || {});
 		};
--- a/src/Native/Runtime.js
+++ b/src/Native/Runtime.js
@@ -8,16 +8,10 @@ if (!Elm.fullscreen) {
 			NONE: 2
 		};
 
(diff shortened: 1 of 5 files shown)


src/Native/Runtime.js
```diff
@@ -10,10 +10,15 @@ if (!Elm.fullscreen) {
...
-		Elm.fullscreen = function(module, args)
+		Elm.fullscreen = function(module, args, connected)
 		{
+			if(typeof connected == 'undefined') {
+				connected = true;
+			}
 			var container = document.createElement('div');
-			document.body.appendChild(container);
+			if(connected) {
+				document.body.appendChild(container);
```
vilterp: TODO: think of a solution better than simply leaving this tree unconnected to the document. As is, there are 2 identical DOM trees (this one and the one owned by the debugger), so we incur DOM mutation costs twice.

evancz:
Doesn't match with things anymore

vilterp:
😢 alas, it does not. I'll take another look at 0.17 core to think about how to make this work…

evancz:
I don't think it'll need any changes here anymore. Take a look at elm-reactor. I cleaned up a bunch of stuff, but ultimately had to turn certain features off to get it releasable. The issues now are not "how can I hack the code?" They are things like "What does it mean to rewind when an effect manager cannot follow suit?" So it requires some thinking about the fundamental capabilities we will be able to provide and how to make that happen.

vilterp:
K, will take a look.

PR #309, opened 2015-07-26
Fixing small `an` to `a` typo in Maybe.elm

machuga:
```markdown
Transform an `Maybe` value with a given function:
```

looks like it was intended to be

```markdown
Transform a `Maybe` value with a given function:
```

in the comments so just giving it a quick adjustment.


ajhager:
:+1: LGTM

evancz:
Nice catch, thank you @machuga!

PR #315, opened 2015-07-30
Improve docs for Signal.forwardTo

mgold:
Because a new version of core may be coming out soon, this is a hopefully very simple and uncontroversial change to the docs for `Signal.forwardTo`. We remove the `port` keyword, which is wrong and misleading. Then, we define the `actions` mailbox in a reasonable way rather than just saying that is exists.

### DIFF
--- a/src/Signal.elm
+++ b/src/Signal.elm
@@ -337,9 +337,10 @@ mailbox =
 {-| Create a new address. This address will tag each message it receives
 and then forward it along to some other address.
 
-    type Action = Undo | Remove Int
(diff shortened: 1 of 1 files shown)


evancz:
Nice, thank you for fixing this! It's a shame the type checker doesn't know about comments :P

mgold:
Thanks for merging. Although, `port actions : Mailbox Action` is a valid incoming port declaration, so maybe even automated compilation of docs examples would have missed it!

PR #316, opened 2015-07-30
Change description of List.member

jessitron:
because this function is called 'contains' in many languages, I hit
this documentation, searched for 'contains,' and didn't find the right
method. Putting 'contains' in the description for 'member' help with this

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -100,7 +100,7 @@ isEmpty xs =
       _  -> False
 
 
-{-| Figure out if a value is a member of the list.
+{-| Figure out whether a list contains a value.
 
     member 9 [1,2,3,4] == False
     member 4 [1,2,3,4] == True


evancz:
Great idea!

I'd be curious to know the list of languages that use that naming convention. Perhaps the best move is to rename this to `contains` in `Set` and `List` and `Array` and such. If you are interested, it would be cool to have a PR in the style of #317 outlining which languages use what and potentially get that added to #322.

Anyway, thanks for the fix!

Mouvedia:
I wanted to do something similar for `Array`. JS has `Array.prototype.includes` (which was shortly supported by certain browsers as `Array.prototype.contains`) so I wanted to use `include` instead.
The problem is there's no member method listed there.

@evancz why is that?

PR #320, opened 2015-08-01
Remove use of window object in Date and Time native modules.

ajhager:
Fixes elm-lang/elm-repl#70 by changing a few instances of `new window.Date` to `new Date`.


evancz:
Great, thank you!

In past versions of native modules, I think there was some clash between our `Date` and JavaScript's `Date` but I guess that is resolved now.

Issue #319, opened 2015-08-01
trimLeft and trimRight not defined in Internet Explorer


jvoigtlaender:
It seems the latest IE (as well as earlier ones) does not support JavaScript's `trimLeft` and `trimRight` on strings (while it supports `trim`, at least the latest IE). As a consequence, Elm code using `String.trimLeft` and `String.trimRight` will fail on IE.

jvoigtlaender:
Confirmed at [MSDN](https://msdn.microsoft.com/de-de/library/ecczf11c%28v=vs.94%29.aspx) that IE does not support `trimLeft` and `trimRight` on strings. Also found some other mentions online of these functions being "non-standard". So maybe avoid relying on them?

TheSeamau5:
Yeah, Mozilla recommends against both `trimLeft` and `trimRight`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/TrimLeft

PR #321, opened 2015-08-01
implement trimLeft and trimRight

jvoigtlaender:
... since they are not standard, thus not supported in all browsers.

Addresses https://github.com/elm-lang/core/issues/319 and thus issues like https://github.com/elm-lang/package.elm-lang.org/issues/40.

### DIFF
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -144,11 +144,11 @@ Elm.Native.String.make = function(localRuntime) {
 	}
 	function trimLeft(str)
 	{
-		return str.trimLeft();
+		return str.replace(/^\s+/,"");
 	}
 	function trimRight(str)
 	{
-		return str.trimRight();
+		return str.replace(/\s+$/,"");
 	}
 
 	function words(str)


evancz:
Ah, nice find! Why do you use `\s` in your regex but mozilla uses a more complex sequence in [their polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim) for trim? I know they are trimming both ends, but they are detecting more than just `\s`.

Also, does `trim` work?

dnalot:
It looks like the EcmaScript specification defines a couple extended Unicode points as whitespace:
```
WhiteSpace ::
<TAB>
<VT>
<FF>
<SP>
<NBSP>
<ZWNBSP>
<USP>
```
For practical purposes they're the same.

dnalot:
Also the MDN link reports `trim` compatibility starting with IE9.

evancz:
Okay, I guess let's go with this. If folks find it odd that `trim` and `trimLeft` work on slightly different kinds of whitespace, let's fix that in another PR.

Thanks @jvoigtlaender and @dnalot!

jvoigtlaender:
I had modelled it on both stuff I "found on the internet" as well as your implementation of `words`:
```javascript
	function words(str)
	{
		return List.fromArray(str.trim().split(/\s+/g));
	}
```
Thought it made sense to do trimming by eliminating the same kind of whitespace that is used for finding word boundaries for splitting in that function.

PR #331, opened 2015-08-04
Don't create div container in fullscreen mode

suryagaddipati:


evancz:
To what extent have you tested that things like `Window.dimensions` work with this change? I suspect it's not as simple as this, but I don't recall exactly the reasons for that div.

suryagaddipati:
Ah, I guessed there was a reason for it but not sure why it was there.  
It was messing up my css, I ended up fixing my css. But the extra div is certainly unexpected. 

I will try to test it with `Window.dimesions`

evancz:
Okay :) I'd check out the signals in `Mouse` as well. I don't recall the exact reason it ended up like this, and perhaps it's a reason that is no longer true.

Anyway, please reopen once you have more data!

Issue #335, opened 2015-08-06
Strange behavior with rotations


jvoigtlaender:
Moved over from https://github.com/elm-lang/elm-compiler/issues/665.

Consider this code:
```elm
import Graphics.Element exposing (..)
import Graphics.Collage exposing (..)
import Color exposing (..)
import Time exposing (..)
import Signal exposing (..)

blacksquare : Form
blacksquare = filled black (square 400)

yogi : Form
yogi = toForm (image 200 200 "imgs/yogi.jpg")

rotating : Form -> Time -> Form
rotating form t = rotate (degrees (inMilliseconds t)) form

canvas : Time -> Element
canvas t = collage 500 500 [rotating blacksquare t, rotating yogi t]

main : Signal Element
main = map canvas (every (20*millisecond))
```
One would expect that both the black square and Yogi will rotate as time progresses. Both are rotated by the same function with the same argument. But only Yogi rotates.

@mgold observed that the behavior is correct if one replaces `every (20*millisecond)` by `foldp (+) 0 (fps 30)`. So maybe it is an issue with large numbers.

jvoigtlaender:
It must have to do with different handling of rotations for "plain" `Form`s and for `Form`s that come from `Element`s. In particular, the black square *will* rotate if one replaces its definition
```elm
blacksquare = filled black (square 400)
```
by
```elm
blacksquare = toForm (collage 400 400 [filled black (square 400)])
```
So the question is:  
Why does the output rendered for `rotate t (toForm (collage 400 400 [ form ]))` change as `t` changes, while the output of `rotate t form` does not change as `t` changes (for large `t`)?

jvoigtlaender:
Okay, so:
* "plain" `Form`s are rotated with [this line](https://github.com/elm-lang/core/blob/a8451ab719223d9324adc517bb7bb83449c1cd8b/src/Native/Graphics/Collage.js#L338), while
* `Form`s that are actually embedded `Element`s are rotated with [this line](https://github.com/elm-lang/core/blob/a8451ab719223d9324adc517bb7bb83449c1cd8b/src/Native/Graphics/Collage.js#L392).

It appears one of these two methods works fine with large numbers (for the angle by which to rotate), while the other doesn't.

mgold:
I did some profiling and `Math.sin` is not any slower for large inputs than small ones. However, it's possible `ctx.rotate` _is_ slow for large angles. You could try `theta = form.theta % (Math.PI * 2)` and see if that does anything.

[...]

PR #338, opened 2015-08-06
Fix rotation of shape forms

jvoigtlaender:
Fixes https://github.com/elm-lang/core/issues/335.

Previously, something like
```elm
main : Signal Element
main = map (\t -> collage 500 500 [rotate (degrees t) (filled black (square 400))]) (every 20)
```
did not work since the `t` got to large for how rotation of `Shape`-`Form`s (as opposed to `Element`-`Form`s) is implemented.

### DIFF
--- a/src/Native/Graphics/Collage.js
+++ b/src/Native/Graphics/Collage.js
@@ -335,7 +335,7 @@ Elm.Native.Graphics.Collage.make = function(localRuntime) {
 		}
 		if (theta !== 0)
 		{
-			ctx.rotate(theta);
+			ctx.rotate(theta % (Math.PI * 2));
 		}
 		if (scale !== 1)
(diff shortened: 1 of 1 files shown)


evancz:
Awesome, thank you!

mgold:
Thanks Janis. Should I be worried about the cost of modulus every frame?

evancz:
I was thinking it may make sense to define `twoPi` or something, but I suspect the mod is fine.

TheSeamau5:
You mean, `tau`? http://tauday.com/

mgold:
+1 for tau!

evancz:
@mgold, do a PR instead of a +1. I also know about tau, but I'd still like to call it `twoPi

PR #343, opened 2015-08-11
fix result of String.toFloat "." (#336)

Pisys:
which is now a Err instead of Ok NaN.
Also added tests for String conversion.

### DIFF
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -252,15 +252,16 @@ Elm.Native.String.make = function(localRuntime) {
 			return Result.Err("could not convert string '" + s + "' to a Float" );
 		}
 		var start = 0;
-		if (s[0] == '-')
+		var dotCount = 0;
+		if (s[0] == '-' || s[0] == '.')
 		{
 			if (len === 1)
 			{
 				return Result.Err("could not convert string '" + s + "' to a Float" );
 			}
 			start = 1;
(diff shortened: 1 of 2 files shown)


evancz:
I wonder if this is how we should be implementing this in the first place. Maybe it would be better to use a regex like `[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?` or to just put `parseFloat` in a `try` block and trust that it'll do the right thing?

I also wonder why `parseFloat('.') == NaN` in JavaScript. Can you try to find out why that is true?

Pisys:
parseFloat is less strict than the current implementation. It allows whitespaces, exponents, and only demands a prefix of the input string to be a float. Since "." is not a Float parseFloat returns NaN!

However, parseFloat can also parse "Infinity", which is a Number in JavaScript but not in Elm! So I discourage the mere wrapping of parseFloat until there is a explicit Infinity thing in Elm, such as `x = Infinity` (instead of `x = 0/0`).

jvoigtlaender:
How is this pull request related to https://github.com/elm-lang/core/pull/342?

jvoigtlaender:
I'm closing this for now, since it appears to be overridden by https://github.com/elm-lang/core/pull/342. Please re-open with comment if you disagree.

PR #346, opened 2015-08-13
Add JSHint configuration file and apply many suggestions

Dandandan:


src/Native/Graphics/Element.js
```diff
@@ -334,6 +334,7 @@ Elm.Native.Graphics.Element.make = function(localRuntime) {
 
 			case 'Z':
 				transform = 'translateX(' + ((-w/2)|0) + 'px) ';
+				break;
```
evancz: These actually are supposed to fall through. Only excuse I can make is that this was written by a younger me ;P

src/Native/Json.js
```diff
@@ -137,9 +137,11 @@ Elm.Native.Json.make = function(localRuntime) {
 			if (isObject) {
 				var keyValuePairs = List.Nil;
 				for (var key in value) {
-					var elmValue = decoder(value[key]);
-					var pair = Utils.Tuple2(key, elmValue);
-					keyValuePairs = List.Cons(pair, keyValuePairs);
+					if (value.hasOwnProperty(key)) {
```
evancz: Why does this make this code better? Seems like it'll take some extra time and either (a) not protect us from anything because of how the Elm API is designed or (b) lead to weird results where we aren't allowed to grab fields that we want. Can you explain more about this?

evancz:
If the `hasOwnProperty` stuff is better, can you do a separate PR to add it back in? I removed it because I wasn't sure how it changed the existing behavior and I have not heard of folks with problems there.

evancz:
Otherwise, nice work, thank you! :D

evancz:
Also, is it possible to ask for tab indentation only? Right now it says 2, does that mean it wants spaces?

src/Native/Json.js
Raynos:
```js
function Foo() {
  this.myField = null;
}
Foo.prototype.doAThing = function doAThing() {}

var f = new Foo();
for (var k in f) console.log(k); // myField, doAThing
```

Using a `for var in` loop on an object with methods is unsafe.

evancz: Sure, but decoders won't let functions through, so that case never happens. So that's covered by (a)

Issue #337, opened 2015-08-06
Json.Decoder.int Bug / Ints in Elm


agrafix:
The javascript spec allows numbers up to 2^53-1 (`9007199254740991`) and Elm generally is fine with these large integers, too. The json decoder fails for inputs larger than `2147483647`. 

I've found the error in the Native.Json part:
https://github.com/elm-lang/core/blob/2a8dcb522f71541c299a8a465bb39bc17bb9ec67/src/Native/Json.js#L53

It checks that `(value|0) === value`, but this check will fail for any input larger than `2147483647` because binary operators only work on 32bit ints.

I did not send a pull request (yet?) because I'm not sure what the correct step is. The Bitwise module allows parameters of type `Int`, and this will produce broken results iff the number is larger. Should elm distinguish between `Int` (32 Bit) and `LargeInt` (53 Bit)? Or should we just fix the json parser?

evancz:
PR #153 was open with a potential fix, but we closed it so @ajhager or @krisajenkins could do a PR with the faster version of the fix.

ajhager:
And the new PR is on its way. :)

agrafix:
Very cool, looking forward to it! :-) @ajhager

agrafix:
@evancz what do you think about the behavior of the bitwise module? maybe there should at lease be a warning that is only works for 32 Bits Ints?

jvoigtlaender:
Is this completely addressed by #352 (which was merged)? If so, please close the issue. If not, please say what is not addressed.

evancz:
Not sure what happened here, so I am going to close. Someone please reopen with fresh information if something else needs to happen.

PR #352, opened 2015-08-16
Add support for large integers in Json decoder

ajhager:
The current Json int decoder only recognizes 32 bit integers. This PR provides a small test suite for the int decoder and a new decoder that works correctly with large numbers.

This benchmark (where 'original' is the old code and 'blend' is the code from this PR) shows (on OSX with Chrome at least) that the new code is slightly slower in the case that most decoded values are *not* ints, and nearly equivalent speed in the typical case when most values *are* ints.

ref #153 and #337


evancz:
Looks great to me, thank you! :D

@krisajenkins, does this have the behavior you were looking for?

evancz:
Great, thanks everyone!

PR #353, opened 2015-08-16
more style improvements

Dandandan:


Dandandan:
This runs now without issues.

evancz:
I'm not a major fan of splitting up the `return` and setting the content. How many passes are you thinking of taking over this code? With each one I am afraid we will accidentally introduce some issue, so I comb through them all very carefully.

What is the end goal of this? How will we know when we reached that goal?

Dandandan:
I share your concern about errors. Is there a way to make this easier, for example, one PR for each module or type of linting error?
I guess the end goal is to make the JS runtime easier to maintain, the linting check could even be added to the CI server when the linter passes (almost) all checks.

evancz:
I think this is too out of date to make sense to try to merge. I'm not sure how much more aggressive I want to be with style stuff at the moment. This stuff just makes me nervous!

PR #354, opened 2015-08-16
test.js needs to be imported once Elm object is no longer global under node

ajhager:
This PR should only be accepted when and if elm-lang/elm-make#43 is. That PR adds support for node, and so the elm-io-ports code should no longer be prepended to test.js, but instead require test.js as a node module.


jvoigtlaender:
Is this pull request still relevant? It looks like it might be overridden by https://github.com/elm-lang/core/pull/447.

PR #359, opened 2015-08-19
fix Regex.replace with AtMost

Pisys:
passing (AtMost n) behaved actually like (AtMost n+1).
Also added some test cases.


evancz:
Nice find, thank you!

Issue #362, opened 2015-08-21
Initial value of Keyboard.presses is shown as <internal structure>


grzegorzbalcerek:
Consider the following program:

    module KeyboardSignals2 where

    import Graphics.Element exposing (show)
    import Keyboard
    import Signal exposing ((<~))

    main = show <~ Keyboard.presses

It shows:

     <internal structure>

initially. Only after pressing a key, it shows its code. But, shouldn't a signal always have a value consistent with its data type?

jvoigtlaender:
Fixed by https://github.com/elm-lang/core/pull/363.

But note that by using `Keyboard.presses` you are in for further trouble: https://github.com/elm-lang/core/issues/326

PR #363, opened 2015-08-21
Fix initial value of Keyboard signals

jvoigtlaender:
Fixes https://github.com/elm-lang/core/issues/362 (and likely analogous issues for other signals in the `Keyboard` module).


evancz:
Thanks for figuring the root of that out!

Issue #364, opened 2015-08-22
Example for Json.Decode.value wrong in documentation


jvoigtlaender:
The example at http://package.elm-lang.org/packages/elm-lang/core/2.1.0/Json-Decode#value is missing some brackets. Adding them gives:
```elm
variadic2 : (a -> b -> List c -> value) -> Decoder a -> Decoder b -> Decoder (List c) -> Decoder value
variadic2 f a b cs =
    customDecoder (list value) (\jsonList ->
        case jsonList of
          one :: two :: rest ->
              Result.map3 f
                (decodeValue a one)
                (decodeValue b two)
                (decodeValue cs rest)

          _ -> Result.Err "expecting at least two elements in the array")
```
But that's not type-correct according to http://elm-lang.org/try. (It complains about mismatched types `Value` vs. `List Value` in the 2nd argument of `(decodeValue cs rest)`.)

jvoigtlaender:
A possible fix is this:
```elm
variadic2 : (a -> b -> List c -> value) -> Decoder a -> Decoder b -> Decoder c -> Decoder value
variadic2 f a b c =
    customDecoder (list value) (\jsonList ->
        case jsonList of
          one :: two :: rest ->
              Result.map3 f
                (decodeValue a one)
                (decodeValue b two)
                (combine (List.map (decodeValue c) rest))

          _ -> Result.Err "expecting at least two elements in the array")
```
where
```elm
combine : List (Result x a) -> Result x (List a)
combine = List.foldr (Result.map2 (::)) (Ok [])
```
is a function analogous to http://package.elm-lang.org/packages/Apanatshka/elm-signal-extra/5.4.1/Signal-Extra#combine. (Or to http://hackage.haskell.org/package/base-4.8.1.0/docs/Data-Traversable.html#v:sequenceA.)

jvoigtlaender:
Closing via https://github.com/elm-lang/core/pull/365.

PR #365, opened 2015-08-22
Fix the example for "crazily formatted data"

jvoigtlaender:
Fixes https://github.com/elm-lang/core/issues/364.


evancz:
Thank you!

PR #371, opened 2015-08-25
Reword docs for move to reflect order of args

### LINKED ISSUE #370: Documentation for Graphics.Collage.move implies (y,x) rather than (x,y)
From [here](https://github.com/elm-lang/core/blob/2.1.0/src/Graphics/Collage.elm#L220)

>`(move (10,10) form)` would move `form` ten pixels up and ten pixels to the
right.

when in actuality the code is right first, then up.

```
move : (Float,Float) -> Form -> Form
move (x,y) f =
  { f | x <- f.x + x, y <- f.y + y }
```

eeue56:
Fixes #370

### DIFF
--- a/src/Graphics/Collage.elm
+++ b/src/Graphics/Collage.elm
@@ -220,9 +220,8 @@ groupTransform matrix fs =
   form (FGroup matrix fs)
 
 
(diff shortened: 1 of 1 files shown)


evancz:
Nice, much clearer! Thank you :)

PR #375, opened 2015-08-28
Remove multi-way if usage, and use = for record updates.

laszlopandy:
@evancz 
@JoeyEremondi

### DIFF
--- a/src/Color.elm
+++ b/src/Color.elm
@@ -189,9 +189,12 @@ rgbToHsl red green blue =
 
     hue =
       degrees 60 *
-        if  | cMax == r -> ((g - b) / c) `fmod` 6
-            | cMax == g -> ((b - r) / c) + 2
-            | cMax == b -> ((r - g) / c) + 4
+        if cMax == r then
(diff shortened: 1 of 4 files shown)


evancz:
Looks good to me, thank you!

hughfdjackson:
If this change is intended for use with 0.16 is in master, should we update the `elm-version` in `elm-package.json`? 

I'm failing to get master to compile on the latest version available via http://elm-lang.org/install, which I believe to be 0.15.1.

laszlopandy:
Yes. This is true. But I believed that we don't need to bump until we
publish a new version of core to the package repository.

hughfdjackson:
:) From the perspective of someone looking to submit a first PR, it would be time-saving if the `elm-package.json` file reflected the current requirements of master.  Took a little head-scratching to realise I was using the wrong compiler version. 

I'm happy to bump the version requirements in a PR if you and @evancz (et. al.) are on board with that.

laszlopandy:
Makes sense to me.

Issue #373, opened 2015-08-27
millisecond docs imply a "milliseconds" variable exists


eeue56:
Had one of my students confused by this line

>Units of time, making it easier to specify things like a half-second (500 * milliseconds) without remembering Elm’s underlying units of time.

in the [docs for time](http://package.elm-lang.org/packages/elm-lang/core/2.1.0/Time#millisecond) due to the fact it says `500 * milliseconds` and not `500 *millisecond` or `500 milliseconds`.

I think something along the lines of

> Units of time, making it easier to specify a time without remembering Elm’s underlying implementation of time.
> `halfSecond = 500 * millisecond` would create a variable with the value of 500ms, or half a second.

Might make it a little more clear?

jvoigtlaender:
Closed via https://github.com/elm-lang/core/pull/376.

PR #376, opened 2015-08-28
Reword millisecond docs to give example of usage

eeue56:
As per #373, makes millisecond docs simpler to understand

### DIFF
--- a/src/Time.elm
+++ b/src/Time.elm
@@ -31,7 +31,7 @@ type alias Time = Float
 
 
 {-| Units of time, making it easier to specify things like a half-second
-`(500 * milliseconds)` without remembering Elm&rsquo;s underlying units of time.
+`(500 * millisecond)` without remembering Elm&rsquo;s underlying units of time.
 -}
 millisecond : Time
 millisecond =


jvoigtlaender:
I don't really like the talk of "variable" here (in the new documentation comment). Does the `core` documentation use language like "variable" elsewhere?

Generally, wouldn't a much easier fix to the problematic aspect of the documentation you point out be to simply replace `(500 * milliseconds)` by `(500 * millisecond)`? Why is the rest of the rewording necessary from your perspective?

eeue56:
I've removed the word variable now - I agree it didn't feel right there.

That would be an easier fix, but a less clear fix I feel. I wanted an explicit code example that people could use right away, rather an example just in the brackets, which doesn't seem as obvious that it's meant to be code. Adding `halfSecond = 500 * millisecond` makes it more explicit and obvious than `(500 * millisecond)` in the middle of a sentence, basically.

eeue56:
Though I could just change it to remove the `s` if you think that's clear enough.

jvoigtlaender:
I actually think just removing the plural `s` would be enough.

jvoigtlaender:
That it's code, should be clear from type-setting on the docs page.

eeue56:
Okay - changed it so it's now just the 's' removal, and got rid of the previous commits I made too.

PR #379, opened 2015-08-30
Minor Runtime.js cleanup

rtfeldman:
* Delete an obsolete comment
* Use stricter `typeof container.appendChild === "function"` instead of `typeof container.appendChild !== 'undefined'`


evancz:
Seems okay, thank you!

Issue #378, opened 2015-08-29
Regex, runtime error


gunpinyo:
In core package, module Regex
```elm
regex : String -> Regex
```

This function can produce runtime error if string contain invalid pattern.

Here is an example form elm-repl
```
b = Regex.regex("[a-z")
SyntaxError: Invalid regular expression: /[a-z/: Unterminated character class
```

It would be better if we change regex to be
```elm
regex : String -> Maybe Regex
```
it will return `Nothing` if string contains invalid pattern.

eeue56:
I think it makes sense for regex to be `regex : String -> Result Regex` here, so that the error message is captured. If it's okay, I'll make a pull request for this so that it can be merged if it's approved

mgold:
Yeah, it's either that or keep the runtime error. Most regexes won't be generated dynamically (in signals or functions), so there's a case for keeping the runtime error since it will be easier to work with the absence of a Result or Maybe. (We seem to be getting useful error messages, so I support a Result in the API.)

rtfeldman:
In the spirit of [these guidelines](https://github.com/elm-lang/elm-plans/blob/master/CONTRIBUTING.md) I've made [a gist](https://gist.github.com/rtfeldman/a3614e123381e7751b4c) for a potentially interesting solution to this. Any discussion of that idea is welcome over on the gist itself, but let's not derail this issue with that stuff. :smile:

jvoigtlaender:
Is this "the same" as https://github.com/elm-lang/core/issues/157? Should we close that other issue, then?

eeue56:
I've created #380  that will change `regex` to return a `Result` instead, but I think it needs further discussion before merging. Having to use `case..of` everywhere doesn't quite fit right. 

Perhaps leaving `regex` as it's own function and introducing `safeRegex : String -> Result String Regex` might be a better solution, with `regex` intended for static expressions and `safeRegex` for dynamic ones. 

This would remove the need to introduce a new syntax as per rtfeldman's gist, though leave a runtime error possible.

drbo:
Regex literals might be a solution.  That way invalid regexps are a compile time error

mgold:
That's been brought up before, and I agree would be a good solution. But it involves a compiler change.

tmcw:
[...]

PR #380, opened 2015-08-30
Change regex to return Result

eeue56:
Changed the `Regex.regex` function to return a `Result` as per #378 and updated and added tests to this effect. I haven't changed any of the documentation for any other function in Regex.elm, as I feel that [rtfeldman's comment](https://github.com/elm-lang/core/issues/378#issuecomment-136084687) on providing a built-in regex syntax needs to be discussed further - certainly it makes code more bulky and there is no way to create a `Regex` without the `regex` function.

### DIFF
--- a/src/Native/Regex.js
+++ b/src/Native/Regex.js
@@ -13,6 +13,7 @@ Elm.Native.Regex.make = function(localRuntime) {
 
 	var List = Elm.Native.List.make(localRuntime);
 	var Maybe = Elm.Maybe.make(localRuntime);
+	var Result = Elm.Result.make(localRuntime);
 
 	function escape(str)
 	{
@@ -23,8 +24,19 @@ Elm.Native.Regex.make = function(localRuntime) {
 		return new RegExp(re.source, 'gi');
 	}
 	function regex(raw)
-	{
-		return new RegExp(raw, 'g');
+	{	
+		var compiledRegex = null; 
+		
+		try 
+		{
+			compiledRegex = new RegExp(raw, 'g');
+		}
+		catch (err)
+		{
+			return Result.Err(err.message);
+		}
+		
+		return Result.Ok(compiledRegex);
 	}
 
 	function contains(re, string)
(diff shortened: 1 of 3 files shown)


rtfeldman:
The more I think about it, the more I like this idea. You can use it to implement the current behavior by using `Debug.crash` to handle the `Err` case, which could be offered as a convenience function like `regexOrCrash`

eeue56:
Yeah, that's true. I like the `Debug.crash` approach

robertjlooby:
Has any more thought been given to making this change?

src/Native/Regex.js
```diff
@@ -23,8 +24,19 @@ Elm.Native.Regex.make = function(localRuntime) {
...
 	}
 	function regex(raw)
-	{
-		return new RegExp(raw, 'g');
+	{	
+		var compiledRegex = null; 
+		
+		try 
+		{
+			compiledRegex = new RegExp(raw, 'g');
```
tmcw: minor nitpick that you could express this more succinctly as 

```js
return Result.Ok(new RegExp(raw, 'g'));
```

Here, and avoid using the temporary variable - the code would have the same behavior since `new RegExp(raw, 'g')` is evaluated first, would throw, and not instantiate a Result.Ok

tmcw:
Yep, as I posted [on the other ticket](https://github.com/elm-lang/core/issues/378#issuecomment-231613057), I 2nd this approach as flexible as lower-impact than a syntax extension.

seanhess:
This would be very helpful to me for the project I am currently working on. Right now we are blocked without it.

jvoigtlaender:
@seanhess, if you are blocked by this in a concrete project, maybe bring it up on the mailing list? Here, that comment might avoid notice for a while.

eeue56:
@seanhess here's how to unblock you:

- Create a function that looks like this:

```javascript

var isValidRegex = function(regex, modifiers){
  try {
    new RegExp(regex, modifiers);
  } catch (e) {
    return [regex, false];
  }

  return [regex, true];
}
```

- Make a port that looks like `validateRegex : Cmd String`
- Make a port that looks like `isValidRegex : Sub (String, Bool)`

If you really want, it's trivial to introduce this as a native function into your project. Don't let yourself get blocked on this! it's a trivial issue to solve :)

seanhess:
@eeue56 ssh, you're removing the imperative to get this merged ;D

Thanks though, that's a great workaround! I forget that native modules are an option for end applications since they can't be published.

evancz:
Tracking in #722. These changes are quite easy, so I don't think a PR is needed. Just need to figure out the appropriate time to actually make a change like this.

PR #382, opened 2015-08-31
Avoid double definition of KeyCode alias

jvoigtlaender:
Previously existed both in `Char` and `Keyboard` modules.

Now that the next compiler version will have dead code elimination, there should be no harm in importing the alias from one into the other.

Defining `KeyCode` only in one place makes for more future-proofness.

(Also, the `KeyCode` in the type signatures in `Char` module will turn into links to the definition site in `Keyboard` module on the package site, so there's no harm in terms of documentation.)

### DIFF
--- a/src/Char.elm
+++ b/src/Char.elm
@@ -1,6 +1,6 @@
 module Char
     ( isUpper, isLower, isDigit, isOctDigit, isHexDigit
-    , toUpper, toLower, toLocaleUpper, toLocaleLower, KeyCode, toCode, fromCode
+    , toUpper, toLower, toLocaleUpper, toLocaleLower, toCode, fromCode
     ) where
 
 {-| Functions for working with characters. Character literals are enclosed in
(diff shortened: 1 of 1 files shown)


evancz:
Nice, this makes a lot of sense!

evancz:
This introduces a dependency cycle.

```
  ┌────┐
  │   Keyboard
  │    ↓
  │   Set
  │    ↓
  │   Dict
  │    ↓
  │   String
  │    ↓
  │   Char
  └────┘
```

I think the centralized thing needs to live in `Char`. Not sure yet though.

evancz:
Alright, I did the reverse of this and it is all set!

The master branches of elm-compiler / elm-package / elm-make are able to produce code again. It will not run properly, but we can at least get all the normal error reporting again.

jvoigtlaender:
Ah, I had tried to check the module imports for dependencies manually, but missed that path. (Compiling to get the check automatically was not possible at the time.)

evancz:
No worries, it only started working again in the past few days :)

PR #383, opened 2015-08-31
Adding/fixing docu links

jvoigtlaender:
Added various useful links from some module to another one. No changes to text otherwise (except for one correction of an article word), and no changes to code at all.

In one case, fixed a link that was 404'ing.

### DIFF
--- a/src/Char.elm
+++ b/src/Char.elm
@@ -84,8 +84,8 @@ toLocaleLower =
 type alias KeyCode = Int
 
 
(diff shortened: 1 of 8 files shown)


evancz:
Ah, I think I did things in the wrong order, I guess the other PR blocks this one?

jvoigtlaender:
I thought the two PRs were independently mergeable, but they were indeed not. Fixed now.

evancz:
Looks great, thank you!

PR #384, opened 2015-09-01
Fix punctuation, capitalization and spelling in documentation

ianbollinger:
Fix a few typos and a few other orthographic mistakes.

### DIFF
--- a/src/Basics.elm
+++ b/src/Basics.elm
@@ -89,13 +89,13 @@ turns =
   Native.Basics.turns
 
 
-{-| Convert polar coordinates (r,&theta;) to cartesian coordinates (x,y). -}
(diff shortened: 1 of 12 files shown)


evancz:
Can you split the spelling out into its own PR, those are all fine.

I don't think the other stuff is as clear. For example, [timestamp](https://en.wikipedia.org/wiki/Timestamp) can be used as one word. With the capitalization stuff as well, I think a lot of that varies based on usage.

Looks like Cartesian should be caps ([evidence](https://en.wikipedia.org/wiki/Cartesian_coordinate_system)) but there are others that are less certain. Smaller PRs make it easier to figure out. If you can do a small change and say on what grounds is it "better English" then it'll be easier to change. For example, `Json` is a specific thing within Elm, and it handles JSON values. Not sure which is more appropriate.

PR #385, opened 2015-09-01
Clarify isUpper and isLower use ASCII

ianbollinger:
Clarify that `Char.isUpper` and `Char.isLower` only recognize ASCII letters.

### DIFF
--- a/src/Char.elm
+++ b/src/Char.elm
@@ -26,13 +26,13 @@ isBetween low high char =
       (code >= toCode low) && (code <= toCode high)
 
 
-{-| True for upper case letters. -}
+{-| True for upper case ASCII letters. -}
 isUpper : Char -> Bool
 isUpper =
   isBetween 'A' 'Z'
 
 
-{-| True for lower case letters. -}
+{-| True for lower case ASCII letters. -}
 isLower : Char -> Bool
 isLower =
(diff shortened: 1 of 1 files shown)


evancz:
What do you think of using [this approach](http://stackoverflow.com/a/1077692) instead? Looks like we could capture unicode stuff as well, but honestly, I'm not 100% sure what that means in practice.

jvoigtlaender:
I'm going to merge the documentation change here. It is a fact that the current implementations of `isUpper` and `isLower` work only for ASCII characters. Making them work for unicode characters would be a separate issue. Right now, the documentation should state the correct thing, which this PR achieves.

rofrol:
@evancz I like this approach better http://stackoverflow.com/questions/1027224/how-can-i-test-if-a-letter-in-a-string-is-uppercase-or-lowercase-using-javascrip/9728437#9728437

And here is a quick workaround in elm:

```
isUpper: Char -> Bool
isUpper c =
  let
    s = String.fromChar c
  in
    s == String.toUpper s && s /= String.toLower s

isLower: Char -> Bool
isLower c =
  let
    s = String.fromChar c
  in
    s /= String.toUpper s && s == String.toLower s
```

PR #386, opened 2015-09-02
Update Decode.elm

rick68:
fix a typo


evancz:
Oops, thank you for the fix!

Issue #271, opened 2015-06-12
Array.indexedMap's index is 1023 maximum


evancz:
(Originally reported by @tyomalu in https://github.com/elm-lang/elm-lang.org/issues/262)

Here is the issue:
```elm
import Array
import Array (Array)
import Text (asText)

testArray: Array number
testArray = Array.initialize 3000 identity

mapArray: Array a -> Array a
mapArray array =
    Array.indexedMap (\i el -> 
        -- here max i == 1023
        case (Array.get i array) of
            Just x -> x
            otherwise -> el
    ) array


main =
    testArray |>
    mapArray |>
    asText
```
mapArray should just return array itself with values 0..2999, but it returns 0..1023, 0..1023, 0..951.
I've found that index in function passed to indexedMap reaches 1023 maximum.
I suggest the problem is in Array.js indexedMap_ function:
```javascript
    function indexedMap_(f, a, from) {
      var newA = { ctor:"_Array", height:a.height, table:new Array(a.table) };
      if (a.height > 0) { newA.lengths = a.lengths; }
      for (var i = 0; i < a.table.length; i++) {
        newA.table[i] = a.height == 0 ? A2(f, from + i, a.table[i])
                                      : indexedMap_( f, a.table[i]
                                                   , i == 0 ? 0 : a.lengths[i - 1]); // should be i == 0 ? from : from + a.lengths[i-1]);?
      }
      return newA;
    }
```

jvoigtlaender:
Closing this, since it appears it's fixed by https://github.com/elm-lang/core/pull/391. (Right, @eeue56?)

eeue56:
Yup, fixes it so can be closed

PR #391, opened 2015-09-04
Make Array.indexedMap work for all array sizes

eeue56:
Make Array.indexedMap work for all array sizes, instead of hitting 1024 and wrapping around, as #271 states. Added a test case from that example to ensure that the issue is fixed in future verisons


evancz:
Awesome, thank you for taking a look at the Array issues!

Do I have a [contributors agreement](https://github.com/elm-lang/elm-compiler/blob/master/ContributorAgreement.pdf) from you?

eeue56:
No problem. Yup, I sent it from enalicho@gmail.com in March I think

evancz:
Thanks!

PR #393, opened 2015-09-05
Task docs

rtfeldman:
This revises the `Task` docs to be more helpful to `elm-effects` users searching for how to obtain tasks that can never fail.

The resulting docs json file: [`task-docs.json`](https://gist.github.com/rtfeldman/ef89951043fafa67ad1b)

### DIFF
--- a/src/Task.elm
+++ b/src/Task.elm
@@ -200,16 +200,15 @@ mapError f task =
   task `onError` \err -> fail (f err)
 
 
-{-| Helps with handling failure. Instead of having a task fail with some value
-of type `x` it promotes the failure to a `Nothing` and turns all successes into
-`Just` something.
+{-| Translate a task that can fail into a task that can never fail, by
+converting any failure into `Nothing` and any success into `Just` something.
 
     toMaybe (fail "file not found") == succeed Nothing
     toMaybe (succeed 42)            == succeed (Just 42)
 
 This means you can handle the error with the `Maybe` module instead.
 -}
-toMaybe : Task x a -> Task y (Maybe a)
+toMaybe : Task x a -> Task never (Maybe a)
 toMaybe task =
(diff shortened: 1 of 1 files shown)


vilterp:
Having a type variable `never` here and a type `Never` in `elm-effects` might be confusing. Maybe an argument for moving `Never` type to `core` and using it here?

rgrempel:
I might be misunderstanding, but I believe the lower-case `never` is just documentation -- it essentially means that the error type is unconstrained -- since it never happens, the type system need not care about it. Whereas the upper-case `Never` is, technically, an actual type. So (and I may be missing something here) it seems like one would be over-constraining the type checker by making `Task.toMaybe` return a `Task Never (Maybe a)`. Whereas making it return a `Task never (Maybe a)` is just documentation.

evancz:
Yeah, @rtfeldman and I discussed this. Richard, can you give more context on these things? Often when we discuss things offline, that info does not make it into the PR or issue and leads to confusion.

I'd be curious to see @vilterp's thoughts on the reasoning.

rtfeldman:
Yep, sorry about that!

My understanding is that @rgrempel's [comment](https://github.com/elm-lang/core/pull/393#issuecomment-138017972) is correct as to what the distinction implies.

As for motivation, the reason I wanted to rename it was for searchability; if you are currently working with elm-effects and you think "okay, I need to get to a Never", you reflexively bring up the Task docs and search for "never" to see what comes back. Currently nothing comes back, so I have to hunt function by function until I find one whose docs indicate it'll do what I need.

Changing it to `Never` would be overly restrictive, but changing it to `never` makes it searchable. Granted, it could make people wonder "why is it capitalized in one place and not the other?" but that seems like the least painful of the options on the table.

PR #394, opened 2015-09-06
Error in docs

simonh1000:
... I think


jvoigtlaender:
Yes, that was a mistake in the documentation indeed. Thanks!

PR #395, opened 2015-09-06
Bumped Elm version requirement to 0.16+

hughfdjackson:
Recent changes merged into master don't compile with 0.15.x. To
avoid confusion for contributers, this commit requires Elm 0.16+.

The specific PRs that seem to have caused problems for me are #382, which results in a cyclical dependency error, and #375, which complains at duplicate keys in Records for `Graphics.Element`'s `width`, etc.

### DIFF
--- a/elm-package.json
+++ b/elm-package.json
@@ -41,5 +41,5 @@
     ],
     "native-modules": true,
     "dependencies": {},
-    "elm-version": "0.15.0 <= v < 0.16.0"
+    "elm-version": "0.16.0 <= v < 0.17.0"
 }


evancz:
Right now nothing will compile with the master branch of elm-compiler. Making this change means we need to do a version bump that will ripple through all the elm repos. elm-package depends on 0.16, etc.

Due to some silly things with cabal, I tend to wait til the last moment to bump all the dependencies.

So given that nothing can be built with the latest compiler anyway, can we wait on this a bit?

hughfdjackson:
@evancz :) Absolutely.  Closing this PR.

Issue #176, opened 2015-02-16
Array.repeat bug with powers of 32 sizes


TheSeamau5:
For some reason, the following code:

```elm
import Array 
import List
import Text (asText)

main = 
  asText <|
    (Array.fromList (List.repeat 32 -740239360)) == (Array.repeat 32 -740239360)
```

yields `False`.

You can convince yourself that the repeat function is using the same number. 

This was found using the following elm-check property:

```elm
property2 "Repeat works the same on lists and arrays" 
          (\n a -> Array.fromList (List.repeat n a) == Array.repeat n a) 
          (int 0 100) anyInt
```

which yielded:

```
Repeat works the same on lists and arrays has failed with the following input: (32,-740239360)
```

TheSeamau5:
This bug seems to affect calling repeat with a size that is a power of `32`

For example, the following yields true:

```elm
import Array 
import List
import Text (asText)

main = 
  asText <|
    (Array.fromList (List.repeat 31 1)) == (Array.repeat 31 1)
```

so, does this:

```elm
import Array 
import List
import Text (asText)

main = 
  asText <|
    (Array.fromList (List.repeat 33 1)) == (Array.repeat 33 1)
```

But these do not:

```elm
import Array 
import List
import Text (asText)

main = 
  asText <|
    (Array.fromList (List.repeat 32 1)) == (Array.repeat 32 1)
```

```elm
import Array 
import List
import Text (asText)

main = 
  asText <|
    (Array.fromList (List.repeat 1024 1)) == (Array.repeat 1024 1)
```

But this one works:

```elm
import Array 
import List
import Text (asText)

main = 
  asText <|
    (Array.fromList (List.repeat 0 1)) == (Array.repeat 0 1)

```

mgold:
Array equality [seems to have some issues](https://groups.google.com/d/msg/elm-discuss/TyMdNOEuOE8/YiDAEGhLEY8J), and the same fix may resolve a host of problems. In particular, it seems to be comparing tree representations rather than the abstract flattened lists.

TheSeamau5:
Cool! Looking forward to the fix.

mgold:
I don't have a fix, but a proper implementation of array equality ought to fix both issues.

TheSeamau5:
Oh...ok... That doesn't matter. I'll still look forward to a fix.

eeue56:
[...]

PR #396, opened 2015-09-08
Correct tree height generation for arrays of size 32^n

eeue56:
"Fixes" #176 by making initialize and fromList generate trees of the same height for the same amount of elements. This fixes the `eq` problem by making sure that the way trees are generated (in terms of height and leafing) is consistent between the two functions.
Doesn't fix the issue of equality checking on arrays with different heights yet the same elements, however.


evancz:
Fix for equality is better done a different way. Overall fix for Array is better done with the Elm rewrite that is in progress.

Issue #397, opened 2015-09-08
 Exception when native submodules of Window present


colinmccabe:
When:

* A submodule of `Window` (for example, `Window.Foo`) is used in an app
* and `Window.Foo` contains native code (i.e. it has a `Native.Window.Foo`)
* and `Window` itself is used nowhere in the app

an exception occurs in `Runtime.js`: `Uncaught TypeError: Cannot read property 'resizeIfNeeded' of undefined`

Here are the relevant lines:

https://github.com/elm-lang/core/blob/2.1.0/src/Native/Runtime.js#L423
https://github.com/elm-lang/core/blob/2.1.0/src/Native/Runtime.js#L434

The problem is that `elm.Native.Window` is defined because of the presence of the native submodule, but `elm.Native.Window.values` is undefined because the `Window` module itself is not used in the app. An exception might occur in either of the two if conditions.

I stumbled upon this while writing a library. Any library (or future version of core) that defines a module under `Window` might encounter this issue.

jvoigtlaender:
Since there is a pull request addressing this issue, I am closing the issue itself.

PR #398, opened 2015-09-08
Fix exceptions when native submodules of Window present

colinmccabe:
Fixes issue #397


evancz:
This will not work the same way in the next release. Thanks for the fix though!

Issue #349, opened 2015-08-15
Array slicing beyond the end fails for arrays 32 elements or longer


jconti:
As I noted in the Google Group (https://groups.google.com/d/msg/elm-discuss/8RXblDRVjas/gpEnMLy-BwAJ) I have created a small example of a runtime failure of Array.slice:

```elm
import Array
import List
import Graphics.Element exposing (..)

-- Fails for value of 32, works for 31
aryLast = 32

sliceOffEnd =
  let
    ary = Array.fromList [0..aryLast]
  in
    Array.append (Array.slice 1 aryLast ary) (Array.slice (aryLast + 1) -1 ary)

main : Element
main =
  show sliceOffEnd
```

eeue56:
I've put a fix for this in #399 as I've been hitting the issue a lot with my matrix library.

evancz:
Closing in favor of #649 which will track all array issues and progress on a new implementation.

PR #399, opened 2015-09-09
Fix Array.slice for `from` > Array.length and `to` > Array.length

eeue56:
Adds checks for `from` and `to` being great or equal to the length of the array, fixing #349 by preventing `a.height` issues.

### DIFF
--- a/src/Native/Array.js
+++ b/src/Native/Array.js
@@ -418,6 +418,19 @@ Elm.Native.Array.make = function(localRuntime) {
 	// optimized.
 	function slice(from, to, a)
 	{
+		// if the start of the slice goes past the end of the array, 
+		// then just return empty
+		// -1 since array is 0 based
+		if (from >= length(a)){
+			return empty;
+		}
+
+		// if the end of the slice goes past the end of the, 
+		// just make to the length of the array
+		if (to >= length(a)){
+			to = length(a);
+		}
+		
 		if (from < 0)
 		{
 			from += length(a);
--- a/src/Native/Array.js
+++ b/src/Native/Array.js
@@ -426,8 +426,8 @@ Elm.Native.Array.make = function(localRuntime) {
 		}
 
 		// if the end of the slice goes past the end of the, 
-		// just make to the length of the array
-		if (to >= length(a)){
+		// just make `to` the length of the array
+		if (to >= length(a) - 1){
 			to = length(a);
 		}
 		


evancz:
Can you explain a bit more? It seems like the function has certain behavior as is, and you are changing that behavior. Some folks have suggested negative indexes for various functions. I think it'd be helpful to have a more complete explanation of what exactly is changing here and how it fits into other things.

eeue56:
Sure -

The existing behaviour is maintained. The only change here is that when Array.slice is given arguments that goes over the length of the Array passed in, they default to the length of the array instead.

Consider the following -

```
Array.slice 1 (Array.length array + 1) array 
```

where `Array.length array` is > 32.

Because the `array` has a length greater than `M`, which is `32`, the array height is  greater than 0, meaning it skips the tail-case for recursion in Array.sliceRight and Array.sliceLeft -

```
// Handle leaf level.
		if (a.height === 0)
		{
			var newA = { ctor:'_Array', height:0 };
			newA.table = a.table.slice(0, to);
			return newA;
		}
```

When the sliceRight function carries on, it uses

```
		// Slice the right recursively.
		var right = getSlot(to, a);
		var sliced = sliceRight(to - (right > 0 ? a.lengths[right - 1] : 0), a.table[right]);
```

to get the next slice. Here is where the problem is - when `to` is bigger than the length of the array, `getSlot` will return a table index that's greater than the number of tables, causing the next call to `sliceRight` to take an undefined object as a parameter, cause the `object undefined doesn't have property a.height` error seen in #349.

The changes I've made make it so that when `Array.slice` is called, if `to` is too big then it'll just default to the length of the Array - restricting the indexing done by `getSlot` and therefore getting the correct table for passing down to the next recursion level.
The same applies for `from` here too - if `from` begins from past-the-end of the Array, then it just returns an empty array. As far as I could figure from the documentation, this is consistent with the already existing behaviour, and prevents errors with `sliceLeft`.

```
// if the start of the slice goes past the end of the array, 
		// then just return empty
		// -1 since array is 0 based
		if (from >= length(a)){
			return empty;
		}

		// if the end of the slice goes past the end of the, 
		// just make `to` the length of the array
		if (to >= length(a) - 1){
			to = length(a);
		}
```

These changes make it so that:

- Any start index past the length returns an empty Array - this is consistent with current behaviour.
- Any end index past the length replaces the end index with the length - this is also consistent with current behaviour

Another fix could be to make getSlot error out or return a default index when something is over-the-length, but I feel that is far more messy.

mgold:
Old thread, but I'd really like to see this fixed - Array is unreliable enough to discourage its use, and this is a part of that.

The other part is things like negative indices and making the API nicer. One change I'd like to see is [this line](https://github.com/eeue56/core/blob/b942111a2f8e0c7669c69656d76ce24255bc9153/src/Native/Array.js#L438), `<` can become `>`, so a `to` of `0` means "slice until the end of the array" -- assuming that won't break anything, which it might.

eeue56:
@mgold - I have plans to rewrite Array with the NRI team from scratch. There's too many bugs in the current implementation to track them all down safely, and the code isn't documented enough to make a sensible approach on it.

mgold:
:clap: :+1: Would this be a native module or would you try to do this in Elm?

eeue56:
It'll be more pure Elm than the current Array implementation, but a lot of it will have to be Native. It's possible to implement it fully in just Elm, but the performance and size overhead would take a hefty toll on any large array. We'll be sure to benchmark things as we go along, though

:+1: For alternative Random implementations too

mgold:
Sounds good. Can I request `singleton` and `reverse` functions, and support for `(++)`? :pray:

evancz:
I think it makes sense to focus on the rewrite in Elm.

PR #400, opened 2015-09-09
Fix touches from erroring when running in embedded mode

### LINKED ISSUE #390: Touch events outside of container div throws exception
Consuming Touch.taps causes an exception to be thrown when an Elm moduile is embedded in an HTML page and you tap (or click) outside the container div.

"Uncaught TypeError: Cannot read property 't' of undefined" occurs in the end(e) function.

eeue56:
Fixes #390 by making sure that the end event target started within the node we care about - the embedded target.


evancz:
It's not like this anymore.

PR #405, opened 2015-09-11
Improve collage documentation

jvoigtlaender:
Motivation for the change is here: https://github.com/elm-lang/core/pull/392#issuecomment-138097389.

### DIFF
--- a/src/Graphics/Collage.elm
+++ b/src/Graphics/Collage.elm
@@ -269,9 +269,10 @@ alpha a f =
   { f | alpha = a }
 
 
-{-| Create a collage with a given width and height, from a collection of 2D
-forms. There are no strict positioning relationships between forms, so you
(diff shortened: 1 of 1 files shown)


eeue56:
+1

evancz:
It seems kind of tortured to me. How about this?

> Create a collage with certain dimensions and content. It takes a width and height argument to specify dimensions, and then a list of 2D forms to describe the content.
>
> Unlike with Elements, these 2D forms can be moved and rotated however you like. The forms are drawn in the order of the list, i.e., `collage w h [a, b]` will draw `b` on top of `a`.

eeue56:
Better than the current and the new proposal too, +1

jvoigtlaender:
I have changed the comment in an additional commit here, using Evan's suggestion.

evancz:
Great, thank you!

PR #406, opened 2015-09-11
Improve merge documentation

jvoigtlaender:
Trying to pick up lessons from this discussion: https://github.com/elm-lang/core/pull/392.


eeue56:
Still slightly confusing but I think it's the best minimal change, so +1

PR #408, opened 2015-09-16
Added pure

Fresheyeball:
I find myself writing this function *alot*, perhaps its worth including?

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -298,6 +298,12 @@ concatMap : (a -> List b) -> List a -> List b
 concatMap f list =
   concat (map f list)
 
+{-| Turn a value into a list containing one value. Useful for point free style.
+    pure x == [x]
+-}
+pure : a -> List a
+pure x = [x]
+
 
 {-| Get the sum of the list elements.
 
--- a/src/List.elm
+++ b/src/List.elm
@@ -4,6 +4,7 @@ module List
     , repeat, (::), append, concat, intersperse
     , partition, unzip
     , map, map2, map3, map4, map5
+    , pure
     , filterMap, concatMap, indexedMap
     , foldr, foldl
     , sum, product, maximum, minimum, all, any, scanl


evancz:
I'd rename it to `singleton` to be parallel with `Dict` and `Set`. I'll add this to the meta issue about core changes.

evancz:
The process for adding stuff to core is described [here](https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions), so we'll go through and think about all these things in a big batch.

Fresheyeball:
Ok. Fair enough.

Fresheyeball:
https://github.com/circuithub/elm-list-extra/pull/15 merged

evancz:
Can you share an example of when you use it? What kind of code calls for this?

Fresheyeball:
basically I use it in composition chains, for example 

https://github.com/Fresheyeball/elm-students/blob/a5ed3ddf557f9a14a2a5f06f0f3d6c171e38ccb2/src/Test.elm#L180

```elm
pure : a -> List a
pure x = [x]

test : List H.Html
test = let
  details =
    pure << H.details [ A.style [ ("overflow", "hidden") ] ]
         << (::) (H.summary [] [ H.text "Tests" ])
         << pure << display << quickCheck << suite "Students"
  in details
    [ checkupdate
    , checkMetrics ]
```

evancz:
I ask because that's how I'd use it as well, but I was wondering if this is ever nicer code than just naming things. Your example would become:

```elm
test : List H.Html
test = let
  toDetails checks =
    H.details
      [ A.style [ ("overflow", "hidden") ] ]
      [ H.summary [] [ H.text "Tests" ]
      , display (quickCheck (suite "Students" checks))
      ]
  in
    [ toDetails
        [ checkUpdate
        , checkMetrics
        ]
    ]
```

I personally think that is better code. Do you have any other examples?

Fresheyeball:
I think you could make a similar criticism of other code using pure, and its fair enough. That said I also find `pure` useful in development workflow even if I remove it later (which usually happens).

evancz:
Okay, thanks for sharing the example, that was very helpful!

Fresheyeball:
Cheers!

jvoigtlaender:
If I may chime in: There are uses of `pure`/`List.singleton` in `start-app` [here](https://github.com/evancz/start-app/blob/f72265a0f351dac624faec6661a273cfbde4aa15/src/StartApp.elm#L90) and [here](https://github.com/evancz/start-app/blob/f72265a0f351dac624faec6661a273cfbde4aa15/src/StartApp.elm#L105) that are not as easily replaced as exercised for the above example. It seems these are conceptually different uses than the ones you discussed above.

Fresheyeball:
While I'm still in favor of `List.singleton`, it can be easily replaced in `start-app`:

```
        inputs =
            Signal.mergeMany (messages.signal :: List.map (Signal.map singleton) config.inputs)
```

vs

```
        inputs =
            Signal.mergeMany (messages.signal :: List.map (Signal.map (\x -> [x])) config.inputs)
```

jvoigtlaender:
Sure. :smile:

Actually, when I wrote that code I defined it as `singleton`, because I was hoping that at some point `List.singleton` would be moved from http://package.elm-lang.org/packages/circuithub/elm-list-extra to `core`, at which time I would come back to this code and use that function.

jvoigtlaender:
So what I was trying to get at above was that that in these lines we really want `singleton` to be a function explicitly passed to another function, not inlining it like in the other examples above.

Fresheyeball:
Agreed :+1:

PR #410, opened 2015-09-17
Update Decode.elm

simonh1000:
Added example based on my pain, and max/Janis help

### DIFF
--- a/src/Json/Decode.elm
+++ b/src/Json/Decode.elm
@@ -512,6 +512,10 @@ succeed =
 
 {-| Handle an array with exactly one element.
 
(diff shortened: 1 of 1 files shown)


evancz:
Can you make the code look like this:

```elm
extractString : Decoder String
extractString =
    tuple1 identity string
```

Then it should be all set!

jvoigtlaender:
Since the second commit addresses @evancz's point about formatting of the code, I'm merging this pull request.

evancz:
Great, thanks @simonh1000 and @jvoigtlaender!

PR #413, opened 2015-09-28
Add Char.toInt with tests

bradurani:
Adds `Char.toInt` which seems like it should be there based on what's in the other classes. I discovered there's no easy way to map a `Char` such as `'1'` to the `Int` `1`. You'd think `'1' |> toString |> String.toInt` would do it but `toString '1'` returns `"'1'"`!
I can't find an easy way to convert a `Char` to a `String` at all. I thought about adding a method for that, but I wouldn't know what to call it, since `toString` is taken.

### DIFF
--- a/src/Char.elm
+++ b/src/Char.elm
@@ -20,7 +20,7 @@ module Char
 
 import Native.Char
 import Basics exposing ((&&), (||), (>=), (<=))
-
+import Result exposing (Result)
 
 isBetween : Char -> Char -> Char -> Bool
 isBetween low high char =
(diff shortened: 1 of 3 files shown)


jvoigtlaender:
About getting from `Char` to `Int`: Isn't that what `Char.toCode` is for? (Or at the very least, you could implement your `Char.toInt` in Elm by using the existing `Char.toCode` without doing any native stuff as you do in your pull request.)

About getting from a `Char` to a `String`: It seems you have overlooked the existence of `String.fromChar`.

jvoigtlaender:
I'm closing this issue.

@bradurani, if you want to propose addition of `Char.toInt`, please:
* Do so via a pull request to https://github.com/zeckalpha/char-extra. This is the current standard practice of suggesting completely new stuff (rather than fixing existing stuff) for `core`: collecting it into `*-extra` libraries first.
* Try doing it without using native stuff. The existing `Char.toCode` should enable you to implement `Char.toInt` in pure Elm.

Concerning "`Char.toString`", see my previous comment.

bradurani:
Makes since. I'm getting the hang of all this. Thanks for explaining @jvoigtlaender

Issue #412, opened 2015-09-26
Time type alias defined in different places


jvoigtlaender:
There is one definition of
```elm
type alias Time = Float
```
in `Time.elm` and an identical one in `Task.elm`.

This seems conceptually wrong, and potentially hurtful for later developments. So I think it would be better to have that definition only in one place, and import it everywhere else it is needed.

What would be options for this?
* One natural place to put the single definition of `Time` would be `Time.elm`, but that is problematic, since `Time.elm` imports `Signal.elm`, which imports `Task.elm`, so having `Task.elm` import `Time.elm` would lead to a cycle.
* The definition could be put into `Task.elm` and `Time.elm` could import it from there. It seems a bit odd, though, to have the primary definition of `Time` (and the place where it is documented) be in `Task.elm`.
* The definition could be put into `Basics.elm` and be imported into `Time.elm` and `Task.elm` from there.

Is anything wrong with the third option? It would essentially give `Time` the same status as `Int`, `Float`, `String` as pre-defined Elm types. It seems that is an okay thing to happen.

mgold:
`Task` does not expose the `Time` alias, and the compiler can figure out what type it wants, so I don't see much of a downside to the current arrangement. I suppose I could get behind placing it in `Basics`, if that wouldn't make `Task` and `Time` much more complicated, but again I don't see the need to.

jvoigtlaender:
It's not a big issue, of course. More a matter of principle and style and looking into possible futures. Consider:
* At some point it is decided that `Time` should not be `Float`, but maybe `Int` or even some completely other type. Then there will be two places where this has to be changed. No big problem, I'll grant, because at least these two places are both in the same package `core`, so the compiler would likely catch this before release. (Actually, it would only catch it if there is some code in `core` in which `Task.Time` and `Time.Time` are unified. It might be the case that there is no such code. Then, `core` would compile just fine and could be released, but clients couldn't use the modules as intended, because they wouldn't be able to pass `Time`s created with functions from the `Time` module to functions from the `Task` module that expect a `Time`.)
[...]

PR #415, opened 2015-09-29
Move Time type alias(es) to Basics module

jvoigtlaender:
Follow-up to the discussion in https://github.com/elm-lang/core/issues/412.


jvoigtlaender:
Do this now, so it doesn't need a jump for core from 3.0.0 to 4.0.0 too soon?

jvoigtlaender:
@evancz: An additional concrete benefit of this change (apart from what we discussed in the other thread linked above) would be that `Time` in the type signature at http://package.elm-lang.org/packages/elm-lang/core/latest/Task#sleep would then properly link to a definition of `Time` (which currently it does not).

PR #417, opened 2015-10-03
Provide a binding to the .toFixed API

agrafix:
Convert a float into a string including a fixed number of decimal places. I think this is a very frequently used function when building user-interfaces so it might be good to but this in the core libraries.

### DIFF
--- a/src/Basics.elm
+++ b/src/Basics.elm
@@ -446,6 +446,17 @@ isInfinite : Float -> Bool
 isInfinite =
   Native.Basics.isInfinite
 
+{-| Convert a float into a string including a fixed number of decimal places.
+
+    toFixedRep 45.0 0 == "45"
+    toFixedRep 3.141592 2 == "3.14"
+    toFixedrep 3.141592 10 == "3.1415920000"
+
+-}
+toFixedRep : Float -> Int -> String
+toFixedRep =
+  Native.Basics.toFixedRep
+
 
 {-| Turn any kind of value into a string. When you view the resulting string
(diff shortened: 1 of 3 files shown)


rgrempel:
According to the [Mozilla documentation of `toFixed`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed), it is possible for Javascript's `toFixed` to throw exceptions. I suppose one wouldn't have to worry about the `TypeError`, since the Elm type checker will save us from that. However, there is the `RangeError` to consider.

So, to guarantee no runtime errors, one would have to make the return type something like a `Result String String` instead of just returning a string. Here's a version of the native code that returns a `Result`:

https://github.com/rgrempel/elm-web-api/blob/master/src/Native/WebAPI/Number.js#L31-L37

and what the interface might look like:

https://github.com/rgrempel/elm-web-api/blob/master/src/WebAPI/Number.elm#L75-L81

rgrempel:
I guess the other option, if you wanted to return a simple `String` instead of a `Result`, would be to clamp the number of digits between 0 and 20 -- at least, the Mozilla documentation suggests that an exception will not be thrown in those cases.

evancz:
This is added to #322 and is easy enough to add. Makes sense to make the decision before writing the code.

witoldsz:
@evancz `toFixed` is a really nice to have. I need to format data in a table and seeing this task as closed and referenced in #721 aka "to be done in the future" makes me sad again - yesterday I was in trouble, found the solution: #842 but it's also postponed for the same reason.

PR #418, opened 2015-10-03
Simple string replacement

agrafix:
I think simple string replacement w/o Regex would be a good addition to the core library for a simple search and replace.


jvoigtlaender:
Have you read the last section of https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md, about the protocol for suggesting functions to be added to core?

evancz:
This should be in the next release https://github.com/elm-lang/core/blob/master/src/String.elm#L144-L154

With regex being in a separate package, it makes sense to cover the common case. That way folks do not have to reach for regex or parsers to do this sort of thing.

PR #421, opened 2015-10-09
Add List.zip

dasch:
This function makes it easy to zip together two lists:

```elm
zip [1,2,3] ['a','b'] == [(1,'a'), (2,'b')]
```

It took me a while to figure out that `zip == map2 (,)`. The principle of least surprise would dictate that common things have names.

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -2,7 +2,7 @@ module List
     ( isEmpty, length, reverse, member
     , head, tail, filter, take, drop
     , repeat, (::), append, concat, intersperse
-    , partition, unzip
+    , partition, zip, unzip
     , map, map2, map3, map4, map5
     , filterMap, concatMap, indexedMap
     , foldr, foldl
@@ -20,7 +20,7 @@ list must have the same type.
 @docs head, tail, filter, take, drop
 
 # Putting Lists Together
-@docs repeat, (::), append, concat, intersperse
+@docs repeat, (::), append, concat, intersperse, zip
 
 # Taking Lists Apart
 @docs partition, unzip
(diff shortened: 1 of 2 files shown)


jvoigtlaender:
@dasch, please read https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions for the protocol concerning suggestions of new functions for `core`.

In the specific case of `zip`, you need not even start a new `list-extras`. The function is already here: http://package.elm-lang.org/packages/circuithub/elm-list-extra/3.7.0/List-Extra#zip

dasch:
Okay - how would the process look for suggesting getting that function
moved into core, then? The CONTRIBUTING.md file seems to leave that part as
TBD...

As a newcomer, this is sort of frustrating.

jvoigtlaender:
This is really for @evancz to answer, but these days and I guess at least another week or so he will be very busy with the Elm 0.16 release.

So I can only say what I learned in other conversations, various tickets here on the `core` repo, the mailing lists etc.:

There is going to be a larger overhaul of the whole of `core` at some point, possibly for Elm 0.17 or Elm 0.18. At that point, the various `*-extras` packages will be taken into account. And I guess there will also be general discussions in this repo and on the mailing list at that time, to make coherent decisions.

So the process for suggesting getting that function moved into `core` is to get it into a `*-extras` package. That can be a new one you create yourself, or it can be an existing one by someone else to which you make a pull request (for example, https://github.com/Apanatshka/elm-signal-extra is being developed that way, by collecting `Signal`-related extra functions by various people). Or, as I pointed out for the specific case of `zip`, there may be no need to do anything at all, because there is already such a `*-extras` package (ready for consideration when the time for the `core` overhaul comes) that contains it.

Hmm, I think I have more or less just repeated what is already in my earlier comment and the link I provided there. So if this is not enough information about the process, what additional information do you need? What should be added to https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions to make things clearer?

dasch:
Thanks for taking the time to respond – I guess your description of the process i simply better :-)

PR #424, opened 2015-10-11
Stop toString from descending into arbitrary objects

ThomasWeiser:
The function `toString` has code to convert Elm-records into a string representation.
 This code missed a checked whether the JS object is really an Elm-record.

 Problem: If toString is called on an recursively linked JS object (not an Elm-record, e.g. from a native library function), this will lead into an infinite recursion.

 The check used to be in the code up to commit 08bf8095e24feefc1b13acd7e428081e6d5c20c9. I don't know if Evan removed it intentionally for some reason.

### DIFF
--- a/src/Native/Utils.js
+++ b/src/Native/Utils.js
@@ -460,7 +460,7 @@ Elm.Native.Utils.make = function(localRuntime) {
 		{
 			return '<Signal>';
 		}
-		else if (type === 'object' && probablyPublic(v))
+		else if (type === 'object' && '_' in v && probablyPublic(v))
 		{
 			var output = [];
 			for (var k in v)


evancz:
Yes, it was removed because records no longer have a `_` field that distinguishes them.

Can you give concrete examples of values that mess this up?

ThomasWeiser:
Ah, didn't know about the dropped `_` field.

My example is a test app for elmfire. The native code returns a reference to internal data of the Firebase library. This data happens to have a cyclic structure.

I now have a workaround, where I [drop](https://github.com/ThomasWeiser/elmfire/blob/8ace6e543d6c6bc0eb1d54dfc53466b84f205e92/test/src/Tests.elm#L67-L75) this cyclic Json data from the record before printing.

Does this mean, that cyclic structures are disallowed as a `Json.Encode.Value`? This would be kind of a burden for native wrappers around JS libraries. But it would be a valid argument to not have to do cycle checks in functions like `toString` (and probably in comparing and equality checking too).

PR #428, opened 2015-10-21
Missing comma in example code

kitofr:
I think there's a comma missing here in the example.
This was derived from 2.1.0, but didn't have access to create PR against it.


evancz:
Nice catch! I sometimes wish the compiler could check code in comments, but then I think about implementing that :P

Anyway, I ended up taking out that big starting comment in favor of more targeted comments throughout. Thanks for the fix though!

kitofr:
@evancz yes that IS nice :) elixir has doc-tests for that purpose which would be just awesome in ELM as well.

PR #429, opened 2015-10-27
Fix link

jvoigtlaender:


evancz:
Thanks!

And I think I'm going to expedite Pete's reactor work, so we'll be getting a new version of reactor and core relatively soon I hope.

PR #439, opened 2015-11-20
punctuation fix

jvoigtlaender:


jvoigtlaender:
Why close this? Are you fixing it elsewhere? Because the text as currently written *is* broken. 

It reads, at http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Mouse#isDown, as follows: "The current state of the mouse True when any mouse button is down, and false otherwise."

Surely that needs to be "The current state of the mouse. True when any mouse button is down, and false otherwise." instead.

evancz:
My mistake, not sure what happened.

Issue #438, opened 2015-11-19
"up.props not defined" error in elm reactor 0.16


fredcy:
The customButton native code tries to access `up.props.width` but `up.props` is not defined. Looks like an intermediate `_0` object was introduced between `up` and the `props` property.

fredcy:
@evancz This problem manifests when running elm reactor in 0.16 but looks to me to be a significant problem in core/src/Native/Graphics. Some recent change added a wrapper layer in Element named `_0` but there is still code around expecting to access properties like `props` directly from the element. Case in point is the `customButton` function mentioned above.

jvoigtlaender:
Is this still relevant or was it fixed by https://github.com/elm-lang/core/commit/09f3dd2b11f34987bf6b32201fd1899baff5d646?

fredcy:
It was fixed by the latter. I'm closing this now.

PR #440, opened 2015-11-20
Unwrap customButton inputs to get props from _0 object

fredcy:
This change fixes https://github.com/elm-lang/core/issues/438 and is part of fixing the problems reported in https://github.com/elm-lang/elm-reactor/issues/156 and https://github.com/elm-lang/elm-reactor/issues/157.


evancz:
Ah, dang, I did not see this earlier! I did the same fix today: https://github.com/elm-lang/core/commit/09f3dd2b11f34987bf6b32201fd1899baff5d646

evancz:
Thanks for hunting this down!

PR #446, opened 2015-11-23
Update tests for 0.16

klaftertief:
I took the liberty to update the tests for 0.16, hope this is ok. This includes

- updated dependencies
- updated `run-test.sh` to use `elm-console/elm-io.sh`
- removed `Basics.otherwise` test.

I disabled the `letQualified` test in `CodeGen.elm` because it did not compile and I din't know exactly how to fix it.


jvoigtlaender:
@klaftertief, do you know why there is still a build failure with your PR? @deadfoxygrandpa did similar work later in #447. Probably only one of the two should be kept open. If you are still interested in this, maybe you could take a look at his PR and see whether he covers all you did? And if not, bring it up over there so that we can end up with a single PR that hopefully addresses the test build issue once and for all?

klaftertief:
The build failure is because of [cabal hell](https://travis-ci.org/elm-lang/core/builds/92681162#L210). This is why @deadfoxygrandpa [switched to the npm installer](https://github.com/deadfoxygrandpa/core/commit/281bd3fd07a5b4c28a23b99b6dbb528cc05624ab).

The other changes, updating the test runner and the tests itself, are basically the same. So I'm closing this pull request in favour of #447 in the assumption that using the npm installer is the way to go.

jvoigtlaender:
Good to know. Thanks!

PR #450, opened 2015-11-28
Remove <~ from documentation examples

jvoigtlaender:
Several examples were still using this now outlawed signal map operator. I replaced those uses by `Signal.map`, so that the examples actually work with the current version of `core`.


jvoigtlaender:
@evancz, I think this is an important documentation fix. We want example code to actually work.

Issue #419, opened 2015-10-06
Graphics.Collage missing sprite export


keenbug:
There is a definition for `sprite` in src/Graphics/Collage.elm but it is neither used there nor in the export list.

evancz:
It is not publicly exposed for now. There was some bug in Chrome that caused the images to skip the cache and flicker very badly. So it needs work and testing to make sure it is in shape for release. I'll add this to the meta issue about `Graphics.*`

Issue #332, opened 2015-08-06
Aggregating issues about Graphics.*


evancz:
**Do not add issues or ideas as comments here. Open a separate issue for that.**

A bunch of issues and ideas have come in about `Graphics.*` over time and I'm trying to aggregate them all in one place.

**Issues:**

  * https://github.com/elm-lang/elm-compiler/issues/805 - bug with `email`
  * https://github.com/elm-lang/elm-compiler/issues/835, https://github.com/elm-lang/elm-compiler/issues/727 - events swallowed on `<input>` elements
  * https://github.com/elm-lang/elm-compiler/issues/800 - centering of `<input>` elements
  * https://github.com/elm-lang/elm-compiler/issues/867 - flow right inputs inside of flow down
  * https://github.com/elm-lang/elm-compiler/issues/890 - clicking `Html` in elements
  * https://github.com/elm-lang/elm-compiler/issues/775 - bug with `clickable`
  * https://github.com/elm-lang/elm-compiler/issues/773 - pixel mismatch
  * #334 - something weird with clicks

**API Ideas:**

  * https://github.com/elm-lang/elm-compiler/pull/601 - add `Graphics.Input.slider`
  * https://github.com/elm-lang/elm-compiler/issues/683 - quadratic curves in `collage`
  * https://github.com/elm-lang/elm-compiler/issues/680 - rounded corners somehow
  * https://github.com/elm-lang/elm-compiler/issues/533, https://github.com/elm-lang/core/pull/307 - get local mouse position
  * #419 - add `sprite` back

mgold:
I'll add:
* New shapes in `Graphics.Collage`, most notably `arc`. Part of the problem is when an arc is `traced`, do you see just part of a circle, or does it go into the center like a slice of pie?

* Collision detection, either with [backbuffers](https://groups.google.com/d/msg/elm-discuss/Ee2_MfauicE/_635u_oJ4_UJ) or geometrically.

PR #456, opened 2015-12-04
Fix image flickering in Graphics.Collage.sprite

nphollon:
I am resurrecting @silverio's pull request #191 from earlier this year. This fixes the performance problem with sprite sheets (see #419 and #332).

[Here is a comparison that highlights the improvements.](https://nphollon.github.io/sprite_animation.html) Each of these examples animates [this sprite](https://nphollon.github.io/sprite_animation/fireflower.png) at 30 fps.

The first example uses `Graphics.Element.croppedImage`. This function inserts an `img` node into the DOM. The position of the cropping frame gets set in an `onload` callback. This means that for a moment when the image is loaded, the crop frame is positioned at (0, 0). Leads to a fierce flicker.

The second example uses the current implementation of `sprite`, which lurks unpublished in `Graphics.Collage`. On Chrome, I see the frame-rate plummet after about a second. On Firefox, the animation simply doesn't work.

The third example uses the `sprite` implementation in this pull request. It works great on my machine in both Firefox and Chrome. No flickering or drop in fps.

What do other people see?

EDIT: Fixed hyperlink to the examples page

### DIFF
--- a/src/Graphics/Collage.elm
+++ b/src/Graphics/Collage.elm
@@ -1,6 +1,7 @@
 module Graphics.Collage
     ( collage, Form
-    , toForm, filled, textured, gradient, outlined, traced, text, outlinedText
+    , toForm, sprite, filled, textured, gradient, outlined, traced
+    , text, outlinedText
     , move, moveX, moveY, scale, rotate, alpha
     , group, groupTransform
     , Shape, rect, oval, square, circle, ngon, polygon
@@ -20,7 +21,7 @@ so moving a form 10 units in the y-axis will move it up on screen.
 @docs collage, Form
 
 # Creating Forms
-@docs toForm, filled, textured, gradient, outlined, traced, text, outlinedText
+@docs toForm, sprite, filled, textured, gradient, outlined, traced, text, outlinedText
 
 # Transforming Forms
 @docs move, moveX, moveY, scale, rotate, alpha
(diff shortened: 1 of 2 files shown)


thisredone:
This is **not** the only place this bug is present.
You might want to also cache the Graphics.Collage.textured
`renderForm` with `form.ctor` of `FShape` calls `drawShape` which calls `setFillStyle` which then calls the `texture` function which also creates a `new Image` every time.
What's even worse is that it seems that every time an image is created the `img.onload` forces the update which then creates even more images which do the same over and over again.

It's surprising that this issue has been present for so long. It makes images in Graphics.Collage effectively unusable (and most likely quite frustrating).

Note: the title of this PR doesn't reflect the nature of the problem. It reads like a feature addition but it's actually more like a major bug fix.

nphollon:
@thisredone Thanks for pointing that out. I renamed the issue to highlight the problem instead of the solution. I will take a look at `Graphics.Collage.textured` when I get a chance.

villasv:
Indeed this makes Craphics.Collage quite unusable for me, because I was doing an Elm demo that had a rolling background. Except in the Edge browser, the flickering makes the game simply unplayable.

Is there any workaround? Maybe specifically applied to `croppedImage`?

evancz:
Sorry for the slow reply. It has been busy times. All the `Graphics.*` modules have moved to `evancz/elm-graphics` so it makes sense to retarget stuff like this. There is a lot going on still, but I expect to be thinking about improving `<canvas>` support and the learning avenue in the future.

nphollon:
@evancz Should I re-open this PR over on evancz/elm-graphics?

evancz:
Anything that is adding to the public API, I'd rather not add in right now. If there's *also* a bug fix in here, the bug fix should be moved.

PR #458, opened 2015-12-05
Document potential gotcha. Fail is when a 'value', not 'key', can't be decoded

simonh1000:


jvoigtlaender:
Maybe the same information should be added to the documentation of the dict-function as well?

jvoigtlaender:
@simonh1000, why close this?

simonh1000:
Because I got totally confused with Git/Github!
Need to start again from scratch

jvoigtlaender:
Isn't just reopening this pull request here an option? (There's a button with that functionality at the bottom.)

PR #466, opened 2015-12-20
Remove extra paren from Signal docs

jergason:
There are two closing parens in the docs describing the Signal type.

screen shot 2015-12-20 at 9 22 59 am

Now there is only one!


jergason:
NM, there are like 4 other PRs already open to fix xthis.

PR #477, opened 2016-01-09
Allow toString to handle null values

Greatlemer:
When I was trying to use ElmTest to compare two Html elements I began getting problems as there were some null attributes which toString couldn't handle. This change prevents that issue from happening.


jvoigtlaender:
I'm closing this, as a duplicate of https://github.com/elm-lang/core/pull/183. Please comment over there if there is something to add to the discussion of how this should be handled.

PR #481, opened 2016-01-14
Have toString display Date values in ISO format

fredcy:
The core `toString` function returns the String value "{}" for all `Date` values. That makes it hard to display Date values for debugging. This patch extends `toString` to display `Date` values in ISO format like "\<date 2016-01-14T15:06:41.393Z>".

Here is a test program that I've been using. On regular intervals it displays the value of Time.every (to show liveness) and the corresponding Date value:

```elm
import Date
import Effects
import Html exposing (li, text, ul)
import StartApp
import Time

app =
  StartApp.start
    { init = ( 0, Effects.none )
    , inputs = [ Time.every <| Time.second * 5 ]
    , update = \t -> always ( t, Effects.none )
    , view = always showTime
    }

showTime t =
  ul
    []
    [ li [] [ t |> toString |> text ]
    , li [] [ t |> Date.fromTime |> toString |> text ]
    ]
```

### DIFF
--- a/src/Native/Utils.js
+++ b/src/Native/Utils.js
@@ -379,6 +379,9 @@ Elm.Native.Utils.make = function(localRuntime) {
 		{
 			return '"' + addSlashes(v, false) + '"';
 		}
+		else if (v instanceof Date) {
+			return '<date ' + v.toISOString() + '>';
+		}
 		else if (type === 'object' && 'ctor' in v)
 		{
 			if (v.ctor.substring(0, 6) === '_Tuple')


mietek:
This is a duplicate of https://github.com/elm-lang/core/pull/442.

jvoigtlaender:
So, can this pull request be closed? I notice that while it addresses the same issue as #442, it does propose a slightly different code change. Is the one here better than the other one? Why?

fredcy:
I didn't notice https://github.com/elm-lang/core/pull/442 before opening this request. I'm OK with closing this request as a duplicate.

jvoigtlaender:
Okay, I'm closing it, but copying your example program over to the other discussion, because it could be useful for testing over there.

PR #491, opened 2016-01-24
Remove Unicode symbols from documentation

jvoigtlaender:
... because they kill people in situations like [this](https://github.com/elm-lang/elm-make/issues/33#issuecomment-173915831).


jvoigtlaender:
@evancz, this presence of Unicode symbols in the `core` documentation (and the inability of at least some `elm-make` binaries, under some circumstances, to read those symbols in) is hurting people continuously. For no good reason, it seems. (Or are those symbols *that* important to have in the documentation?)

For yet another incident like that, see https://github.com/elm-lang/elm-make/issues/33#issuecomment-181078584.

PR #492, opened 2016-01-25
Added vimKeys as another movement signal

meditans:
Otherwise, if one wants to implement vimKeys by its own, it has to rewrite in his code `dropMap`, `toXY`, and `type alias Directions`, which are not exported.

### DIFF
--- a/src/Keyboard.elm
+++ b/src/Keyboard.elm
@@ -146,6 +146,9 @@ wasd : Signal { x:Int, y:Int }
 wasd =
   dropMap (toXY { up = 87, down = 83, left = 65, right = 68 }) keysDown
 
+vimKeys : Signal { x:Int, y:Int }
+vimKeys =
+  dropMap (toXY { up = 75, down = 74, left = 72, right = 76 }) keysDown
 
 {-| Whether an arbitrary key is pressed. -}
 isDown : KeyCode -> Signal Bool


jvoigtlaender:
Please read this: https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions.

There doesn't seem to be a `keyboard-extras` package around yet, but you could start one.

Also, `dropMap` is just a combination of `Signal.dropRepeats` and `Signal.map`, so no real hindrance to a client programmer, and `Directions`, being a `type alias` rather than a `type`, is also not really hidden away. A client pogrammer can readily use the type `{ up : KeyCode, down : KeyCode, left : KeyCode, right : KeyCode }`. So the only real annoyance for a client programmer who wants `vimKeys` is that they need to reimplement the `toXY` function.

So a viable course of action would be to propose to export that function for client use. That would be a separate issue you may want to open.

The current issue, about adding a completely new function, I am closing in accordance with the policy about new functions linked to above.

jvoigtlaender:
BTW, there was once a more generic function capturing "arrow behavior": http://package.elm-lang.org/packages/elm-lang/core/1.0.0/Keyboard#directions

It was removed when going to `core` version `2.0.0`. I don't know what decision process led to that.

meditans:
I see, thanks for the useful explanations :)

PR #494, opened 2016-01-29
Fix unsubscribing from ports

RGBboy:
Hey, so I found a weeeee bug in the unsubscription of a port. It would only `pop` the last added subscription instead of the one you are actually unsubscribing. Let me know if there is anything else that needs to be done to get this merged.

### DIFF
--- a/src/Native/Port.js
+++ b/src/Native/Port.js
@@ -94,7 +94,7 @@ Elm.Native.Port.make = function(localRuntime) {
 		}
 		function unsubscribe(handler)
 		{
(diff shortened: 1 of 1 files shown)


dfreeman:
I think this still behaves badly in the case where it's given a handler that isn't subscribed, since `array.splice(-1, 1)` removes the last element of `array`.

RGBboy:
@dfreeman nice catch. Fixed in db7efc54963afcffa4b127de8fbe4b6314779947.

evancz:
Thanks! It works this way in the upcoming version of things.

PR #497, opened 2016-02-02
Move functions F2-F9 & A2-A9 outside conditional

robertjlooby:
Using a function statement inside a conditional is non-standard (most browers [other than Firefox](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function#Conditionally_created_functions) will declare the function anyways). Since the only point of having these inside the conditional seems to be to prevent them from being redefined to the same thing if `Runtime.js` is included twice on the same page, it seems safe to move them out from the conditional.

The benefit to doing this is that it allows the [Closure Compiler](https://developers.google.com/closure/?hl=en) to be used on the generated `elm.js` (at least in `WHITESPACE_ONLY` and _possibly_ on `SIMPLE_OPTIMIZATIONS`). A simple "Hello World" app:

```
module Main (..) where

import Graphics.Element exposing (show)


main =
    show "Hello World!"
```

When compiled is 147K.

Minimized on `WHITESPACE_ONLY` it is 96K (and works fine).

Minimized on `SIMPLE_OPTIMIZATIONS` is 61K (and works fine).

On a somewhat larger project of mine that compiles to 562K, making this change also allows `WHITESPACE_ONLY` compilation to 292K (works) and `SIMPLE_OPTIMIZATIONS` compilation to 202K (also works).

It's possible there's something I'm missing, but I don't think there's any downside to moving these outside the conditional.

### DIFF
--- a/src/Native/Runtime.js
+++ b/src/Native/Runtime.js
@@ -500,137 +500,144 @@ if (!Elm.fullscreen) {
 			return true;
 		}
 	}());
+}
 
-	function F2(fun)
-	{
-		function wrapper(a) { return function(b) { return fun(a,b); }; }
-		wrapper.arity = 2;
-		wrapper.func = fun;
-		return wrapper;
-	}
+function F2(fun)
+{
+  function wrapper(a) { return function(b) { return fun(a,b); }; }
+  wrapper.arity = 2;
+  wrapper.func = fun;
+  return wrapper;
+}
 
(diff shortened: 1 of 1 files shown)


evancz:
I'm not sure. It is going to just work different in the next release such that Closure Compiler has a much easier time with things. I've also verified that everything works with simple optimizations in the new version of things.

PR #498, opened 2016-02-06
Fix Mouse.isDown behavior for non-fullscreen components

erich-9:
The signal _Mouse.isDown_ doesn't work as explained in the documentation:

> True when any mouse button is down, and false otherwise.

For embedded components pressing or releasing a mouse button outside the component doesn't change the value of the signal _Mouse.isDown_, even if the mouse is moved back into the component afterwards.

Here is an example (save it as _Main.elm_ and compile `elm-make Main.elm --output=elm.js`):

```elm
module Main (main) where

import Graphics.Element exposing (show)
import Mouse

main = Signal.map show Mouse.isDown
```

Now embed it in an HTML document:
```html
<html>
  <head>
  <style>
    div.fbox { width: 50px; height: 50px; border: solid gray 2px }
  </style>
  </head>
  <body>
    <div id="x" class="fbox"></div>
    <script src="elm.js"></script>
    <script>
      var x = document.getElementById('x');
      Elm.embed(Elm.Main, x);
    </script>
  </body>
</html>
```

### DIFF
--- a/src/Native/Mouse.js
+++ b/src/Native/Mouse.js
@@ -24,10 +24,10 @@ Elm.Native.Mouse.make = function(localRuntime) {
 	localRuntime.addListener([clicks.id], node, 'click', function click() {
 		localRuntime.notify(clicks.id, Utils.Tuple0);
 	});
-	localRuntime.addListener([isDown.id], node, 'mousedown', function down() {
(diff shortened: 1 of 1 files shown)


evancz:
Mouse no longer lives in core. Not sure if it is still an issue in `elm-lang/mouse`.

PR #499, opened 2016-02-06
Set style options of croppedImage before onload

erich-9:
The current implementation of _Graphics.Element.croppedImage_ loads the image from its source. Only afterwards, when the _onload_ event is triggered, the style options are set.

In this commit, I set the options not depending on the image's natural size as early as possible (namely, _marginLeft_ and _marginTop_).

This reduces unpleasant flickering, at least as long as the image is not scaled. Compare with #456.


evancz:

PR #500, opened 2016-02-08
Add keyCode signal for keydown events

justinwoo:
This commit adds a signal for `keyCodes` from the `keydown` event to the exports.

Currently, `presses` are exposed, but they don't capture keyboard arrow key inputs. This makes the natural "key-repeat" `keydown` events unusable unless one either sets up a port or writes their own Native function, but that would be wasteful since the `keydown` event is already listened to in the core Keyboard module.

### DIFF
--- a/src/Keyboard.elm
+++ b/src/Keyboard.elm
@@ -1,7 +1,7 @@
 module Keyboard
     ( arrows, wasd
     , enter, space, ctrl, shift, alt, meta
-    , isDown, keysDown, presses
+    , isDown, keysDown, presses, downs
     ) where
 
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
You'd have to add "downs" to the @docs section as well, or the package won't be publishable. Also, https://github.com/elm-lang/core/pull/463, though probably they don't interfere.

justinwoo:
Ah, thanks.

I think the diff should go through fine with no conflicts, but I could rebase if that one goes in first.

evancz:
With 0.17 this is no longer the responsibility of core.

PR #504, opened 2016-02-19
Handle cursor movement and selection events with Graphics.Input.Field

thebritican:
Currently a `Graphics.Input.Field` does not update its `Selection` on events other than `input`.

The following will send a message whenever the input field's selection range changes or if the caret is moved via a `click`, `keyup`, or `focus`.

### Issues
Many events are fired this way, and since we are mutating the `field` object with `elm_old_*` fields, a very fast typer will notice some jostling with the caret. Looking into this more soon and would love any input/help!

### DIFF
--- a/src/Native/Graphics/Input.js
+++ b/src/Native/Graphics/Input.js
@@ -344,34 +344,53 @@ Elm.Native.Graphics.Input.make = function(localRuntime) {
 
 		field.elm_handler = model.handler;
 		field.elm_old_value = field.value;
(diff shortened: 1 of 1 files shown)


evancz:
Sorry for the slow reply. It has been busy times. The `Graphics.Input.Field` module did not make it into `evancz/elm-graphics` so there is nothing to fix anymore. It is recommended to use the `elm-lang/html` text fields now.

thebritican:
Glad to hear it!

PR #507, opened 2016-02-25
Debug.log: remove dead code, always use console.log

jonathanperret:
The code to use `process.stdout.write` that was introduced in https://github.com/elm-lang/elm-compiler/commit/142c2c likely never worked: the `var x = x || {}` trick only ever works for variables already defined in the same scope as the `var` statement, which `process` definitely is not under Node.

Additionally, `console.log` works just fine in Node, and adds a newline where `process.stdout.write` does not (which would have been immediately obvious if that code path had ever been taken).

### DIFF
--- a/src/Native/Debug.js
+++ b/src/Native/Debug.js
@@ -12,15 +12,7 @@ Elm.Native.Debug.make = function(localRuntime) {
 	function log(tag, value)
 	{
 		var msg = tag + ': ' + toString(value);
-		var process = process || {};
-		if (process.stdout)
-		{
-			process.stdout.write(msg);
-		}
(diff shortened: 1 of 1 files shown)


evancz:
I don't know if this changed since the code was changed away from just calling `console.log` directly in the first place. Maybe it has. I am changing how logging works a bit in 0.19 so I will review it then and verify what is necessary.

Thank you for bringing this to my attention!

jonathanperret:
You did remove the problematic code in f4976fb299105f8e13fc9ffe0569c4df708a4d73 so this PR is indeed no longer useful.

PR #447, opened 2015-11-24
Update tests for Elm 0.16

deadfoxygrandpa:
I don't think the core library tests need to depend on building Elm Platform from source. The build has been erroring out on travis for quite some time. I'm going to try updating everything relevant to get the tests working again.

PR #508, opened 2016-02-25
Ignore tests/raw-test.js generated for tests

jonathanperret:
Following the merge of #447, a file `tests/raw-test.js` is generated during a test run. It should be ignored by Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 elm-stuff
 tests/test.js
+tests/raw-test.js


evancz:
I don't think the testing pipeline works this way anymore. @eeue56 may know better. I am going to close assuming I am correct. Let's open a new one if I am mistaken.

jonathanperret:
You're right, 600e1d11806a7672c3559843583a53aa35c7e8fc removed `raw-test.js`. Good catch.

PR #509, opened 2016-02-26
Travis build: use a node_js container

jonathanperret:
This could shave a bit off the Travis build times.


jonathanperret:
This seems to be the last obstacle to returning Travis builds to green. So, ping @evancz hopefully?

PR #517, opened 2016-03-03
Update example decoder name to match its type

joshuaclayton:
This updates the decoder name in the `object3` `Json.Decode` example. I believe `point` was there previously because of the documentation for `object2`.


jvoigtlaender:
Strongly an improvement!

PR #521, opened 2016-03-14
Use List.foldr in Task.sequence

sgillis:
Using foldl instead of a recursive function prevents that we reach the maximum call stack size when sequencing a lot of tasks.

### DIFF
--- a/src/Task.elm
+++ b/src/Task.elm
@@ -145,12 +145,7 @@ This can be useful if you need to make a bunch of HTTP requests one-by-one.
 -}
 sequence : List (Task x a) -> Task x (List a)
 sequence tasks =
-  case tasks of
-    [] ->
-        succeed []
-
-    task :: remainingTasks ->
-        map2 (::) task (sequence remainingTasks)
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
But you are returning the results in the wrong order. That is, the result list will be reversed in comparison to the current implementation.

jvoigtlaender:
Use `List.foldr` instead, and you will also get a non-recursive implementation (using JavaScript looping), but retain correct semantics.

jvoigtlaender:
Of course, the `foldr` based solution will share the "maximal call stack size" problem with the current recursive implementation. So maybe the best solution would be to use `List.foldl` and afterwards a `Task.map` with `List.reverse`.

jvoigtlaender:
Somewhat related (in terms of solution): https://github.com/elm-lang/core/issues/601.

turboMaCk:
This seems like iprovement to me. Original implementation does not use tail call. What is state of this?

evancz:
I ended up merging #881 because that's the one I saw first after `List.foldr` was reimplemented in Elm.

PR #523, opened 2016-03-14
Improve List.scanl

jvoigtlaender:


evancz:
I messed up with this branch. Why is this better? Is it faster in benchmarks? If so, please reopen.

Sorry for the trouble with branches.

jvoigtlaender:
Reopened as a branch in my fork, with more explanation: https://github.com/elm-lang/core/pull/574.

Issue #520, opened 2016-03-14
Using two Keyboard signals, the second did not work


jessitron:
Using both Keyboard.keysDown and Keyboard.ctrl, mapping then merging them to get my input signal: the second one (Keyboard.ctrl) did not fire. Replacing Keyboard.keysDown with a constant signal caused Keyboard.ctrl to work. 

Is this expected? Can only one of the signals that internally feeds off keysDown be used?
If so, documentation would be nice,

I'm in Safari on a Mac.

jvoigtlaender:
Isn't the last sentence of the documentation of http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Signal#merge clear enough on this?

jvoigtlaender:
You may want to use http://package.elm-lang.org/packages/Apanatshka/elm-signal-extra/5.7.0/Signal-Extra#fairMerge.

jessitron:
Thanks for the references.

It is not clear that keysDown and ctrl are the same signal underneath. (I can see it in the source, but not the docs)

jvoigtlaender:
Yes, it's nowhere specified in the docs which signals come from the same origin.

Do you have a proposal for how to do this? For the `Keyboard` module or in general?

jvoigtlaender:
No reply from @jessitron. Am closing this.

jessitron:
Sorry I missed this notification.
It's OK to close it, it isn't specifically a bug, but it would be nice if
there was a comment at the top of the Keyboard file indicating which
signals are mutually exclusive.

On Sun, Mar 20, 2016 at 6:43 PM, Janis Voigtländer <notifications@github.com

jvoigtlaender:
@jessitron, I try to address this in https://github.com/elm-lang/core/pull/532 now.

jessitron:
\o/

PR #532, opened 2016-03-21
Clarify joint behavior of Keyboard signals

jvoigtlaender:
Addressing https://github.com/elm-lang/core/issues/520.


evancz:
It's not going to be the same in the future.

Issue #540, opened 2016-03-27
Possible dead code in Native.Graphics.Element


rgrempel:
There appears to be a little bit of dead code here, in version 3.0.0.

https://github.com/elm-lang/core/blob/3.0.0/src/Native/Graphics/Element.js#L504-L508

At least, I don't think it is doing anything -- it's always possible that I'm missing something.

Note that it does not cause any tangible problems, as far as I can tell. So, it would not be a high priority to fix -- just thought I'd record it in case it was useful at an appropriate moment.

halfzebra:
Created a pilot PR for this issue #544

evancz:
Sorry for the slow reply. It has been busy times. All the `Graphics.*` modules have moved to `evancz/elm-graphics` so it makes sense to retarget stuff like this.

I was not immediately confident that it was doing nothing. I'd want stronger assurances before making this change. Whoo JavaScript! ;)

PR #544, opened 2016-03-30
Removed dead code from Native/Graphics/Element

halfzebra:
#540 removed dead code from `src/Native/Graphics/Element.js`


evancz:

PR #545, opened 2016-04-01
Added warning to `String.padLeft` doc comment.

tymorrow:
Never hurts to cover all of our bases.


jvoigtlaender:
I get the joke, but will you close the issue when 1st April is over?

PR #546, opened 2016-04-02
Add section on setting up and running tests locally in README.md

vikram7:
I spent a bit of time trying to figure out how to run tests locally, and thought adding a small section on it in the README file might help others who may have the same issue.


evancz:
The path to running tests changes, so I'm not sure this is correct (or will stay correct) so I do not want to add instructions that may mislead people later.

In the long run, it should be true that you can just call `elm test` from the root of any Elm project.

PR #556, opened 2016-04-07
Updates (::) documentation: Adds one more example.

rebelwarrior:
This makes clear that to call the `(::)` as a function you need to wrap the `::` in parenthesis by adding another example.
The example uses `(::)` as a function so it shows how it *would* be used with a pipe operator. 
--This example doesn't use the pipe operator itself `|>` so it's much simpler than the previous one I submitted.


rebelwarrior:
@jvoigtlaender Janis, this one is another documentation improvement. Please consider.

evancz:
I'd rather not have this in the docs here. Beginners are likely to see `1 :: [2,3]` pretty early on with Elm, whereas knowing that infix operators can be used as functions is kind of advanced. I suspect that adding this will be more confusing for beginners who do not need to know about this feature that is not directly related to cons.

rebelwarrior:
There is one reason I disagree. 

The pipe operator. 

If you don't know you can do that you can't chain cons in a pipe operation. 

It is rather advanced so maybe we can put it in another place?

evancz:
I mean, it should be in the docs somewhere that you can use infix operators as functions. Not here though.

Issue #502, opened 2016-02-12
Documentation suggests that `==` works on tasks, but it doesn't


jvoigtlaender:
For example, the documentation of `Task.map2` (http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Task#map2) contains this line:
```elm
map2 (+) (succeed 9) (succeed 3) == succeed 12
```
However, issuing that exact line in a program will actually evaluate to `False`. Very confusing, for example to @paparga [here](https://github.com/elm-lang/core/issues/501).

Maybe the documentation shouldn't be using `==` on tasks, and indeed maybe it should explicitly mention that `==` is wonky on tasks, as the documentation already says elsewhere about `Dict` etc.

(Note: For `Dict` etc., equality could potentially be made unwonky, but for tasks not.)

halfzebra:
Here's my take on fixing those confusing in docs #557

jvoigtlaender:
Thanks, I like your changes there!

I'll copy my original comment from above to the opening comment over there, so that Evan sees it all in one place when reviewing the PR.

Issue #501, opened 2016-02-11
Task.map2 example is not working


paparga:
Hi!

I tried Task.map2 example 
`succeed 5 == map2 (+) (succeed 2) (succeed 3)`
and it returns `False`

Is this ok?

Pablo

jvoigtlaender:
See [this comment](https://github.com/elm-lang/core/issues/403#issuecomment-152402738) of @evancz. Equality on `Task`s is in the same league as the others mentioned there. So: `Task.map2` *is* working, `==` on tasks *isn't*, and shouldn't be expected to.

jvoigtlaender:
Closing this issue, opened another one pointing out the real problem more explicitly: https://github.com/elm-lang/core/issues/502.

PR #557, opened 2016-04-10
Corrections for Task documentation concerning use of ==

halfzebra:
Corrections for Task module documentation.

The relevant issue #502 was:
> **Documentation suggests that `==` works on tasks, but it doesn't**
> 
> For example, the documentation of `Task.map2` (http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Task#map2) contains this line:
> ```elm
> map2 (+) (succeed 9) (succeed 3) == succeed 12
> ```
> However, issuing that exact line in a program will actually evaluate to `False`. Very confusing, for example to @paparga [here](https://github.com/elm-lang/core/issues/501).
> 
> Maybe the documentation shouldn't be using `==` on tasks, and indeed maybe it should explicitly mention that `==` is wonky on tasks, as the documentation already says elsewhere about `Dict` etc.
> 
> (Note: For `Dict` etc., equality could potentially be made unwonky, but for tasks not.)

I have replaced all equality operators == with comment -- to standardize the style of the docs(before it had both in different places)

Also I have added a note, that == shoudn't be used with Tasks.

### DIFF
--- a/src/Task.elm
+++ b/src/Task.elm
@@ -11,6 +11,8 @@ module Task
 {-| Tasks make it easy to describe asynchronous operations that may fail, like HTTP requests or writing to a database.
 For more information, see the [Elm documentation on Tasks](http://elm-lang.org/guide/reactivity#tasks).
 
+Task equality with `(==)` is unreliable and should not be used.
+
 # Basics
 @docs Task, succeed, fail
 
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
An example of where already the current docs use -- instead of == is the documentation of `onError`, which has these lines:
```elm
fail "file not found" `onError` (\msg -> succeed 42) -- succeed 42
succeed 9 `onError` (\msg -> succeed 42)             -- succeed 9
```
This PR fixes the other examples on that doc page to be given in the same style (and thus avoiding confusing use of `==` on tasks).

evancz:
Thanks, makes sense!

Issue #369, opened 2015-08-25
Result instead of Maybe for List.head and the like


jvoigtlaender:
Continuing from [here](https://github.com/circuithub/elm-maybe-extra/pull/5#issuecomment-134462627).

`List.head` currently has type `List a -> Maybe a`, and similarly `Maybe` appears in return types of other functions like `Array.get`.

Seeing the `Maybe` as signalling partiality of these functions (e.g., sometimes there is no head, so the function would fail, but instead is signalling the partiality in its return type for explicit treatment by the consumer), one wonders why not `Result` is used instead. (Keeping `Maybe` for characteristic uses of a data structure nature, rather than of a partial function nature. E.g., an optional value from a `.json` file would still be represented as a `Maybe`.)

One issue with this is performance: Does it have a bad impact on performance in typical uses of the function (maybe many of them throughout some recursion in a program) if the type of `head` is changed to `List a -> Result T a` for some type `T`?

Another issue is what the type `T` should be. Possible options:
* `T = String`, and `List.head` returns something like `Err "head of empty list"` in the failing case, `Array.get` returns something like `Err "invalid index"` or `Err "lookup failed at index 17"` on failure, etc.
* `T` is some dedicated union type which represents different failure causes. It could also be several different union types, e.g., one for errors possible on lists, one for errors possible on arrays, etc. Something like the various union types in [this GHC base module](https://hackage.haskell.org/package/base-4.8.1.0/docs/Control-Exception.html#t:ArrayException).
* `T = ()`, and the consumer becomes responsible for filling in specific "error messages" if desired. As in `Result.map2 f (...) (Result.formatError (always "head of empty list") (List.head l))`.
* `T` is chosen differently depending on what kind of additional information can meaningfully be provided in case of failure of specific functions. For example, for `List.head` the type `T` would indeed be `()` like in the previous point, but for `Array.get` the type `T` would be `Int` in order to easily give access to the index which led to the failure. One could then, analogously to above, write `Result.map2 f (...) (Result.formatError (always "something went wrong with getting") (Array.get some-expression a))`, but would also have the option of writing something like `Result.map2 f (...) (Result.formatError (\i -> "lookup at index " ++ toString i ++ " failed") (Array.get some-expression a))`.

rgrempel:
One consideration which I think is significant here is internalization. Not that the error type should do it, but that the error type should make it easy to do. This, I think, argues against making the error type a String (e.g. "head of empty list"), because that is awkward to match for translation purposes.

[...]

PR #558, opened 2016-04-13
change String.toInt and String.toFloat to return Maybe instead of Result

gyzerok:
### Whats done
Two little changes to the `String` module API.

```diff
-toInt : String -> Result String Int
+toInt : String -> Maybe Int

-toFloat : String -> Result String Float
+toFloat : String -> Maybe Float
```

### Why?

For more consistency of API with other core modules.

**String**
 * `uncons : String -> Maybe (Char, String)`

**List**
 * `head : List a -> Maybe a`
 * `tail : List a -> Maybe (List a)`

**Dict**
 * `get : comparable -> Dict comparable v -> Maybe v`

Personally I was confused when saw this first time. One may say that we can't get an actual error message with maybe, but there is only one possible error here. And you don't return error for example for while trying to get list head, it's just redundant.

### DIFF
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -205,14 +205,14 @@ function toInt(s)
 	var len = s.length;
 	if (len === 0)
 	{
-		return _elm_lang$core$Result$Err("could not convert string '" + s + "' to an Int" );
+		return _elm_lang$core$Maybe$Nothing;
 	}
 	var start = 0;
 	if (s[0] === '-')
 	{
 		if (len === 1)
 		{
-			return _elm_lang$core$Result$Err("could not convert string '" + s + "' to an Int" );
+			return _elm_lang$core$Maybe$Nothing;
 		}
 		start = 1;
 	}
@@ -221,10 +221,10 @@ function toInt(s)
 		var c = s[i];
 		if (c < '0' || '9' < c)
 		{
-			return _elm_lang$core$Result$Err("could not convert string '" + s + "' to an Int" );
+			return _elm_lang$core$Maybe$Nothing;
 		}
 	}
-	return _elm_lang$core$Result$Ok(parseInt(s, 10));
+	return _elm_lang$core$Maybe$Just(parseInt(s, 10));
 }
 
 function toFloat(s)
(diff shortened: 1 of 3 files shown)


eeue56:
You need to also change the Elm files, not just the native files.

gyzerok:
@eeue56 yeah, you are right. Stupid me..

Now fixed.

eeue56:
I also think adding some tests to https://github.com/elm-lang/core/blob/dev/tests/Test/String.elm might be worthwhile

gyzerok:
Done.

Pisys:
I'd keep the `Result` return type because transforming a String to some number is something which can fail for various errors, not just one. It's just that the current implementation does not serve more specific error messages than the one implemented. 

Additionally `String.toInt` and `String.toFloat` are different to the other functions you mentioned. The reason why they return `Maybe` is because there is something returned _maybe_. `Nothing` is returned in cases where there is _nothing_. That's not an error. 

But a String which can't be turned into a number isn't _nothing_, it's an _error_. This difference has to be reflected by the type.

Since the `Result` can be turned into a `Maybe` very easily (call `Result.toMaybe`), I see no need to change the API. IMHO.

gyzerok:
@Pisys can you provide list of potential errors which you see so far?

Pisys:
The point is that a failng String-to-number conversion is an error and not _nothing_ irrespective of how many or what kind of errors there could be.

jvoigtlaender:
@gyzerok, @Pisys, you both might want to take [this discussion](https://github.com/elm-lang/core/issues/369) into account. Note also that that linked issue being closed does not mean it has been resolved. In fact, it lives on as an item in https://github.com/elm-lang/core/issues/322. That item's name is:
> #369 - clearly define the "use case" for `Maybe` and `Result` in APIs

I don't think it makes much sense to have the discussion that you are having here, until that general item has been dealt with and been decided on (involving Evan).

As a consequence, I'm inclined to close this PR here and to ask you to instead add mention of `String.toInt` and `String.toFloat` to the discussion at #369, so that they are taken into account when this issue is revisited via #322.

jvoigtlaender:
So, can I close here?

evancz:
Yeah, this may be the right thing, but the decision will happen at a different time. It's an easy change, so it's not vital for it to be done via PR.

PR #562, opened 2016-04-18
Change type of LineStyle.dashing from "List Int" to "List Float"

Pisys:
Why?
* Makes relative definition (values between 0 and 1) of LineStyle properties possible
* LineStyle.width also is Float, so it'd be consistent

The background I'm making this PR on: I use to define shapes with relative coordinates and dimensions and finally scale the whole group of shapes to the actual size of the canvas. With LineStyle.dashing being of type `List Int`, this is not possible.

### DIFF
--- a/src/Graphics/Collage.elm
+++ b/src/Graphics/Collage.elm
@@ -92,7 +92,7 @@ type alias LineStyle =
     , width : Float
     , cap   : LineCap
     , join  : LineJoin
-    , dashing : List Int
+    , dashing : List Float
     , dashOffset : Int
     }
 


jvoigtlaender:
Why would you not also change the type of `LineStyle.dashOffset`, then? (ignoring for the moment that it is currently defunct, see https://github.com/elm-lang/core/pull/535)

Certainly if `dashing` is a `List Float`, then `dashOffset` should be a `Float`.

Pisys:
Yes, I forgot to mention that that change should account for dashOffset too, of course.

evancz:

PR #569, opened 2016-04-27
fixes error on documentation for filterMap

rebelwarrior:
This line `filterMap String.toInt ["3", "4.0", "5", "hats"] == [3,5]` would not work since toInt doesn't produce a Maybe but a Result.

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -197,9 +197,11 @@ filter pred xs =
 {-| Apply a function that may succeed to all values in the list, but only keep
 the successes.
 
-    String.toInt : String -> Maybe Int
+    stringToMaybeInt : String -> Maybe Int
+    stringToMaybeInt string = 
+      Result.toMaybe (String.toInt string)
 
-    filterMap String.toInt ["3", "4.0", "5", "hats"] == [3,5]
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Two comments:

1. Don't do the `List.` prefix for `filterMap` there. I think this is done nowhere in the documentation, that documentation for function `f` is prefixed by its module name *inside* that defining module.
2. A similar fix in the documentation for `Signal.filterMap` should probably be done in the same go.

rebelwarrior:
Oh shoot I thought I took the prefix out. That was an oversight. (Copy paste from command line)

I'll do another pull request tonight.

jvoigtlaender:
Cool. Actually, you can keep it simple, I think. Just remove the prefix and squash the commits. I just realized that making a corresponding change to the documentation of `Signal.filterMap` is not really necessary, with the changes to `core` that are on the horizon.

jvoigtlaender:
Thanks for opening the other PR. Closing this one here in favour of the other one.

rebelwarrior:
I could not find it to close it. Thought you'd closed it already. :)

PR #570, opened 2016-04-27
Wrap bitwise like the rest of native

eeue56:
Problem:

Having a bare `return` statement without a wrapping breaks a lot of build tools, because they try and load things as JS and get things wrong. These things are important to watch for build-tools, since that's how they know when packages have been upgraded and to re-run elm-make.

screen shot 2016-04-27 at 13 00 20

Solution:

Wrap the return statement with a function like the rest.

@evancz this is fairly important to be fixed before non-alpha release

### DIFF
--- a/src/Native/Bitwise.js
+++ b/src/Native/Bitwise.js
@@ -1,3 +1,4 @@
+var _elm_lang$core$Native_Bitwise = function() {
 
 return {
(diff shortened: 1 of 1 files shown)


evancz:
Thanks. I can tell it is important ;) If someone imports `Bitwise` their Elm code would just not work at all.

Issue #563, opened 2016-04-19
1 `rem` 0 is NaN


miikka:
Just noticed this (with elm repl 0.16.0 - not sure if that's the relevant version information):

```
> 1 `rem` 0
NaN : Int
```

Issue elm-lang/elm-compiler#262 seems to say that `NaN` shouldn't be an `Int`, so I assume this is a bug. [Relevant source](https://github.com/elm-lang/core/blob/3.0.0/src/Native/Basics.js#L18).

(I originally opened this issue as elm-lang/elm-compiler#1338 but moved it here as per @mgold's suggestion.)

jvoigtlaender:
Fixing this in https://github.com/elm-lang/core/pull/564.

jvoigtlaender:
I've replaced #576 by the more comprehensive #590.

I'm closing this here, since I think any discussion of this should take the larger picture of other division operators in Elm into account, see #590.

PR #576, opened 2016-04-28
Make `rem a 0` throw an error (just like `mod a 0` resp. `a % 0` does)

jvoigtlaender:
This fixes https://github.com/elm-lang/core/issues/563. It is in line with the treatment of `%`/`mod` as @evancz did here: https://github.com/elm-lang/elm-compiler/issues/262#issuecomment-24883407.

### DIFF
--- a/src/Native/Basics.js
+++ b/src/Native/Basics.js
@@ -8,6 +8,10 @@ function div(a, b)
 }
 function rem(a, b)
 {
+	if (b === 0)
+	{
+		throw new Error('Cannot perform rem 0. Division by zero error.');
+	}
 	return a % b;
 }
 function mod(a, b)


rtfeldman:
I think #262 is actually outdated; both `NaN` and `Infinity` are most definitely a thing in Elm at present:

http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Basics#isNaN
http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Basics#isInfinite

I believe the thinking on `NaN` and `Infinity` is "[they are part of the IEEE floating point number spec](http://steve.hollasch.net/cgindex/coding/ieeefloat.html), and adding checks to diverge from that spec has performance costs for operations that get used all over the place, so let's not make that the default."

Pretty sure it would make more sense to remove the check from `mod`, if anything. 😉

jvoigtlaender:
@rtfeldman, all the arguments you give apply to `Float`, and only to `Float`. Including the arguments about the existence of `isNaN` and `isInfinite`. Both of these have type `Float -> Bool`. But #262 what about `Int`, not `Float`. And both `mod` and `rem` are operations on `Int`. So I don't really see how your reasoning applies.

But in any case, I agree that most importantly `mod` and `rem` should behave consistently. Whether that is by both allowing division by zero or both not allowing it, well, ...

jvoigtlaender:
But if `rem a 0` stays allowed and `a % 0` becomes allowed, and then naturally `a // 0` as well, then `isNaN` and `isInfinite` should be adapted to have type `number -> Bool` rather than `Float -> Bool`, because they should become applicable to `Int` as well.

jvoigtlaender:
Closing this for the more comprehensive https://github.com/elm-lang/core/issues/590.

@rtfeldman, what do you think about that new issue?

PR #578, opened 2016-04-29
Add a function to coerce Cmd Never to Cmd msg

the-g-man:
This is more a suggestion than a pull request but it seemed easiest to illustrate the idea by writing the code.

There should be an idiom for expressing the idea of a sink effect - i.e. one that isn't expected to emit any value when it is called. A possible example might be writing to the storage API. The obvious way to do this is to use a `Cmd Never`. But as things stand there isn't an idiomatic way to incorporate this into an application loop.

Implementing the idiom is trivial but might be surprising if it appeared naked in application code. It seems like it might be better to wrap it up in a core function to create a common idiom for these cases.

### DIFF
--- a/src/Platform/Cmd.elm
+++ b/src/Platform/Cmd.elm
@@ -64,7 +64,7 @@ none =
 {-|-}
 sink : Cmd Never -> Cmd msg
 sink = 
-  map (Debug.crash "Cmd Never passed to Cmd.sink resulted in a value")
+  map (\_ -> Debug.crash "Cmd Never passed to Cmd.sink resulted in a value")
 
 
 {-|-}
--- a/src/Platform/Cmd.elm
+++ b/src/Platform/Cmd.elm
@@ -3,6 +3,7 @@ module Platform.Cmd exposing
   , map
   , batch
   , none
+  , sink
   , (!)
   )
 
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
The strict (instead of lazy) semantics of Elm will hit you in the face as soon as you try to use the sink-function in your program. Hint: the Debug.crash call there is not a good idea.

the-g-man:
Actually, as it is, it won't even type-check. It was written in haste at work while waiting for something to compile.

How about...

    map (\_ -> Debug.crash "Cmd Never passed to Cmd.sink resulted in a value")

?

jvoigtlaender:
> Actually, as it is, it won't even type-check.

Oh, of course it would have type-checked. The type of `Debug.crash` is `String -> a`, polymorphic in `a`, so it would have been instantiable to an arbitrary function type as return type well enough.

> How about...

Yes, that's better in that it won't immediately crash. Whether it does exactly what you want depends on the definition of `Cmd.map`, which I haven't looked into.

jvoigtlaender:
Yep, should work.

andersk:
What do you think about adding `absurd : Never -> a` (#593), and solving this with `map absurd`? It seems safer than making the user write a `Debug.crash` call.

evancz:
I'd rather not add something like this. There is no motivating example. No reason to add it then.

andersk:
I added my motivating example to #593.

PR #579, opened 2016-04-29
Update test suite to work with 0.17

eeue56:
## Problem:

- core tests didn't work with 0.17

## Solution:

- Use updated elm-test 
- Update syntax
- to run now, you just need to do `./set-up-tests.sh` and it will take care of everything
- Update travis (currently failing due to Json.Decode possible bug, see other notes)

## Other notes:

- No longer need to clone elm-console or anything else! No need for any external native packages
- Every test passed, except for decoding large ints in Json.Decode. Based on [this](pp.greenhouse.io/candidates?sort_by=last_activity&sort_order=desc&stage_status_id=2) commit, it might just be that the tests are outdated, and should actually be failing for ints like `1801439850948` anyway. Updating the tests to expect decoding to fail means the whole test suite passes.
- When 0.17 is released, I'll make a PR to use 0.17 final binary and update elm-test package to be elm-community/elm-test


evancz:
Awesome, thank you!

PR #581, opened 2016-04-30
update readme to list new default imports

eeue56:
## Problem:

- default imports are wrong on package site (and readme)

## Solution

- update default imports based on [this](https://github.com/elm-lang/elm-compiler/blob/master/src/Elm/Compiler/Imports.hs#L13) from the compiler


evancz:
I already updated it, but I wasn't able to get it to show up on the package website yet.

evancz:
Oops, did not see this earlier! I updated it eventually then.

PR #583, opened 2016-05-01
fix native platform error message grammar mistake

maxgurewitz:


maxgurewitz:
Don't think test failures have anything to do with this change.

PR #586, opened 2016-05-02
Accept large integer values in JSON decoders

spektroskop:
In 0.17, decoding integer values larger than 2147483647 from JSON fails:

`UnexpectedPayload "Expecting an Int at _.foo.bar but instead got: 2934520482"`

This change seems to fix the issue. It's basically reverting to the logic used in 0.16 with the return values changed to match the API used in 0.17.

### DIFF
--- a/src/Native/Json.js
+++ b/src/Native/Json.js
@@ -314,14 +314,19 @@ function runHelp(decoder, value)
 				: badPrimitive('a Bool', value);
 
 		case 'int':
-			var isNotInt =
-				typeof value !== 'number'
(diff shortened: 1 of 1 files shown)


liamcurry:
[Here's the commit where this was changed FYI](https://github.com/elm-lang/core/commit/4ecf9d850344bc687ebe0b2ab61e494a310583ac#diff-a1b3e8eaff04f6ea0707f2b1d1b11612R317)

evancz:
I guess I failed at [De Morgan's Law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws). Thanks for the fix!

PR #587, opened 2016-05-04
Assign neutral colors a hue of 0° when converting to HSL

klaftertief:
Converting a neutral color e.g. black (in RGB representation) to a HSL representation yields a hue value of `NaN` (because of calling `%` on `Infinity`.

``` Elm
import Color

hsl = Color.toHsl (Color.rgb 0 0 0)
-- == { hue = NaN, saturation = 0, lightness = 0, alpha = 1 }
```

This can be avoided by special casing a chroma of `0` and assigning it a hue of `0`.

### DIFF
--- a/src/Color.elm
+++ b/src/Color.elm
@@ -187,13 +187,16 @@ rgbToHsl red green blue =
     c = cMax - cMin
 
     hue =
-      degrees 60 *
-        if cMax == r then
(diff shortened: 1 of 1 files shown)


evancz:
I'm not managing the color stuff for now. I think folks can explore in the package ecosystem in the meantime.

PR #588, opened 2016-05-04
Remove dead code from Random

mgold:
Previously, seeds carried around two unused functions `split` and `range`. Every seed carried around the same `next` function. By removing these, we slightly decrease the memory footprint of a `Seed` and reduce the runtime size by, I haven't measured, but a few hundred bytes.

You could go further by placing the `State` directly in the `Seed` value constructor without the use of a record, but I'm trying to be minimally invasive.

### DIFF
--- a/src/Random.elm
+++ b/src/Random.elm
@@ -102,7 +102,7 @@ int a b =
           0 -> (acc, state)
           _ ->
             let
-              (x, state') = seed.next state
+              (x, state') = next state
             in
               f (n - 1) (x + acc * base) state'
 
@@ -362,9 +362,6 @@ you want to use a generator, you need to pair it with a seed.
 type Seed =
   Seed
     { state : State
-    , next  : State -> (Int, State)
-    , split : State -> (State, State)
-    , range : State -> (Int,Int)
     }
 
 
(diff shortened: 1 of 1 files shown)


rtfeldman:
Assuming this works, any reason not to do `type Seed = Seed State` and save a record?

mgold:
As mentioned above, none except being minimally invasive.

Evan has expressed interest in replacing the RNG with PCG so in that larger PR I can be more assertive.

rtfeldman:
Gotcha. 👍  Makes sense to me!

evancz:
Can you link to the implementation of the alternate approach that is "more random"?

mgold:
```elm
type Int64
  = Int64 Int Int

type Seed
  = Seed Int64 Int64
```
Then [this code](https://github.com/mgold/elm-random-pcg/blob/2.0.0/src/Random/Pcg.elm#L178-L338). Also, [these 64-bit arithmetic helpers](https://github.com/mgold/elm-random-pcg/blob/2.0.0/src/Random/Pcg.elm#L858-L945).

mgold:
That (about independent seeds) being said, it's possible to incorporate the better RNG *now* and add independent seeds *later*. Don't let that prevent you from adding the RNG to a patch release.

PR #589, opened 2016-05-09
Fix module embedding

klaftertief:
Renames embed functions to prevent infinite recursion when calling `embed()`


evancz:
Thanks! I did it a different way, but thanks for finding this and explaining what was wrong. The point of those names was so you get a nicer stack trace in JS, so it doesn't make sense to change them.

klaftertief:
Thanks for the quick fix. And I also prefer your changes.

PR #591, opened 2016-05-10
Use Math.trunc in Basics.truncate

Pisys:
Native Basics.truncate is implemented using a bitwise operator. The problem is that Javascript transforms the bit length from 64 to 32 bit for any number involved in bitwise operation. 
ES6 offers Math.trunc which offers the same functionality but keeps the bit length. We should use this instead.

### DIFF
--- a/src/Native/Basics.js
+++ b/src/Native/Basics.js
@@ -73,7 +73,7 @@ function isInfinite(n)
 
 function truncate(n)
 {
-	return Math.trunc(n);
+	return n < 0 ? Math.ceil(n) : Math.floor(n);
 }
 
 function degrees(d)
--- a/src/Native/Basics.js
+++ b/src/Native/Basics.js
@@ -73,7 +73,7 @@ function isInfinite(n)
 
 function truncate(n)
 {
-	return n | 0;
+	return Math.trunc(n);
 }
 
 function degrees(d)


justinmimbs:
Right on. But since `Math.trunc` isn't universally supported yet, may I suggest using `n < 0 ? Math.ceil(n) : Math.floor(n)` in the meantime?

Pisys:
Yes, thanks for your suggestion. I made another commit.

lukewestby:
Like @Pisys mentioned, the current `truncate` only works for numbers in (-858993459, 858993459). I benchmarked @Pisys's change against a couple other methods that don't exhibit this issue and this method seems to be pretty much on par with `x - x % 1` and about 10x faster than `parseInt()`.

http://jsbin.com/zazitofaco/edit?html,js,console

evancz:
I'm putting this in #721 because it will change how people's programs work.

As far as I know, JS does not have 64 bit integers. So when converting from float to int, why would we expect to get anything besides a 32 bit integer?

yjkogan:
Yeah my understanding is that all numbers in javascript are `double`s

lukewestby:
JS will faithfully represent integers up to 53 bits even if the engine uses doubles internally to store them past 32 bits. As a result, other core functions that call into built-in Math stuff will work on larger values whereas things that are done by hand like `truncate` will clamp to 32 bits. An SCCEE that demonstrates this situation:

```elm
truncate 900719925474097.4 == round 900719925474097.4 -- False
truncate 900719925474097.4 == floor 900719925474097.4 -- Also False
```

`round` and `floor`are both implemented directly as the functions in `Math` so they preserve the significand.

PR #596, opened 2016-05-12
Fix dates always being equal

### LINKED ISSUE #468: Value comparison does not work for Date.Date
Simplest reproducible example, doable in the REPL:

```elm
import Date
(Date.fromTime 1) == (Date.fromTime 2) -- => True
```

This is happening because instances of JS `Date` are "exotic" objects and not primitives. They have no own properties and end up falling through both the strict equality and object cases for `Utils.eq()`. They're a rather strange edge case as far as JS object comparison goes, as they need to be coerced to their `Number` representation first with `+` or `valueOf()` first.

The workaround for now that I've been using is to do the same sort of coercion, just in Elm:

```elm
import Date
date1 = Date.fromTime 1
date2 = Date.fromTime 2
(Date.toTime date1) == (Date.toTime date2) -- => False
```

But IMVHO it seems like if `Date` is going to be a built-in data type for the language, it should be possible to compare instances reliably without transformation.

The way I would plan to handle this is by adding a top-level case to `eq` to check `x instanceof Date && y instanceof Date` and then compare `x.valueOf() === y.valueOf()`. If that plan makes sense to you I can submit a PR :smile:

Thanks!

tesk9:
Fixes #468 

Adds tests for Date around equality.

### DIFF
--- a/src/Native/Utils.js
+++ b/src/Native/Utils.js
@@ -20,7 +20,9 @@ function eq(rootX, rootY)
 		if (x instanceof Date && y instanceof Date)
 		{
 			if (x.valueOf() !== y.valueOf())
-			{ return false; }
+			{
+				return false;
+			}
 		}
 		else if (typeof x === 'object')
 		{
--- a/src/Native/Utils.js
+++ b/src/Native/Utils.js
@@ -16,7 +16,13 @@ function eq(rootX, rootY)
 		{
 			continue;
 		}
-		if (typeof x === 'object')
+
+		if (x instanceof Date && y instanceof Date)
+		{
+			if (x.valueOf() !== y.valueOf())
+			{ return false; }
+		}
+		else if (typeof x === 'object')
 		{
 			var c = 0;
 			for (var key in x)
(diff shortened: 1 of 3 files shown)


src/Native/Utils.js
```diff
@@ -16,7 +16,13 @@ function eq(rootX, rootY)
 		{
 			continue;
 		}
-		if (typeof x === 'object')
+
+		if (x instanceof Date && y instanceof Date)
+		{
+			if (x.valueOf() !== y.valueOf())
+			{ return false; }
```
eeue56: Evan's preferred format is to stick the return on the next line too, then align the closing bracket with the first

eeue56: also for what it's worth, I think you should probably just do `return x.valueOf() === y.valueOf()` since there's no point doing anything else.

tesk9: I added a test for the tempting return x.valueOf === thing. It's in a while statement, so returning true could lead to false equality for complex structures.

eeue56:
:+1:

PR #602, opened 2016-05-14
replace err with badPrimitive

lukewestby:
`err` isn't defined, it looks like `badPrimitive` is used elsewhere for the same purpose so perhaps it's just a typo

thanks!

### DIFF
--- a/src/Native/Json.js
+++ b/src/Native/Json.js
@@ -406,7 +406,7 @@ function runHelp(decoder, value)
 		case 'key-value':
 			if (typeof value !== 'object' || value === null || value instanceof Array)
 			{
-				return err('an object', value);
+				return badPrimitive('an object', value);
 			}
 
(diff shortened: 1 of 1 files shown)


eeue56:
@lukewestby does this solve any known issues?

lukewestby:
Yep! See [this thread from elm-dev](https://groups.google.com/forum/m/#!topic/elm-dev/nTNS3QrdVSo)

cc @rluiten

rluiten:
I have a test case that produces the failure.
Here it is, once luke suggested the fix, i figured it was expecting an object but got a list.

[DecoderBug.elm.txt](https://github.com/elm-lang/core/files/264944/DecoderBug.elm.txt)

eeue56:
@lukewestby can you add the fail state to the test suite? thanks!

lukewestby:
I made a PR into @rluiten's fork with the test case, closing this one

PR #604, opened 2016-05-15
Improved Json.Decode message in case of failures

mariosangiorgio:
[JSon.Decode.fail](http://package.elm-lang.org/packages/elm-lang/core/4.0.0/Json-Decode#fail) let the user specify an error message, which is very useful to figure out what's going on.

With this pull request the error is not ignored anymore and it can be displayed.

Before the error message was

    FAILED. I ran into a `fail` decoder

With the new code it also reports the reason:

    FAILED. I ran into a `fail` ($REASON) decoder

### DIFF
--- a/src/Json/Decode.elm
+++ b/src/Json/Decode.elm
@@ -466,6 +466,11 @@ customDecoder =
 
         _ ->
             fail (tag ++ " is not a recognized tag for shapes")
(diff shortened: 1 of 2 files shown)


evancz:
This should be brought in now. Thanks!

PR #605, opened 2016-05-15
Add test case for #602 pull request fix.

rluiten:


eeue56:
@lukewestby @rluiten could one of you please update this PR with context from #602 ?

lukewestby:
@eeue56 sure!

The primary occurrence of this issue is described in [this thread from elm-dev](https://groups.google.com/forum/#!topic/elm-dev/nTNS3QrdVSo). One of the tests in @rluiten's `trie` package is attempting to decode key-value pairs on something that is not an object. Instead of resulting in `Err` as expect, it's throwing a runtime error because the `err` function used on https://github.com/elm-lang/core/blob/master/src/Native/Json.js#L409 does not exist and should be `badPrimitive` instead. As of Sun May 15 2016 10:58:49 GMT-0500 (CDT) this PR is still waiting on https://github.com/rluiten/core/pull/1 and will include both the fix and the test case covering this branch of `runHelp` in Native.Json once merged.

eeue56:
@evancz ping (this is fairly important)

tl;dr -

`err` should be `badPrimitive`, this renames that and adds a test case to make sure dict decoding is more tested

PR #614, opened 2016-05-19
Don't re-assign taggers when applying them

### LINKED ISSUE #609: Sub Msg is lost when a subscription triggers twice rapidly
I have written a module to keep track of the state of the keyboard arrow keys using the `Keyboard.ups` and `Keyboard.downs` subscriptions. The problem is when two `"keyup"` events occur rapidly, the second `Msg` does not make it to the update function.

The function mapping the subscription data (`keyCode`) to a `Msg` is run for both occurrences, but somewhere between this and the update function the second `"keyup"` event's `Msg` is lost.

I've tagged my project at a point where the issue is easy to reproduce with some debug logging. Holding down the left and right arrow key, then quickly releasing both will result in the one of the arrow keys getting stuck in the down state due to the lost `Msg`.

```elm
Main.elm:835 lstate: KeyDown
Main.elm:835 update: LeftKeyChange KeyDown
Main.elm:835 rstate: KeyDown
Main.elm:835 update: RightKeyChange KeyDown
-- repeated events from holding right key removed
Main.elm:835 rstate: KeyUp
Main.elm:835 lstate: KeyUp
Main.elm:835 update: RightKeyChange KeyUp
```

https://github.com/dancras/eggdude-elm/tree/debug-skipped-messages
https://github.com/dancras/eggdude-elm/blob/debug-skipped-messages/src/ArrowKeys.elm

I'm using OS X Chrome Version 50.0.2661.94 (64-bit)

dancras:
Fixes #609

### DIFF
--- a/src/Native/Platform.js
+++ b/src/Native/Platform.js
@@ -384,11 +384,14 @@ function toEffect(isCmd, home, taggers, value)
 {
 	function applyTaggers(x)
 	{
-		while (taggers)
+		var current = taggers;
+
+		while (current)
 		{
-			x = taggers.tagger(x);
-			taggers = taggers.rest;
+			x = current.tagger(x);
+			current = current.rest;
 		}
+
 		return x;
 	}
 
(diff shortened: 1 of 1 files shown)


evancz:
Ah, just found this and fixed it from an issue with websockets https://github.com/elm-lang/core/commit/cce1dc54f15ff9bae5d2900faf2661f2d1cd83b2

I'll have a patch for core out in the next few days! Want to figure out a couple other issues first.

Issue #618, opened 2016-05-21
Json.Decode:  Uncaught TypeError: Cannot read property 'tag' of undefined


ivanceras:
This will occur when decoding json with recursive data structure type.
 
```
Uncaught TypeError: Cannot read property 'tag' of undefined
```
pointing to this line of code

```js
function runHelp(decoder, value)
{
	switch (decoder.tag)
```

Here is a short [gist](https://gist.github.com/ivanceras/84e3f71f1f8f57b54e0a9c013904723b) for this issue

jvoigtlaender:
This is a known issue, with known current workarounds. Check issue 873 at the elm-compiler repo.

maca:
I cannot find a workaround while trying to receive some string using a port, in the most trivial example:

JS:
```  elmApp.ports.incomming.send("hello")```

Elm:
```

PR #619, opened 2016-05-21
A quick fix on https://github.com/elm-lang/core/issues/618

ivanceras:
Making sure `decoder` and `result` is not `undefined`

---
This change is [Reviewable](https://reviewable.io/reviews/elm-lang/core/619)

### DIFF
--- a/src/Native/Json.js
+++ b/src/Native/Json.js
@@ -310,200 +310,214 @@ function run(decoder, value)
 
 function runHelp(decoder, value)
 {
-	switch (decoder.tag)
-	{
-		case 'bool':
-			return (typeof value === 'boolean')
-				? ok(value)
-				: badPrimitive('a Bool', value);
+    if (decoder){
+        switch (decoder.tag)
+        {
+            case 'bool':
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
This is not how this issue will be fixed.

jvoigtlaender:
Well, let's leave that decision to Evan, actually. Maybe he wants to fix the Json decoder special case that way for the moment. But the general solution is of course to fix the compiler issue.

ivanceras:
I should have used the word `workaround`. This is just a workaround, which solves the issue while a proper fix has not yet landed. This lets me continue work on my projects.

jvoigtlaender:
Yes, I realize it's just a stop-gap measure for a specific case. Which is why I don't believe Evan is going to want to merge it into core. 

As I commented in the issue itself, there are existing workarounds, without changing core. Mainly using the lazy-function from package elm-lang/lazy.

evancz:
It looks like this just makes the crash happen later. It will still definitely happen though as far as I can tell. Better to actually fix things.

Issue #625, opened 2016-05-26
String.reverse doesn't handle unicode characters properly


eoftedal:
Examples from repl:

With combining unicode character (U+0300), the combining character becomes attached to the wrong character:
```
> String.reverse "màn"
"ǹam" : String 
```
With astral symbols (four byte UTF-8 character: U+1D306) it seems the surrogate pair is reversed:
```
> String.reverse "𝌆"
"��" : String
```

eoftedal:
```
> String.length "𝌆"
2 : Int
```

eoftedal:
Some inspiration: https://mathiasbynens.be/notes/javascript-unicode

eoftedal:
Pull request for reversing: https://github.com/elm-lang/core/pull/627

evancz:
Tracking in #735

eoftedal:
@evancz In #725 you should also include the count issue from my second comment in this thread

PR #626, opened 2016-05-26
Fix String.length for chars outside of BMP

eoftedal:


eoftedal:
Partial fix of: https://github.com/elm-lang/core/issues/625

Hurtak:
Array.from is not well supported by the browsers at the moment (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from)

Does elm have polyfills for that? 
If not maybe we could use `Array.prototype.slice.call(str).length`

ianmackenzie:
For what it's worth, I recently had elm-community/string-extra#6 merged in to add `toCodePoints` and `fromCodePoints`, so in a future release of `elm-community/string-extra` you should be able to use those (although it will of course be much slower than a native solution).

evancz:
This implementation is not viable because of browser support. The fastest "correct" implementation is probably something like this:

```javascript
function length(str)
{
    var len = str.length;
    var realLength = 0;
    var i = 0;
    while (i < len)
    {
        var word = str.charCodeAt(i);
        i += (0xD800 <= word && word <= 0xDBFF) ? 2 : 1;
        realLength++;
    }
}
```

This is O(n) in the length of the string though.

In normal circumstances, a language could keep two numbers with each string. One for the length in bytes (which JS does) and one for the length in characters (which JS does not do). When you append, the language runtime can just add *both* numbers. That way it is O(1) to look up either with an overhead of 32-bits extra, which is not too bad.

To add this info in JS means wrapping all strings in objects with the extra info, which I think would be crazy. So we either get O(n) `length` or we get O(1) `length` with slower everything else. I think both of these cases are bad enough that the current behavior is "better" overall given the information I have now. 

If people have specific scenarios that are causing trouble, I would be interested to hear about that. That said, I expect it'll only be viable to make `length` work properly when we have our own string representation someday.

evancz:
With some recent commits, `foldl` and `foldr` both support characters outside of the BMP, so it will be possible to get the real length with an expression like this.

```elm
String.foldl (\_ n -> n + 1) 0 "😃😃😃"
```

This has the O(n) performance, but it will give 3. This should be out with 0.19.

jvoigtlaender:
But shouldn't it at least be mentioned in #725 that this String.length correctness issue exists?

PR #627, opened 2016-05-26
String reversing supporting astral characters and combining unicode chars

eoftedal:
Code borrowed from @mathiasbynens:
https://github.com/mathiasbynens/esrever/blob/master/esrever.js


eoftedal:
Partial fix of: https://github.com/elm-lang/core/issues/625

lukewestby:
@eoftedal what are the performance effects of this change?

et:
Awesome @eoftedal! Shouldn't there also be a test [here](https://github.com/elm-lang/core/blob/master/tests/Test/String.elm#L18)?

Pilatch:
If there are performance implications, which I'd bet there are, maybe this could instead become another function.

eoftedal:
To be honest, I don't know the performance implications. I also don't know if people use a string reversing function enough that that matters. I can't really think of many valid use cases for it.

eoftedal:
Array.from(str).length seems to be about 10 times slower than str.length

eoftedal:
[...str].length seems to be about twice as fast as Array.from(str).length

src/Native/String.js
```diff
@@ -29,7 +29,7 @@ function concat(strs)
 }
 function length(str)
 {
-	return str.length;
+	return Array.from(str).length;
```
paplorinc: :+1:, just found this bug for e.g. "𝕔":
```elm
String.length "𝕔" == 2
```

Will this be included in `0.19`?

lydell: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Browser_compatibility

screenshot from 2017-01-21 15-32-06

paplorinc: Indeed, how about
```javascipt
return str.split().length;
``` then?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#Browser_compatibility

File: src/Native/String.js [hunk 2]
```diff
@@ -44,10 +44,19 @@ function filter(pred, str)
...
 	return str.split('').map(_elm_lang$core$Native_Utils.chr).filter(pred).join('');
 }
-function reverse(str)
-{
-	return str.split('').reverse().join('');
+
+const regexSymbolWithCombiningMarks = /([\0-\u02FF\u0370-\u1AAF\u1B00-\u1DBF\u1E00-\u20CF\u2100-\uD7FF\uE000-\uFE1F\uFE30-\uFFFF]|[\uD800-\uDBFF][\uDC00-\uDFFF]|[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?:[^\uD800-\uDBFF]|^)[\uDC00-\uDFFF])([\u0300-\u036F\u1AB0-\u1AFF\u1DC0-\u1DFF\u20D0-\u20FF\uFE20-\uFE2F]+)/g;
+const regexSurrogatePair = /([\uD800-\uDBFF])([\uDC00-\uDFFF])/g;
+
+function reverse(string) {
```
paplorinc: could you please `squash` the history, it's a mess :/

src/Native/String.js
paplorinc: Apparently the same bug in available in Java also.
However, instead of simply making the length linear there should be some lazy, memoized solution (amortized constant time), falling back to `str.length`

evancz:
Fixed in https://github.com/elm-lang/core/commit/8ca9c383292cb9cead5d07c9ad5613e2e9ccb4b2 with proof that speed is equal or faster [here](https://jsperf.com/utf16-string-reverse/1) and tests in https://github.com/elm-lang/core/commit/d796fcae2a9dba3dc7c74328a77d473525be3590.

Issue #635, opened 2016-06-02
Problem with small values in Random.int on init


PezzA:
Raised this question on stack overflow and it was suggested to create a bug report here.

http://stackoverflow.com/questions/37522369/elm-random-number-on-init-strange-behaviour

I've been working through the examples here:

http://guide.elm-lang.org/architecture/effects/random.html

And what I'm trying to do with the dice is to get to generate a D6 when then component is created.

```

module Components.DiceRoller exposing (Model, Msg, init, update, view)

import Html exposing (..)
import Html.App as Html
import Html.Attributes exposing (..)
import Html.Events exposing (..)
import Random
import String exposing (..)

main =
    Html.program
        { init = init
        , view = view
        , update = update
        , subscriptions = subscriptions
        }



-- MODEL


type alias Model =
    { dieFace : Int
    }


init : ( Model, Cmd Msg )
init =
    ( Model 0, (Random.generate NewFace (Random.int 1 6)) )



-- UPDATE


type Msg
    = NewFace Int


update : Msg -> Model -> ( Model, Cmd Msg )
update msg model =
    case msg of
        NewFace newFace ->
            ( Model newFace, Cmd.none )



-- SUBSCRIPTIONS


subscriptions : Model -> Sub Msg
subscriptions model =
    Sub.none



-- VIEW


dieFaceImage : Int -> String
dieFaceImage dieFace =
    concat [ "/src/img/40px-Dice-", (toString dieFace), ".svg.png" ]


view : Model -> Html Msg
view model =
    let
        imagePath =
            dieFaceImage model.dieFace
    in
        div []
            [ img [ src imagePath ] []
            , span [] [ text imagePath ]
            ]
```

However if I use 1 6 as the random min max, I only get 1 different value per minute.  This only happens on init, it's fine in the orig example where it a UI action.

mariosangiorgio:
As I mentioned on the StackOverflow answer, this behaviour is a bit odd but I am not sure if that's expected or not. What is happening is that, for this combination of parameters, there are quite big intervals of seeds that generate the same first value of the pseudo-random sequence.

It's worth noting that, if we exclude the first value, the sequence of the numbers produced look random enough.

[...]

PR #636, opened 2016-06-02
Amended constant value used by the pseudo-random number generator

mariosangiorgio:
As I mentioned in my comment in issue #635, I think the constant we're using needs to be amended. The [Haskell implementation](http://hackage.haskell.org/package/random-1.0.1.1/docs/src/System-Random.html#RandomGen) uses that value:
```
mkStdGen32 sMaybeNegative = StdGen (s1+1) (s2+1)
	[...]
	s2      = q `mod` 2147483398

stdNext (StdGen s1 s2) = (fromIntegral z', StdGen s1'' s2'')
		[...]
		s2'' = if s2' < 0 then s2' + 2147483399 else s2'

stdSplit std@(StdGen s1 s2)
			[...]
                        new_s2 | s2 == 1          = 2147483398
			[...]
```
The same value is mentioned in the [original paper](http://webhome.csc.uvic.ca/~wkui/Courses/CSC446/CLCG.pdf).

> For 32-bit computers, we suggest l= 2, m1 = 2147483563, a1 = 40014, m2 = 2147483399 and a2 =
> 40692.

I am by no means expert on pseudo-random number generators, so I have no idea why we might have been using a different values. I believe that's a typo but please double check it before accepting this pull request.

### DIFF
--- a/src/Random.elm
+++ b/src/Random.elm
@@ -432,7 +432,7 @@ magicNum3 = 52774
 magicNum4 = 40692
 magicNum5 = 3791
 magicNum6 = 2147483563
-magicNum7 = 2137383399
+magicNum7 = 2147483399
 magicNum8 = 2147483562
 
 


evancz:
Thanks! There is an alternate implementation of PRNG in the pipeline, so I think it makes sense to prefer that route.

evancz:
Actually, it does make sense to merge this in. Great find, thank you for fixing it in a PR!

PR #639, opened 2016-06-03
Fixes #637; Preserve Err values when using Json.Decode.customDecoder

### LINKED ISSUE #637: Json.Decode.customDecoder swallows Err strings
The function `Json.Decode.customDecoder` has the following type:

`Decoder a -> (a -> Result String b) -> Decoder b`

My expectation as a user is that the decoder produced by `customDecoder` decodes an initial payload using the first argument, then maps the result into the second argument. However, when the second argument is applied and its result resolves to an `Err string`, the `string` is discarded. No matter what, when a custom decoder fails, the string value of Err is "Expecting something custom but instead got: <data>".

You can reproduce this by entering the following into elm repl:
```
import Json.Decode exposing (customDecoder, decodeString, string, (:=))

jsonString = """{ "foo": "bar" }"""

myDecoder = customDecoder ("foo" := string) (\_ -> Err "I want to see this message!")

decodeString myDecoder jsonString
-- Err "Expecting something custom but instead got: {\"foo\":\"bar\"}"
```

noahzgordon:
See issue: https://github.com/elm-lang/core/issues/637

This commit allows Err values returned by the second argument of `Json.Decode.customDecoder` to be preserved when the decoder fails. This way all user-defined error messages are not swallowed by the decoder.

### DIFF
--- a/src/Native/Json.js
+++ b/src/Native/Json.js
@@ -236,6 +236,11 @@ function bad(msg)
 	return { tag: 'fail', msg: msg };
 }
 
+function badWithCustomMessage(msg)
+{
+	return { tag: 'custom', msg: msg };
+}
+
 function badToString(problem)
 {
 	var context = '_';
@@ -272,6 +277,9 @@ function badToString(problem)
 				return 'I ran into a `fail` decoder'
 					+ (context === '_' ? '' : ' at ' + context)
 					+ ': ' + problem.msg;
+
+			case 'custom':
+				return problem.msg;
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


jvoigtlaender:
This will not give a good error message. Try it out.

jvoigtlaender:
Sorry, scrap that.

evancz:
I just ran into this independently and did a fix in 7f6e518378ec354681ef6337d5dfa8182b2e8fca. Someone pointed me here, which is very similar. My fix includes context information, so if you run into a `customDecode` nine levels into a JSON object, the error message will tell you about that context.

That said, it's probably still a good idea to test that the error message produced contains the `Err str` string as a subset. Are you willing to adapt your tests to the alternate implementation?

noahzgordon:
@evancz Gladly! I will open a new PR shortly and post it here once it's done.

PR #640, opened 2016-06-07
Resolves #634; Raise and embed informative error message when rootDom…

### LINKED ISSUE #634: Better error when div for embedding is undefined
Right now it complains that it can't call `appendChild` on `undefined`. Which makes sense, but if the intent is to treat Elm as a black box in production, then perhaps a check on the element existing and giving a clearer error would be helpful.

noahzgordon:
Raises an informative error message when a falsy value is passed into `embed()` and appends it to the document body.

See: https://github.com/elm-lang/core/issues/634

### DIFF
--- a/src/Native/Platform.js
+++ b/src/Native/Platform.js
@@ -16,7 +16,14 @@ function addPublicModule(object, name, main)
 
 	object['embed'] = function embed(domNode, flags)
 	{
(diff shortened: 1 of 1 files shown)


evancz:
Embedding is no longer handled by core. It is in virtual-dom. Not sure if it's still an issue.

PR #646, opened 2016-06-13
Avoiding infinite loop when the search string is empty

lorenzo:
The `String.indexes` function enters an infinite loop when doing  `String.indexes "" "Some string"`, this is because the `i` variable is no advanced on each step of the loop since it will always return 0.

I chose to return an empty list for this cases, as I think it is the most sensible thing to do IMO, but another valid interpretation is that the empty string is contained in all of the indexes of the string.

In case the second interpretation is preferred, I can gladly change it :)

### DIFF
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -192,7 +192,7 @@ function indexes(sub, str)
 	var subLen = sub.length;
 	var i = 0;
 	var is = [];
-	while ((i = str.indexOf(sub, i)) > -1)
+	while (subLen > 0 && (i = str.indexOf(sub, i)) > -1)
 	{
 		is.push(i);
 		i = i + subLen;
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -193,14 +193,16 @@ function indexes(sub, str)
 	var i = 0;
 	var is = [];
 	
-	if (subLen > 0)
+	if (subLen < 1)
 	{
-		while ((i = str.indexOf(sub, i)) > -1)
-		{
-			is.push(i);
-			i = i + subLen;
-		}	
+		return _elm_lang$core$Native_List.Nil;
 	}
+
+	while ((i = str.indexOf(sub, i)) > -1)
+	{
+		is.push(i);
+		i = i + subLen;
+	}	
 	
 	return _elm_lang$core$Native_List.fromArray(is);
 }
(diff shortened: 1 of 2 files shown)


eeue56:
You should add a test to this PR.

lorenzo:
@eeue56 Done

evancz:
Thanks, this change makes sense to me! I'm trying to sort out the contributors agreement stuff at the moment. Can you either ping the email you sent previously or fill [this](https://github.com/elm-lang/elm-compiler/blob/master/ContributorAgreement.pdf) out and send it to me? It's looking like we'll be able to do something nicer soon, so I apologize for the inconvenience.

lorenzo:
@evancz Sent to your email address

src/Native/String.js
```diff
@@ -192,7 +192,7 @@ function indexes(sub, str)
 	var subLen = sub.length;
 	var i = 0;
 	var is = [];
-	while ((i = str.indexOf(sub, i)) > -1)
+	while (subLen > 0 && (i = str.indexOf(sub, i)) > -1)
```
evancz: Wouldn't it be faster to have the check for `subLen > 0` outside the `while` loop? Seems like it only needs to happen once in an `if`, not on every loop.

lorenzo: Sure, I can do that :)

lorenzo:
@evancz Implemented your suggestion

src/Native/String.js
```diff
@@ -192,11 +192,16 @@ function indexes(sub, str)
...
 	{
-		is.push(i);
-		i = i + subLen;
+		while ((i = str.indexOf(sub, i)) > -1)
+		{
+			is.push(i);
+			i = i + subLen;
+		}	
 	}
+
```
evancz: Better, but it can be optimized a bit more :) If you move the `if` above allocating `i` and `is` you can just write something like this:

```javascript
if (subLen > 0)
{
    return _elm_lang$core$Native_List.Nil;
}
```

Which is cool in two ways. First, it means `i` and `is` do not even get allocated. Second, and more importantly, it uses the [early return](http://blog.timoxley.com/post/47041269194/avoid-else-return-early) style, which I try to use in any C-like language I use. As a reader, it means I can be sure, nothing past this point is relevant to this code / some invariant is definitely true past this point.

eeue56: (typo: should be `subLen < 1`)

lorenzo:
@evancz done

evancz:
Great, thank you!

PR #648, opened 2016-06-17
Fix Dict.merge problem where some elements ignored from second dict

fredcy:
I added tests that demonstrate the problem. In particular, the "merge singletons in order" and "partially overlapping" tests will fail without the changes to Dict.merge.


fredcy:
@knewter: This addresses that issue you raised.

knewter:
(for reference, the issue @fredcy is mentioning is here: https://groups.google.com/forum/#!topic/elm-dev/gbYNWTT0TsM)

evancz:
Cool, looks good to me @fredcy. Thanks!

PR #650, opened 2016-07-01
Compare always with anonymous functions

willnwhite:
As writing an anonymous function like this is how you'd do it without `always`.

### DIFF
--- a/src/Basics.elm
+++ b/src/Basics.elm
@@ -587,6 +587,10 @@ handy more often than you might expect. For example, creating a zeroed out list
 of length ten would be:
 
     map (always 0) [0..9]
+
+rather than:
+
+    map (\_ -> 0) [0..9]
 -}
 always : a -> b -> a
 always a _ =


willnwhite:
Have "all checks failed" because I chose to commit to the master branch?

jvoigtlaender:
@willnwhite, no, that's not the problem.

willnwhite:
Am I okay to leave it as it is?

jvoigtlaender:
Yes, you have done the pull request exactly as it's to be done. (And I'll delete our brief exchange here, so as to keep the issue clean and focused on your actual proposed change.)

willnwhite:
Thanks.

evancz:
It's got something like this in the latest docs: https://github.com/elm-lang/core/blob/master/src/Basics.elm#L575-L585

Sorry for the delay. Suggestion makes sense though!

willnwhite:
I have a theory this was left late because its meaning isn't clear without the file, which is in another tab. Can you attest to that? If so, there's a design issue for GitHub.

Issue #652, opened 2016-07-01
`(==)` does not work correctly on records with field named `ctor`


jvoigtlaender:
The current version of `core` (the current `master` branch) evaluates `{ctor = [0]} == {ctor = [0]}` to `False`.

@evancz, probably caused by the recent refactoring in https://github.com/elm-lang/core/commit/6f5fbb8bc1ba2ce0fd3a3acf41e347c447689022?

jvoigtlaender:
Update: This now also happens with the released version `4.0.2` of `core`.

That is, the following program wrongly outputs `False`:
```elm
import Html

main = Html.text (toString ({ctor = [0]} == {ctor = [0]}))
```
while the following program continues to correctly output `True`:
```elm
import Html

main = Html.text (toString ({field = [0]} == {field = [0]}))
```

@evancz, just to let you know that the bug-fix release of `core` also introduced a bug. This one, which wasn't present in version `4.0.1`.

evancz:
The real fix will be to reintroduce the `_` field to Elm records. This will serve as a runtime tag of whether something is an Elm record a union type *or* a JS value.

I feel like you found this by reading the new code, not by having a field with that name, so I'm not certain how urgent this is. That said, I suspect there is a stopgap fix that does not slow things down too much. Perhaps even just doing the non-pseudo-code of `ctor === ctor || eqHelp(ctor, ctor)`

jvoigtlaender:
No, I didn't find this by reading the code. I found it by trying to run the test suite. Note that it contains this (at https://github.com/elm-lang/core/blob/master/tests/Test/Equality.elm#L22):
```elm
        , test "ctor same" <| assert ({ctor = Just 3} == {ctor = Just 3})
```
From there, I was trying whether it has something to do especially with `Just` of `Maybe`, which was not the case...

Only after that, I checked the source code to see what other special handling goes on with the `ctor` field name, and encountered this other failure, with `toString`: https://github.com/elm-lang/core/issues/654.

As an aside, it's a pity that the test suite is currently failing on GitHub.

evancz:
I'm working on the tests. You can try sleuthing too if you'd like.

jvoigtlaender:
I see you got the test suite passing already. A nice green batch on the repo page. Cool!

[...]

PR #655, opened 2016-07-02
Add tests for equality, concerning records

jvoigtlaender:
The purpose of these is to expose, to testing, any special case treatment surrounding the record field name `ctor`. See https://github.com/elm-lang/core/issues/652, which observed a bug in this.


evancz:
Cool, thanks!

PR #659, opened 2016-07-12
Fix List.take causing stack overflow

### LINKED ISSUE #601: List.take throws "Maximum call stack size exceeded" with large numbers
This happens if you try to take with an `n` greater than about 5707 and the list you're taking from has at least that many items. This did not happen in v0.16.

In elm-repl:
```
> List.take 5707 [0..10000]
RangeError: Maximum call stack size exceeded
```

nphollon:
Fixes #601 

`List.take` causes a runtime error if we take approx. 4000-5000 items.

    import Html
    main = [ 1 .. 6000 ] |> List.take 4000 |> toString |> Html.text

I think this is because the recursive definition of `List.take` does not get optimized by the compiler. I have rewritten it to use `foldr` instead of recursion.

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -459,12 +459,16 @@ take n list =
     []
 
   else
-    case list of
-      [] ->
-        list
+    let
+      takeCons x (i, xs) =
+        if n < i then
+          (i - 1, xs)
 
-      x :: xs ->
-        x :: take (n - 1) xs
+        else  
+          (i - 1, x :: xs)        
+    in
+      foldr takeCons (length list, []) list
+        |> snd
 
 
 {-| Drop the first *n* members of a list.


evancz:
I believe this version costs O(n) in all cases. So if I say `List.take 2 [1..1000]` it will traverse 1000 elements. I believe there is a way to write this tail recursively that is O(k) where k is the amount you want to take. That said, this will necessarily reverse the list, meaning you will have to do a reverse on it, increasing allocation.

It is not clear to me that this is a win. This problem exists in OCaml and SML as well. There is no real trick to fix it besides letting the stack be bigger or using a more appropriate data structure for your data set.

nphollon:
So I am building a [collision detection library](http://package.elm-lang.org/packages/nphollon/collision/1.0.2/) that uses a bunch of recursive algorithms. When I started encountering stack overflows, I wasn't surprised; I figured I'd probably done something dumb. I spent a couple days tracking down the problem (as chronicled on [elm-discuss](http://package.elm-lang.org/packages/nphollon/collision/1.0.2/)). Finding out that the issue was in the standard library was pretty annoying.

I think its a big win to get rid of the stack overflow. When somebody is debugging a runtime error, elm-lang/core is going to be very far down on the list of suspects. Even if this isn't encountered that often, its going to be a time sink.

Any of these I think would be better than the status quo:

- replacing `List.take` with a tail-recursive version that is O(k) but reverses the order of the items.
- removing `List.take`
- adding a note to the documentation that `List.take` only works for small values of `n`.

evancz:
I think Yaron outlines a very attractive way to approach this kind of problem [here](https://blogs.janestreet.com/optimizing-list-map/). If we can show the numbers work nicely in Elm, I think it makes sense to go in this direction across the board.

What do you think of that approach? Would you be willing to do some profiling to see if you can make the performance / memory tradeoffs work out favorably?

nphollon:
I think that approach sounds really nice. Gives a fast result for the most common use cases, and switches to an "at least it works" implementation for large lists. I will do some profiling and get back to you.

evancz:
Cool, I am going to close this for now. New issue or PR when you find out more! Fine to continue discussion here if necessary though.

PR #661, opened 2016-07-14
withCmds is more understandable than !

willnwhite:
damn haha, just realized what `!` from `Platform.Cmd` does, amazing :heart_eyes:

Maybe `!` has a history, but I think it could do with a more descriptive name.
I've avoided using it so far because it's harder to understand than the alternative.
I guess shorthand is useful, but I always feel like I never know who's going to need to understand my code quickly.

maybe `withCmds` as an alias for `(!)`. you could do
```{ ...model... } `withCmds` [cmdA, cmdB]```
. but then again that's something people might have to look up once and remember then too

Though that might be an even better "basic" way because it gets rid of the tuple syntax too.

I was wondering if `!` was used for anything else because I think `withCmds` like this is the most readable.

and since this is a thing... http://package.elm-lang.org/help/design-guidelines#avoid-infix-operators

### DIFF
--- a/src/Platform/Cmd.elm
+++ b/src/Platform/Cmd.elm
@@ -65,3 +65,8 @@ none =
 (!) model commands =
   (model, batch commands)
 
+
+{-| Alias for `!`. `{ ...model... } `withCmds` [cmdA, cmdB]` -}
+withCmds : model -> List (Cmd msg) -> (model, Cmd msg)
+withCmds : model commands =
+  model ! batch commands

### CODE REVIEW

**Comment by rtfeldman** [comment]
I think the point of `!` is to reduce verbosity in an super common operation. You can always do this:

```elm
( model, Cmd.none )
```

...but you end up writing that extremely often as your code base grows, so `!` gives you a more concise way to express it:

```elm
model ! []
```

I wouldn't use a `withCmds`, since it'd be more verbose than either of these! 😄

If I didn't want to use `!`, I'd just fall back on the tuple style. Either works fine!

willnwhite:
`( model, Cmd.none )` is clearer than `model ! []`, so I use that. But 
```elm
model `withCmds` [cmdA, cmdB]
```
 is less verbose (and arguably clearer) than 
```elm
( model, Cmd.batch [cmdA, cmdB] )
```

evancz:
Don't want to have PRs open for API change suggestions. I think this is viable, but I also would like to let `(!)` live for longer and see what happens.

willnwhite:
I was suggesting it as an alias for `(!)`, rather than as a replacement, but sure. I guess I should start using it in my own code first!

willnwhite:
`model ! [] == (model, Cmd.none)` is now a loading screen message on Slack.

PR #662, opened 2016-07-14
Removed dead code for equality() in Native/Json

jvoigtlaender:
I noticed that this appears to be dead code. If it is indeed some leftover not needed anymore, it could be removed. If instead it is needed in some way I did not spot, sorry for the disturbance.

### DIFF
--- a/src/Native/Json.js
+++ b/src/Native/Json.js
@@ -507,80 +507,6 @@ function runHelp(decoder, value)
 }
 
 
-// EQUALITY
-
-function equality(a, b)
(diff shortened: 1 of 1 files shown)


rgrempel:
I believe that the code may be used here:

https://github.com/elm-lang/virtual-dom/blob/master/src/Native/VirtualDom.js#L252-L262

jvoigtlaender:
Ah, good to know. I was only thinking about `core` internally. My bad.

PR #666, opened 2016-07-15
Add test for customDecoder messages and fix typo in messages

noahzgordon:
#639 was closed because @evancz came across the same issue and fixed it in the same way, adding some useful context information. This PR extends the test from #639 to fit his implementation. It also changes the error message to use the function name `customDecoder`, which was misspelled as @jvoigtlaender points out in https://github.com/elm-lang/core/commit/7f6e518378ec354681ef6337d5dfa8182b2e8fca

### DIFF
--- a/src/Native/Json.js
+++ b/src/Native/Json.js
@@ -274,7 +274,7 @@ function badToString(problem)
 					+ ':\n\n' + problems.join('\n');
 
 			case 'custom':
-				return 'A `customDecode` failed'
+				return 'A `customDecoder` failed'
 					+ (context === '_' ? '' : ' at ' + context)
 					+ ' with the message: ' + problem.msg;
 
--- a/tests/Test/Json.elm
+++ b/tests/Test/Json.elm
@@ -4,7 +4,6 @@ import Basics exposing (..)
 import Result exposing (..)
 import Json.Decode exposing ((:=))
 import String
-import Debug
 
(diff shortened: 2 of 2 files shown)


evancz:
Looks good. Can you make the test use `String.contains` though? I don't think it makes sense to encode the particular wording of the error message in the test, but it does make sense to make sure the *users* messages appears somewhere in the error message.

noahzgordon:
Sounds better. On it.

noahzgordon:
Updated with your suggestion. 

It gets a bit funky when testing anything but equality for Err values, since to operate on the Err string you have to pattern match and then handle the possibility of Ok values. I've handled this in my projects my using `Debug.crash` to raise with an informative message on the Ok branch, so that's what I've done here. Let me know if you've found a better solution.

evancz:
Probably better to use [pass](http://package.elm-lang.org/packages/elm-community/elm-test/1.1.0/ElmTest#pass) and [fail](http://package.elm-lang.org/packages/elm-community/elm-test/1.1.0/ElmTest#fail) to create the assertion.

noahzgordon:
Cool, didn't know about those functions. Updated to use `fail`.

evancz:
Thanks!

noahzgordon:
First commits to core! 🎉🎉🎉

PR #670, opened 2016-07-19
Add test for dict equality

jvoigtlaender:
To ensure that no regression will sneak in against the semantic improvement from `4.0.3`.


evancz:
Thanks!

PR #671, opened 2016-07-19
Improve the Array get exemple

tibastral:
The exemple wasn't really crystal clear, because it could be understood as : find the element 1 from the list [0, 1] and returns it. As soon as you understand it's an index (by the exemple) everything goes easier.

### DIFF
--- a/src/Array.elm
+++ b/src/Array.elm
@@ -165,10 +165,10 @@ push =
 
 {-| Return Just the element at the index or Nothing if the index is out of range.
 
-    get  0 (fromList [0,1,2]) == Just 0
-    get  2 (fromList [0,1,2]) == Just 2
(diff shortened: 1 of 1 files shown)


evancz:
Makes sense to me, thanks!

tibastral:
Cool ! :) Will be in SF at the end of october. Would love to meet you !

—
Thibaut

On 22 septembre 2016 at 22:23:28, Evan Czaplicki (notifications@github.com) wrote:

Merged #671.

evancz:
Check if there is a meetup on http://www.meetup.com/Elm-user-group-SF/ that overlaps your trip. If not, we can sometimes shuffle them around for stuff, so let @rtfeldman and me know on twitter and we'll see.

PR #672, opened 2016-07-21
Example for flip

willnwhite:


jvoigtlaender:
Criticism of the example:

The name `divideBy` ist not appropriate for a function defined as:
```elm
f a b = a / b
```
It would be appropriate for a function defined as:
```elm
f a b = b / a
```

willnwhite:
Thanks Janis, excellent.

evancz:
I think this would benefit from an example, but I'd prefer to use one that showed up in normal usage. I think the division example does not clarify things very well. Let's start with uses and work from there to an example.

PR #678, opened 2016-08-01
Add documentation for (!) in Platform.Cmd

cobalamin:
What `(!)` is or does is _such_ a common source of confusion and thus a very common question in the Elm Slack, especially in #beginners. It's very commonly used in code in the wild, especially the `model ! []` form.

There's no way for newcomers to figure this out on their own, as there is no comprehensive explanation in the guide on this – and worse, no documentation on the function in the package. So I wrote some. I hope it works 😄

### DIFF
--- a/src/Platform/Cmd.elm
+++ b/src/Platform/Cmd.elm
@@ -60,7 +60,38 @@ none =
   batch []
 
 
-{-|-}
+{-| This function is a shorthand for returning a `(model, Cmd msg)` value. It's a combination of `(,)` and `Cmd.batch`. See how this can make the following `update` function more readable, both for `Cmd.batch`ing commands, and for the common case of returning `(model, Cmd.none)`:
+
+```
+update msg model =
+    case msg of
+        DoThings ->
+            ( model, Cmd.batch [ someCmd, someOtherCmd ] )
+
+        NewValues a b ->
+            ( { model
+                | a = a
+                , b = b
+              }
+            , Cmd.none
+            )
+```
+
+becomes
+
+```
+update msg model =
+    case msg of
+        DoThings ->
+            model ! [ someCmd, someOtherCmd ]
+
+        NewValues a b ->
+            { model
+                | a = a
+                , b = b
+            } ! []
+```
+-}
 (!) : model -> List (Cmd msg) -> (model, Cmd msg)
 (!) model commands =
   (model, batch commands)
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
I'm all for adding documentation.

One comment, though: This specific construct *is* explained in the [FAQ for Learners](http://elm-community.github.io/elm-faq/) linked to from http://elm-lang.org/docs. Here: http://faq.elm-community.org/17.html#what-is-this---construct-in-the-code.

What is wrong with the FAQ that people "in the Elm Slack, especially in #beginners" do not find it, or do not think of looking there?

jvoigtlaender:
Oh, and the way you included code samples in the docs is not the way to do it. Please look at examples in other source files, e.g., `src/Maybe.elm`.

cobalamin:
Fair point!

Nothing specific wrong with the FAQ, but the only way I stumbled on that myself was when I had already worked through the guide and some other materials, and went on to check everything at `elm-lang.org/docs` out of curiosity. It's the last bullet point and it's not bolded, so it's fairly easy to miss, even there.

Coming from doing the guide, checking the package docs seems like a more straightforward progression for me. And maybe some newcomers won't even think that this is a FAQ, but that they're just missing something simple.

Oops, sorry about the documentation example format, I'll rectify that. Can I even put that into this PR, or do I need to close and open a new one?

jvoigtlaender:
> Can I even put that into this PR, or do I need to close and open a new one?

You can add commits to a PR. You can also squash commits inside a PR until it is merged. Force-pushing to PRs is frequently done by people around here.

cobalamin:
Alright, apparently that worked!

cobalamin:
Further digression on the FAQ point: I just noticed the link you shared is actually only in the subsection for specific 0.17 upgrading questions. It's linked with this text:

> See also the Elm 0.17 FAQ about upgrading to Elm 0.17.

As a newcomer right now I would pretty much never check this – "I've not used any version before this current one, so why check the upgrading notes?"

So thanks for the link, I'll share it with people in the Slack, but I think it's hard to find on your own for someone new.

jvoigtlaender:
Yes, you are right, and I also think that the FAQ needs restructuring as Elm 0.16 fades into history.

evancz:
Planning on dropping this operator. If it is not dropped in the final release, it will be added back with documentation.

nathanfox:
I personally like and extensively use the ! operator. Granted I can create my own if it is dropped in a future release, but I'd vote to keep it.

Issue #595, opened 2016-05-11
Ports do not register immediate `send`


knewter:
I've produced an example here: https://github.com/knewter/counter

It's actually published here: http://knewter.github.io/counter/Counter.html

If you load the page, the counter will have value 0.  The `Counter.html` file has the following code:

```javascript
var app = Elm.Main.fullscreen();
app.ports.jsMsgs.send(1);
window.app = app;
```

I mention the last bit because you can open up the browser's console, issue `app.ports.jsMsgs.send(1)`, and the counter will update.  So the inbound port works, but it doesn't work immediately.

knowthen:
I think this is the same bug as this one: https://github.com/elm-lang/core/issues/582

knewter:
it would appear to be the same bug.  This is a simpler case that produces the error though.  I didn't realize they were the same error before.

roine:
As a work around I'm using something like:

    setTimeout(function(){ app.ports.jsMsgs.send(1); }, 0)

That's not pretty but it does the job.

artisonian:
Why not use [programWithFlags](http://package.elm-lang.org/packages/elm-lang/html/1.0.0/Html-App#programWithFlags)?

knewter:
The point is that it is a bug.  The simplified use case is here to expose it. In most cases it is entirely reasonable to use `programWithFlags` for the same solution, but that doesn't change the fact that the bug shouldn't exist!  In my use case, it's problematic to explain how ports work to people if I have to work around a bug when introducing their simplest use case (this qualifies I think)

yoni-tock:
Yep, just hit this in our production app :(. Thanks to @lukewestby for pointing us at this bug!

vjousse:
Same here, pretty annoying when discovering elm.
You often start by integrating it to your existing JS code with ports and you can encounter this bug pretty early in your development process.

lukewestby:
@yoni-tock and I are going to pair on fixing this when he has some time in a couple weeks if no one can get to it earlier

kittykatattack:
Just letting you all know I've experienced this bug too:

https://www.reddit.com/r/elm/comments/4u3icn/incoming_port_not_working/

artisonian:
I've been looking into the issue. So far, I can tell that the app gets initialized using the scheduler (which doesn't do any work until the next turn of the event loop). I imagine three possible fixes:

[...]

PR #685, opened 2016-08-08
Candidate fix for #595

lukewestby:
This PR demonstrates one possible way to deal with calls to send prior to incoming ports being ready ( #595 ). As @artisonian mentioned, `send` can be called before `onEffects` in the port effect manager is called, so there are no `Sub`s around to handle the send. This fix demonstrates queuing things that are sent until `onEffects` has been called for the first time, and then pushing all enqueued values through during that first call.

By @yjkogan and myself.

We demonstrated that this fixes the issue in https://github.com/lukewestby/elm-port-bootup-sscce

### DIFF
--- a/src/Native/Platform.js
+++ b/src/Native/Platform.js
@@ -518,24 +518,40 @@ var incomingPortMap = F2(function subMap(tagger, finalTagger)
 
 function setupIncomingPort(name, callback)
 {
-	var onEffectsCalled = false;
-	var enqueuedBeforeSetup = [];
+	var sentBeforeInit = [];
 	var subs = _elm_lang$core$Native_List.Nil;
 	var converter = effectManagers[name].converter;
+	var currentOnEffects = preInitOnEffects;
+	var currentSend = preInitSend;
 
 	// CREATE MANAGER
 
 	var init = _elm_lang$core$Native_Scheduler.succeed(null);
 
-	function onEffects(router, subList, state)
+	function preInitOnEffects(router, subList, state)
+	{
+		var postInitResult = postInitOnEffects(router, subList, state);
+
+		for(var i = 0; i < sentBeforeInit.length; i++)
+		{
+			postInitSend(sentBeforeInit[i]);
+		}
+
+		sentBeforeInit = null; // to release objects held in queue
+		currentSend = postInitSend;
+		currentOnEffects = postInitOnEffects;
+		return postInitResult;
+	}
+
+	function postInitOnEffects(router, subList, state)
 	{
 		subs = subList;
-		if (!onEffectsCalled) {
-			onEffectsCalled = true;
-			enqueuedBeforeSetup.forEach(internalSend);
-			enqueuedBeforeSetup = null;
-		}
 		return init;
+	}
+
+	function onEffects(router, subList, state)
+	{
+		return currentOnEffects(router, subList, state);
 	}
 
 	effectManagers[name].init = init;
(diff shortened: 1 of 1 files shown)


src/Native/Platform.js
```diff
@@ -528,6 +530,11 @@ function setupIncomingPort(name, callback)
 	function onEffects(router, subList, state)
 	{
 		subs = subList;
+		if (!onEffectsCalled) {
+			onEffectsCalled = true;
+			enqueuedBeforeSetup.forEach(internalSend);
+			enqueuedBeforeSetup = null;
```
knewter: should this be turned into an empty array again?  Feels weird changing its type.

evancz:
Switch to use the style in all the rest of the JS. Notice that `{` is always brought down a line. Notice that `} else {` is always split onto three lines. Also, can we switch `forEach` out for a `for` loop? I'm not sure how old `forEach` is and I avoid things like this in the JS I write or generate as much as possible.

src/Native/Platform.js
evancz: I don't think it'll really matter.

evancz:
There are two alternate approaches.

  1. There is no such thing as `onEffectsCalled`, only `queue = []`. If it is `instanceof Array` you push values on it. If you send all the things, you switch `queue` to be `null` or `undefined` or something. This way there's no potential for synchronization issues between two different variables.

  2. Rather than having ifs, switch between two different send functions. The first one queues things. When it's time to switch, you send all the things and switch to the real send function.

Why is your approach better than these?

lukewestby:
Of the option we posted and the two alternatives I think alternative 2 is the best approach. It has the same benefit as alternative 1 of limiting the potential for variables to become out of sync and it also allows for the `onEffects` and `send` implementations in use for the majority of the application's lifetime to not need to check some condition that is no longer relevant.

I'll for sure change the `forEach` to a loop - `forEach` isn't present in IE8.

evancz:
Cool. Also, thanks for figuring this out and working on the PR!

lukewestby:
No problem! It seems to be affecting a number of people and I got to hang out and code with @yjkogan for a couple hours 😄  win/win/win

yjkogan:
Looks awesome!

evancz:
Cool :) The solution feels a little tricky, but I think that's inherent to the problem. I.e. This seems like the best way to do a hard thing. Point is, thanks @lukewestby and @yjkogan for sorting this one out! :D (Also hi!)

Do you think it makes sense to do 4.0.5 to get this out as quick as possible? I'll have to look back and see if there are any other changes on master, but that seems like the way to go to me.

lukewestby:
I'll defer to @yjkogan, @knewter and anyone else currently experiencing this in production to make a recommendation - I don't have any applications in this situation myself so I don't know how urgent it is.

Thanks!

knewter:
It's not urgent at all because you can always do a `setTimeout(fn(){}, 0)`, it's just nice to keep new people from seeing the confusing result.

evancz:
Eh, I just looked through the commits on master, and this is the only thing since 4.0.4 so I decided to just publish it. If you install things now, it should be 4.0.5 and this should be fixed.

Talk about it here if there are further issues. Otherwise, great work! :)

yjkogan:
Woohoo! Late to the party but same as @knewter. It was easy to work-around once we found the issue but I think this makes the logic simpler (and lets our site load/paint faster!)

Bipsy:
Thank you all for taking time to work on this issue!

PR #687, opened 2016-08-09
Support higher Unicode ranges for Char.fromCode

Janiczek:
```
> Char.fromCode 0x1F0A1 == Char.fromCode 0xF0A1
True : Bool

> Char.fromCode 0x1F0A1
'' : Char

> Char.fromCode 0xF0A1
'' : Char
```

I'm trying 0x1F0A1 = 🂡, I'm getting 0xF0A1 = .
That's because Elm uses JavaScript's `String.fromCharCode` which [doesn't support higher Unicode ranges](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode#Getting_it_to_work_with_higher_values).
To mitigate that, [`String.fromCodePoint`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint) should be used.

js_console

As a sidenote, maybe `toCode` could be also updated to use `String.codePointAt` instead of `String.charCodeAt`?

### DIFF
--- a/src/Native/Char.js
+++ b/src/Native/Char.js
@@ -3,7 +3,7 @@
 var _elm_lang$core$Native_Char = function() {
 
 return {
-	fromCode: function(c) { return _elm_lang$core$Native_Utils.chr(String.fromCharCode(c)); },
+	fromCode: function(c) { return _elm_lang$core$Native_Utils.chr(String.fromCodePoint(c)); },
 	toCode: function(c) { return c.charCodeAt(0); },
 	toUpper: function(c) { return _elm_lang$core$Native_Utils.chr(c.toUpperCase()); },
 	toLower: function(c) { return _elm_lang$core$Native_Utils.chr(c.toLowerCase()); },


lukewestby:
Unfortunately `String.fromCodePoint` [isn't well supported](http://kangax.github.io/compat-table/es6/#test-String_static_methods_String.fromCodePoint) by browsers so it would need to be polyfilled. Even a [low-level polyfill](https://github.com/mathiasbynens/String.fromCodePoint) of this function is a lot slower than the native, as demonstrated by the benchmark below

http://jsbin.com/taraguwaja/1/edit?js,console

evancz:
Fixed in https://github.com/elm-lang/core/commit/bf16ca8a10f29e271d61ca463a1268232e97ac79 with an implementation that is cross-browser and should be reasonably fast. In the BMP, it usually adds on conditional check.

@lukewestby, most polyfills have lots of code to handle all the bad data that could be passed in. What does `.codePointAt` do on objects? On arrays? Etc. In this case I just learned a lot about UTF-16 for some recent parser work, and it turns out that it's not so hard to sort out. I used the formulas from [here](https://mathiasbynens.be/notes/javascript-encoding) for surrogate pairs.

PR #702, opened 2016-08-28
add uriEncode and uriDecode

lukewestby:
These functions are available in evancz/elm-http but I believe they are more broadly useful - at least as useful as working with URLs in general. An example might be in combination with elm-lang/navigation and evancz/url-parser to deal with query strings. It would be great to be able to build things like query string parsers without including a dependency on evancz/elm-http.

Ordinarily this is a scenario where things would go in string-extra but since these require native calls I figured I'd start here.

### DIFF
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -151,6 +151,20 @@ function toLower(str)
 	return str.toLowerCase();
 }
 
+var missingEncodableCharacters = /[!'()*]/g;
+function uriEncode(string)
+{
+  return encodeURIComponent(string)
+		.replace(missingEncodableCharacters, function(c)
+		{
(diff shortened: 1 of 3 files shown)


evancz:
These exist `evancz/http` like [this](http://package.elm-lang.org/packages/evancz/elm-http/3.0.1/Http#uriEncode).

I think a better move would be to move `Http` into core, and it is on my todo list, but I'd like to get 0.18 out before doing something like that.

lukewestby:
:metal: sounds like a great move to me. thanks!

nelsonic:
Hi! where do  the `uriEncode` and `uriDecode` functions "_live_" now? (_thanks!_)

PR #706, opened 2016-09-06
added check for null process.root that arises when a process in the w…

johnnyhh:
…orkQueue is killed.

If I call `Process.kill` on a process in the workQueue, the process's root is set to null, but it stays in the workQueue, leading to a runtime error in `step()` at the line `var ctor = process.root.ctor`. This bug pops up quite frequently in when I kill a process whose task (http request) has completed, but not yet been processed and removed from the workQueue.

### DIFF
--- a/src/Native/Scheduler.js
+++ b/src/Native/Scheduler.js
@@ -131,6 +131,8 @@ function step(numSteps, process)
 {
 	while (numSteps < MAX_STEPS)
 	{
+    if (process.root === null) break;
+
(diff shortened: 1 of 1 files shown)


evancz:
You need to provide an http://sscce.org that uses `elm-lang/core` version 4.0.5. This issue should have been fixed in a much nicer way in that patch release.

If you can provide an http://sscce.org as an issue, we can then think about how to best fix the issue.

PR #707, opened 2016-09-07
Re-implement List.map in pure elm

fredcy:
See discussion at https://groups.google.com/forum/#!topic/elm-dev/He_emRjsly8

The objective is to build a pure-Elm version of map that has good performance for shorter lists and avoids stack overflow for long lists.

There are benchmarks at https://github.com/fredcy/elm-core-benchmark.

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -128,7 +128,48 @@ member x xs =
 -}
 map : (a -> b) -> List a -> List b
 map f xs =
-  foldr (\x acc -> f x :: acc) [] xs
+  mapHelp f xs 0
+
+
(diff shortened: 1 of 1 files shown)


evancz:
Based on your work, @Skinney created #872 which applied this technique to `foldr` and then implemented `map` in terms of that. He had some benchmarks that were pretty easy to run, so I played with those to tune the size vs speed.

Point is, `map` is implemented in Elm now through that! Sorry it took so long to get an answer on this, and thank you for the work! Turned out we could move a bunch of things into Elm with this!

PR #709, opened 2016-09-11
If doc.body is null, throw "move elm.js <script> "

devinrhode2:
Checkout the last few commits on my project here: https://github.com/devinrhode2/structured-elm-todomvc/tree/emberify2
If you rewind and replay the commits you will see logic behind my error message.

### DIFF
--- a/src/Native/Platform.js
+++ b/src/Native/Platform.js
@@ -21,6 +21,13 @@ function addPublicModule(object, name, main)
 
 	object['fullscreen'] = function fullscreen(flags)
 	{
+		// document.body needs to be defined here. Throw if it's not.
+		if (document.body === null || document.body === undefined) {
+			throw new Error('Elm is being initialized above the body tag, ' +
+			'document.body is: ' + document.body + '\n Elm needs to be initialized' +
+			' inside the <body> tag. Please move this script tag: ' +
+			'<script src="elm.js> back into the <body> tag of your index.html file');
+		}
 		return init(document.body, flags, true);
 	};
 }


devinrhode2:
haskell platform was taking a while to install, so I figured I'd just search for and directly edit the js.

devinrhode2:
here's some relevant screenshots.. 
screen shot 2016-09-10 at 8 08 43 pm

evenwithsomethinginlocalstorage

devinrhode2:
I have the most notes in my readme here: https://github.com/devinrhode2/structured-elm-todomvc/tree/47abbf1083bc9db806323cca1912f3c4a0828e7b#build-instructions

devinrhode2:
But most importantly, `document.body` is not defined for javascript in the <head> tag. That's all this is about, of course. Users should simply keep scripts in the body tag to avoid this issue. This is a widely recommended practice and where ember puts their script tags.

lukewestby:
I wonder if we could hide this detail from users by wrapping startup with

```js
function startElmProgram()
{
  // ...
}

if (document.readyState === 'complete')
{
  startElmProgram();
}
else
{
  document.addEventListener('DOMContentLoaded', startElmProgram);
}
```

devinrhode2:
No. That's way less robust. Elm should not fuck with setting up a
document.ready event here in my opinion. If you really want to do that,
copy out the document.ready our of the latest version of jQuery. I don't
think this code structure would work well with that. There's probably a
handful of other good reasons to put scripts at the bottom.

We should, however, hide this from users by creating a proper index.html
file for users. I copied out of few good but missing tags from embers
index.html, however, there's no compile time hook to add content to the
head tag. (will share some links in a sec)

devinrhode2:
So this is the default index.html template for a new ember app:
https://github.com/rwjblue/--new-app-blueprint/blob/modules/src/ui/index.html

Here's a real world index.html that is filled out more (obviously not all
this is quite as important as the things in the default template):
https://github.com/rwjblue/--ghost-modules-sample/blob/grouped-collections/src/ui/index.html
However... I'm probably incorrect. I think ultimately elm shouldn't be
trying to work with the dom until it's ready. I don't know anything about
this code base and architecture. However, I don't think the whole app
startup should wait for document.ready. It should simply be the rendering
layer. This is why I intuitively and quickly responded negatively to your
suggestion

rtfeldman:
> No. That's way less robust. Elm should not fuck with

Let's keep it civil. :)

@devinrhode2 I like the general UX idea here, but have you seen this? https://github.com/elm-lang/core/pull/709#issuecomment-246212555

devinrhode2:
Ok let's try this again: https://github.com/elm-lang/core/pull/711

PR #713, opened 2016-09-16
Catch runtime error when creating a Regex from an invalid expression

aische:
Giving an invalid regex string to the function 'regex' crashes Elm. This is bad if you want to create a regex from user input. A function safeRegex could be added that tries to create the regex, and returns 'Just Regex' if it succeeds, or Nothing in case of an error.

### DIFF
--- a/src/Native/Regex.js
+++ b/src/Native/Regex.js
@@ -15,9 +15,13 @@ function regex(raw)
 	return new RegExp(raw, 'g');
 }
 function safeRegex(raw)
-{	try {
+{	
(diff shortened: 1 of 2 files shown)


jvoigtlaender:
This is kind of a duplicate of https://github.com/elm-lang/core/pull/380. I suggest closing it, and continuing discussion there (if need for discussion, rather than decision, exists).

aische:
Ah I did not see that. Sorry for duplication and the noise!

Issue #712, opened 2016-09-15
Potential zalgo in ports


wishfoundry:
if more than 1 subscriber exists on a port, and if that port attempts to unsubscribe while in an event, there is potential to skip the next listener on that port

https://github.com/elm-lang/core/blob/master/src/Native/Platform.js#L486

jvoigtlaender:
Provide an http://sscce.org/ that manifests a bug from this?

lukewestby:
For a little bit more context:

This manifests itself because subs are iterated in a for loop when a message comes in:

https://github.com/elm-lang/core/blob/master/src/Native/Platform.js#L462-L465

```elm
for (var i = 0; i < subs.length; i++)
{
  subs[i](value);
}
```

If there are 3 callbacks in `subs`, and `subs[1]` removes itself from the list in place by unsubscribing, then `subs[2]` becomes `subs[1]` and the for loop ends because `i` is now equal to `subs.length` but than final callback has never been called. To avoid this other systems that employ a list of subscribers will make a copy of the subscribers array with `Array.prototype.slice()` at opportune moments. The assumption there is that the subscribers array is generally small so that copying is fast, and I think that's a fair assumption here as well.

lukewestby:
I've already started a fix, but in the mean time here's an sscce

_Main.elm_
```elm
port module Main exposing (..)

import Html
import Html.App
import Time exposing (Time)


type alias Model =
    ()


port test : Time -> Cmd msg


type Msg
    = Tick Time


update : Msg -> Model -> ( Model, Cmd Msg )
update (Tick time) _ =
    ( ()
    , test time
    )


subscriptions : Model -> Sub Msg
subscriptions _ =
    Time.every Time.second Tick


main : Program Never
main =
    Html.App.program
        { view = \_ -> Html.text ""
        , update = update
        , subscriptions = subscriptions
        , init = ( (), Cmd.none )
        }
```

_index.html_
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title></title>
  </head>
  <body>
    <script src="out.js"></script>
    <script>
      var app = Elm.Main.fullscreen();
      app.ports.test.subscribe(function unsub(time) {
        console.log('first sub got called with', time);
        app.ports.test.unsubscribe(unsub);
      });
      app.ports.test.subscribe(function (time) {
        console.log('second sub got called with', time);
      });
    </script>
  </body>
</html>
```

[...]

PR #715, opened 2016-09-20
Fix for #712 - don't skip next sub if previous sub unsubscribes itself

lukewestby:
### The Problem
It's possible that someone can unsubscribe from a port in JavaScript in the same tick that the port's subscription callback is called. Since the array that stores these subscription callbacks is shared and mutated in place by `subscribe` and `unsubscribe` that unsubscribing in the same tick as a callback can cause the following subscription to be skipped. This is all demonstrated and explained in #712.

### Prior art

See https://github.com/reactjs/redux/blob/master/src/createStore.js#L63-L67

### The fix
This fix solves that issue by making the subs array immutable with respect to unsubscribing and iterating. When a subscription callback drops itself the subs array will be cloned before the callback is spliced out. Then, during iteration when a message comes in a separate reference is made to the subs array so that changes to that array are only picked up once iteration moves on to the next message.

### Expected behavior after merging
The result is that when a subscription callback drops itself, we wait for the next message before pulling in changes made to the subs array while processing the previous message. If there are multiple messages waiting for the port, the subscription that was removed will only process the ones that came in before it unsubscribed itself.

### Alternatives
I think most alternatives to this would just be variations on the underlying idea of cloning and re-referencing at a particular point. I think this one has the lowest memory and performance cost as opposed to cloning before processing messages, for example.

### DIFF
--- a/src/Native/Platform.js
+++ b/src/Native/Platform.js
@@ -458,10 +458,11 @@ function setupOutgoingPort(name)
 	{
 		while (cmdList.ctor !== '[]')
 		{
+			var currentSubs = subs;
 			var value = converter(cmdList._0);
-			for (var i = 0; i < subs.length; i++)
+			for (var i = 0; i < currentSubs.length; i++)
 			{
-				subs[i](value);
+				currentSubs[i](value);
 			}
 			cmdList = cmdList._1;
 		}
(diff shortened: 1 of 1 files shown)


evancz:
This approach makes sense to me. It seems like the simplest nicest way to do this. Nice work!

Also, this PR is great :D

lukewestby:
:metal: thanks evan!

PR #718, opened 2016-09-21
Add support for native file objects

dawehner:
For some drag and drop support it would be nice if we could decode File objects natively.
This PR adds support for just, and really just that.

### DIFF
--- a/src/Native/Json.js
+++ b/src/Native/Json.js
@@ -508,6 +508,21 @@ function runHelp(decoder, value)
 			}
 			return badOneOf(errors);
 
+		case 'file':
+			var errors = [];
+			if (value instanceof File)
+ 			{
+				return {
+					ctor: '_File',
+					lastModified: value.lastModified,
+					name: value.name,
+				};
+			}
+			else
+			{
+				return badPrimitive('the object has to be of type File', value);
+			}
+
 		case 'fail':
 			return bad(decoder.msg);
 


dawehner:
I like the process bot!

dawehner:
Thank you for the review!
Oh these are actually tabs. I got quite confused when I looked at the git diffs earlier. I'll work on that.

dawehner:
I hope this fixes the CS issues

lukewestby:
Hey @dawehner!

Thanks for the contribution. So if you read the stuff that process bot shared or watch [Evan's talk from Elm Conf](https://www.youtube.com/watch?v=DSjbTC-hvqQ) you'll find that a better way to go about stuff like this is to jump into [Slack](elmlang.herokuapp.com) or the [Elm Discuss mailing list](https://groups.google.com/forum/#!forum/elm-discuss) to find out if the thing you want to do is actually possible without changes given your use case. If it turns out that some change would be needed it's best to discuss it on the [Elm Dev mailing list](https://groups.google.com/forum/#!forum/elm-dev) before writing any code. It might turn out that your idea fits in well with some broader effort other folks are also interested in. Sound good?

dawehner:
It does. In general though it was a fun journey through elm, seeing that its just JS at the end of the day (yeah I'm Mr. obvious ;) )

I'm hanging out in #elm-dev now

lukewestby:
Great! I'm gonna close this now but it'll still be here to reopen if it turns out this was the code we needed all along. Thanks!

dawehner:
@lukewestby Can you give me any pointer of an existing issue around that kind of problem space?

PR #734, opened 2016-10-21
Fix compatibility with google closure advanced optimizations

Skinney:
### Problem

Crash or silent errors when compiled with google closure advanced optimizations.

### Steps to reproduce

Compile the test suite with `tests/run-tests.sh`, take the resulting `test.js` script and compile it with google closure with advanced optimizations (can be done [here](http://closure-compiler.appspot.com/)). Run the result through node.

### Why this commit fixes things

When compiled with advanced optimizations, the closure compiler will be so bold as to change the actual names of object properties. This means that an object like this:

```js
{
    ctor: '_Tuple2',
    _1: 27,
    _2: "Robin"
}
```

Can end up looking like this after minification (pretty printed):

```js
{
    a: '_Tuple2',
    b: 27,
    c: "Robin"
}
```

For the most part, this is just fine in the context of Elm. There were, however, certain places where explicit checks for the `ctor` property was made by using `in` checks, like this: `if ('ctor' in obj) ...`.

The closure compiler will never change the contents of strings, so the result is that a few places, Elm will look for a `ctor` property which no longer exists, as it has been renamed.

This commit fixes the few places this caused a problem, and so this elm package should be compatible with google closures advanced optimizations.

### How much does advanced optimizations matter?

On a hello world type Elm application, we get the following (without gzip):

```
232K	original.js
80K	        uglify.js
76K	        closure.simple.js
72K	        closure.advanced.js
```

### DIFF
--- a/src/Native/Utils.js
+++ b/src/Native/Utils.js
@@ -53,7 +53,7 @@ function eqHelp(x, y, depth, stack)
 		return x.getTime() === y.getTime();
 	}
 
-	if (!('ctor' in x))
+	if (typeof x.ctor === 'undefined')
 	{
 		for (var key in x)
 		{
@@ -348,7 +348,7 @@ function toString(v)
 		return 'null';
 	}
 
-	if (type === 'object' && 'ctor' in v)
+	if (type === 'object' && typeof v.ctor !== 'undefined')
 	{
 		var ctorStarter = v.ctor.substring(0, 5);
 
@@ -357,8 +357,9 @@ function toString(v)
 			var output = [];
 			for (var k in v)
 			{
-				if (k === 'ctor') continue;
-				output.push(toString(v[k]));
+				var val = v[k];
+				if (val === v.ctor) continue;
+				output.push(toString(val));
 			}
 			return '(' + output.join(',') + ')';
 		}
@@ -414,8 +415,9 @@ function toString(v)
 		var output = '';
 		for (var i in v)
 		{
-			if (i === 'ctor') continue;
-			var str = toString(v[i]);
+			var val = v[i];
+			if (val === v.ctor) continue;
+			var str = toString(val);
 			var c0 = str[0];
 			var parenless = c0 === '{' || c0 === '(' || c0 === '<' || c0 === '"' || str.indexOf(' ') < 0;
 			output += ' ' + (parenless ? str : '(' + str + ')');


evancz:
How comprehensively have you tested this? I'd be curious to know what grepping for `"ctor"` and `'ctor'` in all code with native parts would yield. It's fine if you don't have it all, but I'd like to know what you did do this on at least.

src/Native/Utils.js
```diff
@@ -414,8 +415,9 @@ function toString(v)
 		var output = '';
 		for (var i in v)
 		{
-			if (i === 'ctor') continue;
-			var str = toString(v[i]);
+			var val = v[i];
+			if (val === v.ctor) continue;
```
Skinney: Right. Hmm. One other option is to always specify `ctor` as a string, both on creation and on retrieval, but that's a much bigger undertaking.

`{'ctor': '_Tuple0'}`
`obj['ctor']`

Skinney:
I compiled a small application I've just started on and read/skimmed the code looking for constructs I knew closure would have a problem with, I then created a small benchmark to make sure I didn't seriously ruin performance (I didn't). I also grepped for 'ctor' and "ctor" just before submitting this, as a sanity check. So I think core and virtual-dom/html should be fine.

src/Native/Utils.js
evancz: I'm looking into ways to get assets really small through other means, so I'm not sure this kind of stuff is gonna matter that much. Can we quantify how much advanced mode helps with size before getting into PRs to do that? Like, is it 20kb to 19kb? Is it making things faster? From there, we can assess better.

evancz:
I am thinking of other ways of making assets small, so I think it's worth quantifying the benefit that would come from this if we can make it work. So my questions are:

  - How much smaller does this make things? If the minified and gzipped size is going from 24kb to 22kb, I'm not sure how invasive we should be to achieve that.
  - Is the resulting code faster in any measurable way? If it's not faster, I'm also not sure how aggressively we should pursue this.

src/Native/Utils.js
```diff
@@ -357,8 +357,9 @@ function toString(v)
 			var output = [];
 			for (var k in v)
 			{
-				if (k === 'ctor') continue;
-				output.push(toString(v[k]));
+				var val = v[k];
+				if (val === v.ctor) continue;
```
eeue56: (same problem as brought up by evan, didn't see their comment til after!)

Skinney:
I added some quick measurements on the top of the file. The gzipped difference between simple and advanced optimizations is only 1-2kb. It's probably worth looking at a larger project though, as this is pretty close to a hello world type application.

Anyway, I'm not in any personal need to rush this, I just saw an opportunity to fix something people mention now and again with a small commit. Since my proposed solution here is flawed, I've no problems with putting a low priority on this one.

I'll try to compile my array test suite with advanced optimizations tomorrow, and see if it makes a performance difference.

evancz:
Okay, let's close this one for now. We can open a new one if we figure out a safe way to do this / know the size and perf numbers on larger projects.

PR #736, opened 2016-10-22
Set the `index` parameter accurately in Regex.replace.

### LINKED ISSUE #483: Regex.replace does not set the index accurately
In this line in `Regex.replace`:

https://github.com/elm-lang/core/blob/1b28ce5ce8f30ddf549e5f454a772f7b623e09b4/src/Native/Regex.js#L90

there are two problems.

First, the variable `i` will inevitably be equal to `-1` at this point. Thus, saying `i - 1` is necessarily equivalent to `-2`, but obscure.

Second, the `arguments` object is not really an array, and does not support negative indexes. So, the `index` property is always set to undefined.

So, it would be better to replace this line with something like:

 `index: arguments[arguments.length - 2],`

rgrempel:
Fixes #483.

### DIFF
--- a/src/Native/Regex.js
+++ b/src/Native/Regex.js
@@ -75,7 +75,7 @@ function replace(n, re, replacer, string)
 		return replacer({
 			match: match,
 			submatches: _elm_lang$core$Native_List.fromArray(submatches),
-			index: arguments[i - 1],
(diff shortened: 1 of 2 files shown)


evancz:
Thanks!

PR #737, opened 2016-10-22
Reset `lastIndex` in `Regex.split` so that it is idempotent.

### LINKED ISSUE #482: Regex.split modifies the state of the regex
`Regex.split` modifies the state of the `regex` parameter, because it does not reset the `lastIndex` property.

This is essentially the same problem that `Regex.find` had, before the following commit:

https://github.com/jonathanhefner/elm-core/commit/3fcf5831d3ef5c8f6c324c092ea937e5fc3c156b

So, `Regex.split` essentially needs a similar sort of fix.

rgrempel:
Fixes #482.

### DIFF
--- a/src/Native/Regex.js
+++ b/src/Native/Regex.js
@@ -93,6 +93,7 @@ function split(n, re, str)
 	var result;
 	var out = [];
 	var start = re.lastIndex;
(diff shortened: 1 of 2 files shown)


evancz:
Great, thank you!

PR #739, opened 2016-10-26
Adds international date formatting

shutej:


bowsersenior:
Please merge this pull request ASAP or similar functionality. Basic UTC support is a requirement for any serious usage. My colleagues will laugh at me if I introduce Elm to them and they find out there is no UTC functionality in it's Time/Date methods.

ktonon:
I agree that timezone support should be part of the core `Date` library, but until either this PR or #755 get merged, there is http://package.elm-lang.org/packages/elm-community/elm-time/1.0.2/Time-ZonedDateTime

ktonon:
Actually, I don't think `ZonedDateTime` will cut it. It doesn't provide a way to get the current datetime! You could just use http://package.elm-lang.org/packages/elm-lang/core/5.1.1/Date#now to get the current date, and then create a `ZonedDateTime` from that, but then you have no way of  knowing what the timezone is in the first place!

ktonon:
https://github.com/justinmimbs/elm-date-extra/blob/2.0.3/src/Date/Internal/Extract.elm#L55

Shouldn't have to go through this much trouble to get the timezone offset. Package maintainers, please merge, or at least comment on this PR or #755

Zensavona:
...and now we're in May

evancz:
### General Comments

I have tried to explain with @process-bot how these sorts of issues are handled. I hope @ktonon and others will read the links in that comment that get into [batching work](https://github.com/process-bot/contribution-checklist/blob/master/expectations.md) which is elaborated on much more [here](https://youtu.be/DSjbTC-hvqQ?t=18m55s).

I have also provided [this document](https://github.com/elm-lang/projects/blob/master/roadmap.md) to help explain some of the near-term prioritization choices.

### Specific Comments

In this particular case, dealing with dates is very complicated. To address it fully, I would want to review a broad set of usage scenarios to make sure I understood all the possible needs people will have. From there, I would want to review how this is addressed in a bunch of other languages. From there, I would want to find the simplest possible API that is nice to use in practice. This is all hard to do, and a PR with a totally blank description basically does none of this *necessary* background work.

Now, assuming that this PR was great and ready to go, we take releases of `elm-lang/core` very seriously. Every single library in the whole ecosystem depends on core, so it is important to be thorough. In that spirit, it makes sense to align releases of core with releases of the compiler. That means we get many weeks of testing in real codebases before it goes out to everyone in the ecosystem.

### Summary

To take a step back, the big picture prioritization decisions look [like this](https://github.com/elm-lang/projects/blob/master/roadmap.md). To spend the week or more to actually do a *good* release of core means we push some other thing back by a week. Expanding the Web Platform support within Elm is already slated as "the next big project" as documented [here](https://github.com/elm-lang/projects/blob/master/roadmap.md#where-is-the-localstorage-package) and would include stuff like this. As I said before, I think the [batching approach](https://youtu.be/DSjbTC-hvqQ?t=18m55s) has been important to Elm's success, and I am very conscious that in spite of the results it has produced, this process does not suit everyone.

I think it makes sense to end by reemphasizing [this explanation](https://github.com/elm-lang/projects/blob/master/roadmap.md#where-is-the-localstorage-package).

I hope this also helps the folks seeing this issue on HN and wondering if there is some broader context. I also hope that it is no surprise that there is.

evancz:
As I have learned more about producing accurate date information, it seems like it is worth trying to make something like `elm-community/elm-time` as solid and nice to use as possible. I do not have faith that we can build on top of JavaScript's `Date` functions and build anything particularly nice.

Issue #738, opened 2016-10-26
Core provides List.unzip but not List.zip


ceph3us:
I came to Elm after playing with a few functional languages before briefly and was surprised to find that `List.unzip` exists, but not `List.zip`. I know that it's trivial to derive `zip` from `List.map2`, but it seems strange to provide one but not the other.

jvoigtlaender:
This has been discussed on the mailing list in the past. The current situation is deliberate. So this is not reporting a bug or surprising deficiency of the library. Feature requests are usually not discussed here on GitHub, but instead on the mailing list. Moreover, there is the http://package.elm-lang.org/packages/elm-community/list-extra package that does contain `List.Extra.zip`. The guidelines for this repository here explicitly say that new functions should be proposed for such extra packages instead of being proposed straight for inclusion into `core`: https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions

Given all the above, I'm closing the issue here.

jvoigtlaender:
Ask on the mailing list, or search the archive of the mailing list for previous discussion on this. 

Everything else said in my previous comment above stands as well.

danpalmer:
For those finding this in the future and wondering what the discussion was on the mailing list, the only discussion I could find (on elm-discuss or elm-dev) is [this one](https://groups.google.com/forum/#!searchin/elm-discuss/zip|sort:relevance/elm-discuss/CymE5cijGpc/Hynt20LB9EgJ) where several people raise the confusion that the author here raises. There appears to be no conclusion on the discussion.

I understand the contribution guidelines mean this is not the right forum for proposing this change, but it looks like there's some support for it, so it might be worth a proposal on the mailing list.

la-yumba:
for what it's worth, I too think that `zip` should be provided in the core lib, as it's more intuitive than `map2 (,)`

I thought that Elm aimed at being intuitive and lowering the barrier to entry - this choice seems to go against this philosophy

PR #743, opened 2016-10-30
Add Dict.filterMap, analogous to List.filterMap

kirchner:
Add this function, so you don't have to convert the dictionary to a list, apply List.filterMap and then convert back to a dictionary.

### DIFF
--- a/src/Dict.elm
+++ b/src/Dict.elm
@@ -29,7 +29,7 @@ Insert, remove, and query operations all take *O(log n)* time.
 @docs keys, values, toList, fromList
 
 # Transform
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Please read https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions, and as a longer comment/explanation also https://github.com/elm-lang/core/issues/738#issuecomment-256410124.

I'm closing here.

jvoigtlaender:
http://package.elm-lang.org/packages/elm-community/dict-extra/latest would be a natural place to add your function.

Issue #387, opened 2015-09-02
How to handle characters outside the BMP


ianbollinger:
Manipulating characters outside the Unicode BMP will produce a crash on the REPL (on Fedora but not Mac OS X):

```
> String.toList (String.fromList [ '😣' ])
```

...or...

```
> Char.fromCode (Char.toCode '😣')
```

results in

```
elm-repl: fd:7: hGetContents: invalid argument (invalid byte sequence)
```

It seems that either:

1. The Elm compiler should issue a syntax error if a character literal falls outside of the BMP.
2. The Elm core library should be made to work with characters outside the BMP (harder, since JavaScript doesn't work well with them either.)

evancz:
Can you provide more info on how BMP works and how other systems handle this. I believe we are using the underlying JS stuff for this, so I'd like to know how they do this as well. Overall, give more context and we can figure it out from there.

ianbollinger:
Well, I'd say it's an Elm problem because the underlying problem is that you can create `Chars` that are four-bytes wide (instead of two-bytes wide) and then do weird things with them. For instance, '😣' is codepoint 0x1F623, but `Char.toCode '😣' /= 0x1F623`. Thus `fromCode` and `toCode` aren't strictly inverse. Another example is String.toList, which will split a 4-byte codepoint down the middle. Finally, `List.length ['😣'] /= String.length (String.fromList ['😣'])`.

evancz:
Okay, going to continue tracking in #725 so we can consider all the string oddities in one place. Thanks for reporting this!

martincerny:
I think the core should give up on Unicode code points/characters and treat strings as sequences of 16bit words. The thing is that the mapping between Unicode code points and characters as perceived by the user is neither total, nor injective, nor surjective and thus determining character boundaries in Unicode strings is difficult and expensive. 

Further, having the correct character boundaries is of little practical importance. For almost all tasks in web development, I simply pass strings around. Further, all common operations (search for substring, search for a BMP character etc.) work without any special consideration for BMP characters. In particular, parsing XML, JSON, etc. works just fine.

The implication for the case above is that  '😣' would be an invalid character constant, while "😣" would be a valid string. If anyone needs fancy character-boundaries-aware text processing and is willing to pay the performance penalty, it should be dealt with in a specialized UTF package.

Sourcing a bit from http://utf8everywhere.org/ :-)

feihong:
[...]

PR #747, opened 2016-11-02
Make `Char.toCode` Roundtrip

crazymykl:
Synopsis
--
It's quite surprising that certain `Char`s cannot be converted to integers and back losslessly. 

Open Questions
--
- I have made `Char.fromCode` keep its existing behavior when given an `Int` outside of the range of valid Unicode. Should it be made `KeyCode -> Maybe Char` instead?

### DIFF
--- a/src/Native/Char.js
+++ b/src/Native/Char.js
@@ -2,9 +2,35 @@
 
 var _elm_lang$core$Native_Char = function() {
 
+function fromCodepoint(cp)
+{
+	if (cp > 0xFFFF && cp <= 0x10FFFF)
+	{
+		cp -= 0x10000;
+		return String.fromCharCode(Math.floor(cp / 0x400) + 0xD800, cp % 0x400 + 0xDC00);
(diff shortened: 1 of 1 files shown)


crazymykl:
This adresses #387.

crazymykl:
I spoke with @rtfeldman , and losing `O(1)` length is a hard-no. I'd still like the functions on `Char` that convert from keycodes to roundtrip.

jahewson:
Can a `KeyCode` actually contain a surrogate pair? As far as I'm aware [KeyboardEvent.keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) generates only ASCII and Windows-1252 characters.

Incidentally, a `KeyCode` from the Keyboard library is not even guaranteed to be Unicode, it could be Windows-1252. Seems like this could be a broader problem for Elm, which appears to assume that a `KeyCode` is a Unicode code point.

PR #759, opened 2016-11-24
Remove examples that used the old range syntax

ryantriangles:
If the range literal syntax has been removed from the language, then it ought to be removed from documentation examples too -- and I think a simple `[1,2,3,4,5]` is nicer and clearer than `(List.range 1 5)`, even if it's pretty obvious what the latter does without looking.


evancz:
Ah, I made these changes as well before seeing this. Thanks for pointing this out though!

Issue #754, opened 2016-11-16
suspicious code in Array


wishfoundry:
I noticed this line
```javascript
var slot = new createNode(a.height - 1, 0);
```
in:
https://github.com/elm-lang/core/blob/master/src/Native/Array.js#L730

createNode() is the factory here, it should not be new'ed up itself

I'm entirely sure where this will manifest though. If the being appended are balanced it probably can't reach this code.

wishfoundry:
actually this looks like the source of the problem in #474

turboMaCk:
+1

mgold:
Great detective work, but it won't be needed: there's a [fully-Elm version of Array](https://groups.google.com/forum/#!topic/elm-discuss/aAMVveQAG8c) in the final stages of testing.

evancz:
I'm interested in this fix in the meantime. That new stuff needs a performance review, and if this is an easy fix, it won't hurt!

So can someone do a PR with the change and a test that demonstrates what it fixes?

lukewestby:
@wishfoundry how did you demonstrate that this is the cause of the issue in #474? i'm setting a breakpoint at that line while running the examples from that issue and the line in question isn't being reached.

while the use of `new` is semantically incorrect here, it doesn't change the behavior of pure functions like `createNode`.

<img width="305" alt="screen shot 2016-11-29 at 6 51 24 pm" src="https://cloud.githubusercontent.com/assets/1508245/20735319/edbc2e7e-b664-11e6-9fab-6a7e43e2400a.png">

wishfoundry:
@lukewestby newing up a function can prevent it from returning a value:

<img width="256" alt="screen shot 2016-11-29 at 9 26 31 pm" src="https://cloud.githubusercontent.com/assets/818080/20740912/afd6088e-b67b-11e6-92fe-2ea8d6f74ad1.png">

I discovered the bug while attempting to port the rrb code to plain js, during a battery of performance comps against other immutable libs.
The rrb version would fail on any tests with lengths > 32, < 1024. I removed the "new", and they ran again. but I'm still a little unclear the exact reason. but if there is still a future in the current Array code, I'll update here as soon as I get a few moment to rerun them

wishfoundry:
Finally got it. the issue has to do with the final balancing act done in:

https://github.com/elm-lang/core/blob/master/src/Native/Array.js#L644

[...]

Issue #474, opened 2016-01-06
Runtime error with arrays


ckoster22:
Copy/paste into http://elm-lang.org/try

```
import Array exposing (Array)

type alias Box =
  { value : String }

convertTo2D : Array Box -> Int -> Array (Array Box)
convertTo2D boxArray numRows =
  let
    nextRow = Array.slice 0 numRows boxArray
  in
    if Array.length nextRow > 0
    then Array.push nextRow (convertTo2D (Array.slice numRows (Array.length boxArray) boxArray) numRows)
    else Array.empty

oneDimArray = Array.initialize 100 (\index -> { value = toString index })

temp = convertTo2D oneDimArray 10
```

This is boiled down from a larger application. The relevant error is:
"Uncaught TypeError: Cannot read property 'height' of undefined  everything.js:831"

jvoigtlaender:
Did you check it's not one of these [existing issues about arrays](https://github.com/elm-lang/core/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+arrays)?

ckoster22:
I checked. It could possibly be this issue (https://github.com/elm-lang/core/issues/349) but it's unclear whether it's the same. None of them mention the same runtime error.

jvoigtlaender:
Thanks!

jvoigtlaender:
I've played a bit with your example. By unrolling the recursion a bit (repeatedly inlining `convert2D`) and trying with reduced overall size, I reduced it to this:
```elm
import Array

temp = 
  let
    n = 10
  in 
    Array.initialize (4*n) identity
    |> Array.slice n (4*n)
    |> Array.slice n (3*n)
    |> Array.slice n (2*n)
    |> Array.slice n n
```
This shows the same error.

And interestingly, changing `n = 10` to `n = 8` lets the error go away (while it remains for `n = 9`). Noticing that for `n = 8` we have `4*n = 32`, the "magic number" from #349, it might have something to do with that issue after all, or maybe share a common root problem with that one...

evancz:
Closing in favor of #649 which will track all array issues and progress on a new implementation.

wishfoundry:
I suspect this line is the culprit:
```javascript
var slot = new createNode(a.height - 1, 0);
```
in:
https://github.com/elm-lang/core/blob/master/src/Native/Array.js#L730

createNode() generates an new node, it should not be new'ed up itself

PR #765, opened 2016-11-30
Array slice FIX + TEST

turboMaCk:
Replicates issue https://github.com/elm-lang/core/issues/474 in tests
also related to https://github.com/elm-lang/core/issues/754
## Array Slice issue

This test should replicate issue described in https://github.com/elm-lang/core/issues/474 and https://github.com/elm-lang/core/issues/754.
~~Anyway it's quite hard to proof this since master build is failing.~~

base: https://github.com/elm-lang/core/commit/1de91aec07e289da4ebb28dd769ab809fe555d27

### DIFF
--- a/src/Native/Array.js
+++ b/src/Native/Array.js
@@ -419,6 +419,11 @@ function slice(from, to, a)
 
 function sliceRight(to, a)
 {
+	if (!a) 
+	{
+		a = empty;
+	}
+
 	if (to === length(a))
 	{
 		return a;
@@ -459,6 +464,11 @@ function sliceRight(to, a)
 
 function sliceLeft(from, a)
 {
+	if (!a) 
+	{
+		a = empty;
+	}
+
 	if (from === 0)
 	{
 		return a;
@@ -815,6 +825,11 @@ function nodeCopy(a)
 // Returns how many items are in the tree.
 function length(array)
 {
+	if (!array)
+	{
+		return 0;
+	}
+	
 	if (array.height === 0)
 	{
 		return array.table.length;
--- a/tests/Test/Array.elm
+++ b/tests/Test/Array.elm
@@ -6,7 +6,6 @@ import List
 import List exposing ((::))
 import Maybe exposing (..)
 import Native.Array
-
 import Test exposing (..)
 import Expect
 
@@ -106,6 +105,12 @@ tests =
                 , test "filter" <| \() -> Expect.equal (Array.fromList [ 2, 4, 6 ]) (Array.filter (\x -> x % 2 == 0) (Array.fromList <| List.range 1 6))
                 ]
 
+        sliceTests =
+            describe "Slice Array"
+                [ test "array < 32" <| \() -> Expect.equal (Array.initialize 31 identity |> Array.slice 31 32) (Array.fromList [])
+                , test "array == 32" <| \() -> Expect.equal (Array.initialize 32 identity |> Array.slice 32 33) (Array.fromList [])
+                ]
+
         nativeTests =
             describe "Conversion to JS Arrays"
                 [ test "jsArrays" <| \() -> Expect.equal (Array.fromList <| List.range 1 1100) (Native.Array.fromJSArray (Native.Array.toJSArray (Array.fromList <| List.range 1 1100)))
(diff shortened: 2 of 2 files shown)


tests/Test/Array.elm
```diff
@@ -80,6 +80,11 @@ tests =
         , test "filter" <| assertEqual (Array.fromList [2,4,6]) (Array.filter (\x -> x % 2 == 0) (Array.fromList [1..6]))
         ]
 
+      sliceArrayTest = suite "Slice Array"
+        [ test "array < 32" <| assertEqual (Array.initialize 31 identity |> Array.slice 31 32) (Array.fromList [])
```
turboMaCk: this test should pass

tests/Test/Array.elm
```diff
@@ -80,6 +80,11 @@ tests =
...
+        , test "array == 32" <| assertEqual (Array.initialize 32 identity |> Array.slice 32 33) (Array.fromList [])
```
turboMaCk: this test should fail

turboMaCk:
I'm not able to make it pass in CI for some reason. However this works as expected on my local machine.

evancz:
It should get elm 0.18, not 0.17.1.

Also, these tests will be broken right? I think it's better to have the fix and the tests that prove the fix in one PR.

turboMaCk:
Hello!

Sorry I was traveling because of the work for past few days.

First let me explain why I did what I did till now. Tests on master are currently failing. I think this make it really harder to reproduce (/describe) any issue and proof possible fix. I looked at commits in master and found latest which passed. I wanted to demonstrate this issue using tests. This is why this PR uses that commit as Base.

However I think this will be a bit bigger issue and longer run so this is what I did next:

I've created new repository which is using core lib as submodule and build new test suit (0.18.0 !!) around core. I have to admit it was quite challenging itself. However I've managed to do it so this repository https://github.com/turboMaCk/elm-array-tests is ready I think.

Next steps will be:

- Move whole test suit for Array to that repository and make it pass
- Add tests for cases for issues with current implementation
- Try to fix some of the biggest issues
- Based on what we've learn about all of this update tests and in [new implementation](https://github.com/Skinney/elm-array-exploration)

Hope it make sense. Let me know if you have any objections or ideas.

turboMaCk:
Failing test is this one:
https://github.com/turboMaCk/elm-array-tests/pull/2/commits/7674f75e2b5c2e718c625c519c713de78907df19

eeue56:
There is already a fix, in this PR -> https://github.com/elm-lang/core/pull/399
I looked into this issue a lot and you can read my breakdown there.
It was closed as a pure elm version was decided to be the better route.

src/Native/Array.js
```diff
@@ -815,6 +834,7 @@ function nodeCopy(a)
 // Returns how many items are in the tree.
 function length(array)
 {
+  if (!array) { return  0; }
```
turboMaCk: part of actual fix.

src/Native/Array.js
```diff
@@ -459,15 +472,21 @@ function sliceRight(to, a)
 
 function sliceLeft(from, a)
 {
+  if (!a) {
+    a = empty;
+  }
```
turboMaCk: part of actual fix.

src/Native/Array.js
```diff
@@ -419,16 +423,25 @@ function slice(from, to, a)
 
 function sliceRight(to, a)
 {
+  if (!a) {
+    a = empty;
+  }
```
turboMaCk: part of actual fix.

turboMaCk:
## Rebased agains master!

This is runtime exception without fix: https://travis-ci.org/elm-lang/core/builds/181118399#L248

This is proof that this is solution to that problem: https://travis-ci.org/elm-lang/core/builds/181118740#L253

turboMaCk:
## Alternative approach

I would also like to mention one more time that there are some alternative PRs which looks wort exploring again:

https://github.com/elm-lang/core/pull/396
https://github.com/elm-lang/core/pull/399

Now when we have tests working we can prof their correctness easily.

**I'm 100% for choosing alternative fix when we came to conclusion it's better than mine or closing this in case new implementation is ready to ship. However I think it's good idea to make decision which is the best for users of core library.**

evancz:
The latest version of core includes @Skinney's array implementation, so this should be resolved by the rewrite.

PR #769, opened 2016-12-02
Remove duplicate case in a switch statement

tjikini:


evancz:
Weird! Seems reasonable to me. Thanks!

PR #770, opened 2016-12-03
Improve stack traces for port errors

rtfeldman:
Suppose we have a bad port invocation that's going to blow up, in a file called `index.html`:

```js
function invokeBustedPort() {
  app.ports.foo.send(null);
}

// alternatively, for a post-init error: setTimeout(invokeBustedPort, 500);
invokeBustedPort();
```

## Before
The stack trace doesn't include the offending function, `invokeBustedPort`, or a useful line number indicating where the function was invoked:
screen shot 2016-12-03 at 12 24 01 am

## After
Now it includes both!
screen shot 2016-12-03 at 12 30 38 am

## What Changed

The basic idea is to do the decoder check immediately inside `send`, and `throw new Error` inside `send` so it preserves the best stack trace information.

This means that in the `preInit` case it fails slightly sooner, because it does the check before init - and then defers the actual value sending - instead of deferring both.

### DIFF
--- a/src/Native/Platform.js
+++ b/src/Native/Platform.js
@@ -513,12 +513,6 @@ function setupIncomingPort(name, callback)
 
 	function postInitSend(incomingValue)
 	{
-		var result = A2(_elm_lang$core$Json_Decode$decodeValue, converter, incomingValue);
-		if (result.ctor === 'Err')
-		{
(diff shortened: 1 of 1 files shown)


turboMaCk:
This is great! I was dealing with this week ago. This improvement will make it much easier. :+1:

evancz:
Makes sense to me, nice improvement!

PR #771, opened 2016-12-03
Upgrade elm-test & test suite

turboMaCk:
## Why?

I think it's important to have working tests. Otherwise it's really hard to tell if any PR actually fixes problem and doesn't broke anything else.

## How

I was trying to reproduce [some issues with array](https://github.com/elm-lang/core/pull/765) and made [repository with test](https://github.com/turboMaCk/elm-array-tests). I think it make sense to give it back to community so we all can benefit from having working tests for core library

## Choices Made

I'm using (local) npm installation of [elm-test](https://www.npmjs.com/package/elm-test) by @rtfeldman which is de facto community standard. I'm also using `npm test` script as alias to local installation. On the other hand I don't want to make any unnecessary change to current work flow. This is why main script for running tests is almost the same as before (and using `npm test` internally).

Elm-test version is updated. One other change is that I've moved `tests/Test.elm` => `tests/Main.elm` which is more standard way to do it.

## Fixing Tests

There are some breaking changes in elm-test. This means tests (files in `tests/Tests`) should be updated one by one:

- [x] Array.tests
- [x] Basics.tests
- [x] Bitwise.tests
- [x] Char.tests
- [x] CodeGen.tests
- [x] Dict.tests
- [x] Equality.tests
- [x] Json.tests
- [x] List.tests
- [x] Result.tests
- [x] Set.tests
- [x] String.tests
- [x] Regex.tests

**This PR do not and will not include any other changes beside what is necessary to make current tests work as they are.**


tests/Test/Json.elm
```diff
@@ -63,21 +63,21 @@ customTests =
       "I want to see this message!"
 
     myDecoder =
-      Json.customDecoder ("foo" := Json.string) (\_ -> Err customErrorMessage)
+      Json.field "foo" Json.string |> Json.andThen (\_ -> Json.fail customErrorMessage)
```
turboMaCk: testing `andThen` instead of legacy `customDecoder`.

tests/run-tests.sh
```diff
@@ -14,5 +14,6 @@ echo "Linking $CORE_PACKAGE_DIR to $CORE_GIT_DIR"
 rm -rf $CORE_PACKAGE_DIR
 ln -s $CORE_GIT_DIR $CORE_PACKAGE_DIR
 
-elm-make --yes --output test.js Test.elm
-node test.js
+elm-make --yes --output test.js Main.elm
+
+npm test
```
turboMaCk: I think that this depend caching (elm-make do not overwrite & recompile anything but will try though). That doesn't feels right:-/ Please pay attention to this line & integration with `elm-test`

tests/Test/Array.elm
```diff
@@ -7,84 +7,115 @@ import List exposing ((::))
...
+                ]
+    in
+        describe "Array"
+            [ creationTests
+            , basicsTests
+            , getAndSetTests
+            , takingArraysApartTests
+            , mappingAndFoldingTests
+            , nativeTests
+            ]
```
turboMaCk: This is the only file formatted by elm-format. Please let me know if you want to keep old formatting.

.travis.yml
```diff
@@ -3,6 +3,7 @@ language: node_js
 node_js:
   - "4.3"
 install:
-  - npm install -g elm@0.17
+  - npm install -g elm@0.18
+  - npm install
```
rtfeldman: 🤔 what is the `npm install` part here doing?

rtfeldman: Oh I see - it's because of `"test": "node_modules/.bin/elm-test"`

tests/Test/Array.elm
rtfeldman: Nope, looks great!

turboMaCk:
@rtfeldman No problem at all. I did it this way only because it makes `elm-test` version independent on global installation. If it's preferable to use global install I don't see any blocker for switching to that.

turboMaCk:
@rtfeldman done in https://github.com/elm-lang/core/pull/771/commits/2942be8a2665e2a56567512237bc43f7404c3e58 🚀

tests/Test/Array.elm
turboMaCk: I would like to format all tests then. I think it's good time to do so since tests ware broken anyway so there is really low possibility it will make conflicts with any opened PR. What do you think?

turboMaCk: * I'm speaking **only about tests**. Doing same thing in src can cause nasty conflicts I don't want to be associated with :D

mgold:
Looks great to me!

rtfeldman:
Thank you so much @turboMaCk!

tests/Test/Array.elm
rtfeldman: @turboMaCk If you'd like to make a separate PR that runs `elm-format` on all tests, I'd be happy to accept it!

turboMaCk: @rtfeldman Not super important but here we go. https://github.com/elm-lang/core/pull/774.

PR #780, opened 2016-12-10
Add native Array.toIndexedList implementation.

efarrer:
Resolve a TODO for Array.toIndexedList by creating a native implementation of the function.

No change in behavior.

### DIFF
--- a/src/Native/Array.js
+++ b/src/Native/Array.js
@@ -315,6 +315,26 @@ function toList_(list, a)
 	return list;
 }
 
+function toIndexedList(a)
+{
+	return toIndexedList_(_elm_lang$core$Native_List.Nil, a);
+}
+
+function toIndexedList_(list, a)
+{
+	for (var i = a.table.length - 1; i >= 0; i--)
+	{
+		list =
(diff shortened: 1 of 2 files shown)


mgold:
This is available in [Array.Hamt](http://package.elm-lang.org/packages/Skinney/elm-array-exploration/1.0.0/Array-Hamt#toIndexedList)

src/Native/Array.js
```diff
@@ -315,6 +315,26 @@ function toList_(list, a)
...
 }
 
+function toIndexedList(a)
+{
+	return toIndexedList_(_elm_lang$core$Native_List.Nil, a);
+}
+
+function toIndexedList_(list, a)
+{
+	for (var i = a.table.length - 1; i >= 0; i--)
```
Pilatch: `!== -1` is slightly faster than `>= 0`

Pilatch: ... and `~i` may be faster still.

evancz:
Rather not add stuff to `Array` since the existing implementation is going to be phased out.

PR #784, opened 2016-12-15
Adding Unit Tests for Maybe

FRosner:
### Problem

There are no unit tests for the public functions of the `Maybe` module. This makes it impossible to apply refactoring to the existing code without knowing whether this will break the expected behaviour.

Also, unit tests are good examples on how to use an API. All public functions should have [unit tests](https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#bug-fixes).

### Solution

Add unit tests.


rtfeldman:
Thanks @FRosner!

PR #786, opened 2016-12-18
WIP: Fuzz tests for arithmetic

rtfeldman:
This is a good motivating use case for https://github.com/elm-community/elm-test/pull/93 because at present these fuzz tests are easily susceptible to floating point math errors.

(Marking this as WIP until `Expect.within` ships and we switch this implementation from `Expect.equal` to that.)


evancz:
Is work still progressing? Can this be closed?

PR #788, opened 2016-12-19
Maybe.concat and Maybe.concatMap

FRosner:
## Problem

Elm does not have a generic implementation of higher order collection functions like `map`, `filter`, `flatten`/`concat`, etc. While this is not a problem per se it is useful to have the individual APIs available in a consistent way. This way you can generalize the API in your head when using it.

While, e.g., `List.concat` and `List.concatMap` exists, there is no `Maybe.concat` and `Maybe.concatMap` available.

## Solution

Implement `Maybe.concat` and `Maybe.concatMap` in order to have a consistent API of both functions between `List` and `Maybe`. 

## Note

I read that you are strict about extending the core API and new functions should be proposed to `*-extra` first. While this makes sense for contributions that introduce additional concepts, this contribution adds already existing functionality to a place where it is missing. I think that having a consistent collection API is a strong feature and useful to have in the core API.

Anyway, ❤️  Elm.

### DIFF
--- a/src/Maybe.elm
+++ b/src/Maybe.elm
@@ -3,6 +3,7 @@ module Maybe exposing
   , andThen
   , map, map2, map3, map4, map5
   , withDefault
+  , concat, concatMap
   )
 
 {-| This library fills a bunch of important niches in Elm. A `Maybe` can help
@@ -14,8 +15,11 @@ you with optional arguments, error handling, and records with optional fields.
 # Common Helpers
 @docs withDefault, map, map2, map3, map4, map5
 
-# Chaining Maybes
-@docs andThen
+# Special Maps
+@docs concatMap
+
+# Combining Maybes
+@docs andThen, concat
 -}
 
 {-| Represent values that may or may not exist. It can be useful if you have a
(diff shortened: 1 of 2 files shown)


lukewestby:
Some quick thoughts I have:
`concatMap` and `andThen` are the same function, and `concat` can be derived from it as `andThen identity`. So the value here would be mostly in the names of the functions for consistency across List and Maybe rather than in previously unavailable functionality.

In any case, @FRosner, these kind of changes need to be brought up on the mailing list. Can you bring this up on [elm-dev](https://groups.google.com/forum/#!forum/elm-dev) and get some feedback from other folks?

FRosner:
You are totally right @lukewestby ¬_¬

So I guess I'll have to bring this up to the mailing list in terms of the consistency discussions (`flatMap` being called `concatMap` in `List` and `andThen` in `Maybe`).

Thanks for the comment.

richardhaven:
Emerson

As these methods do different things, rather than the same things on different objects, it is almost confusing that they have the same names.

PR #798, opened 2016-12-30
Remove unused imports

gabrielelana:


mgold:
👎  I'm pretty sure that core follows different rules for imports than other Elm programs, so these imports are actually necessary.

nicolaiskogheim:
@mgold But you're not sure?

mgold:
No, I'm not positive.

jvoigtlaender:
Actually, @mgold is right that default import rules do not apply to the core library. At the same time, this does not matter for this PR, since the imports removed are truly unused in the respective files.

evancz:
There is no harm in having extra imports. In the development build, only the top-level declarations that are needed are included, so it does not matter about `imports` for that. It can lengthen compile time, but the dev build also caches all packages, so they are only built once. That means there is no perf difference ever. So I think it is easier to just keep things as is. It's not hurting anything, and it'll take longer to rollback if something is off.

Issue #805, opened 2017-01-13
Text.fromString references missing module


joshsz:
https://github.com/elm-lang/core/blob/master/src/Basics.elm#L472

This document references `Text.fromString`, but the `Text` module is missing and there doesn't seem to be any implementation of `fromString` anymore in core.

I'm not sure where to direct readers to find the functionality in `fromString`, else I'd submit a PR.

joshsz:
Is `Text.fromString` no longer planned to be supported?

mgold:
The `Text` module and type are long gone.

joshsz:
I meant the ability to convert a string into an Elm symbol. 

It would reduce the boilerplate I need in this routing function example:

```
locationUpdate : Navigation.Location -> App.Msg
locationUpdate location =
  case (String.dropLeft 2 location.hash) of

    "ViewDocList" -> ViewState ViewDocList

    "ViewDocument" -> ViewState ViewDocument

    _ -> App.NoOp

    -- Would rather code this as:
    -- msg -> case Text.fromString msg of
    --          Just value -> ViewState value
    --          Nothing    -> App.NoOp
```

Sure I could just pass the strings along into the `ViewState` constructor for the `App.Msg` but somehow that seems contrary to the point of strongly-typed code.

mgold:
There are ways of doing what you want and you should ask on the [mailing list](https://groups.google.com/forum/#!forum/elm-discuss). What you're asking about is not what `Text.fromString` used to do.

jvoigtlaender:
> The `Text` module and type are long gone.

Actually, they are moved: http://package.elm-lang.org/packages/evancz/elm-graphics/1.0.1/Text#fromString

joshsz:
`Text.fromString` in `evancz/elm-graphics` has different functionality (it converts `String -> Text` when I'm looking for `String -> a`)

I'll ask on the mailing list, thanks!

jvoigtlaender:
I know that it has different functionality. I provided the pointer to it so that you can know what functionality it has (in contrast to what you apparently believed its functionality to be - and you saying you don't know where to find it).

[...]

PR #809, opened 2017-01-15
Fix typo in Basics.elm

Shoowa:
The document mention the use of a 'Text.fromString' function, when it actually discussed converting a data type to a String and then returning that result.


jvoigtlaender:
No, changing that to Text.toString would also be wrong. See https://github.com/elm-lang/core/issues/805.

PR #817, opened 2017-01-20
Improve String.toFloat and String.toInt

evancz:
This should allow each parser to accept a **broader range of inputs** and parse them **more quickly**.

For example, `toFloat` can now handle scientific notation like `1e40`, and `toInt` can handle hexadecimal like `0xBEEF`.

This change relies on the fact that `+str` and `Number(str)` do the same thing, and that `Number` is more restrictive in parsing than `parseInt` and `parseFloat`. This is based on information from [this](http://stackoverflow.com/questions/12227594/which-is-better-numberx-or-parsefloatx) and [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number).

I believe it will be faster, but I have not confirmed this.

## Remaining Tasks

  - Confirm that this works with strictly larger range of numbers.
  - Actually race the new and old implementations against each other to compare speed.
  - Attempt to make it crash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 elm-stuff
 tests/test.js
+node_modules/
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -225,7 +225,7 @@ function toInt(s)
 
 function intErr(s)
 {
-	return _elm_lang$core$Result$Err('could not convert string "' + s + '" to an Int');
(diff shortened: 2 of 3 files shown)


evancz:
Just a heads up, I'm working on the perf part / fixing some errors.

evancz:
Okay, this is superseded by https://github.com/elm-lang/core/pull/818

PR #818, opened 2017-01-20
Improve String.toInt and String.toFloat

evancz:
With some parser work I’m doing, I wanted to expand the set of numbers that Elm can easily turn into `Int` and `Float` values.

## Changes

  - `String.toFloat` accepts scientific notation like `1e+40`
  - `String.toInt` accepts hex like `0xBEEF`
  - Added a bunch of tests for these things

### DIFF
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -190,7 +190,7 @@ function endsWith(sub, str)
 function indexes(sub, str)
 {
 	var subLen = sub.length;
-	
+
 	if (subLen < 1)
 	{
 		return _elm_lang$core$Native_List.Nil;
@@ -203,8 +203,8 @@ function indexes(sub, str)
 	{
 		is.push(i);
 		i = i + subLen;
-	}	
-	
+	}
+
 	return _elm_lang$core$Native_List.fromArray(is);
 }
 
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -208,69 +208,77 @@ function indexes(sub, str)
 	return _elm_lang$core$Native_List.fromArray(is);
 }
 
+
 function toInt(s)
 {
 	var len = s.length;
+
+	// if empty
 	if (len === 0)
 	{
-		return _elm_lang$core$Result$Err("could not convert string '" + s + "' to an Int" );
+		return intErr(s);
 	}
-	var start = 0;
-	if (s[0] === '-')
+
+	// if hex
+	var c = s[0];
+	if (c === '0')
 	{
-		if (len === 1)
+		// must be hex
+		if (s[1] !== 'x')
 		{
-			return _elm_lang$core$Result$Err("could not convert string '" + s + "' to an Int" );
+			return intErr(s);
 		}
-		start = 1;
+
+		for (var i = 2; i < len; ++i)
+		{
+			var c = s[i];
+			if (('0' <= c && c <= '9') || ('A' <= c && c <= 'F') || ('a' <= c && c <= 'f'))
+			{
+				continue;
+			}
+			return intErr(s);
+		}
+		return _elm_lang$core$Result$Ok(parseInt(s, 16));
 	}
-	for (var i = start; i < len; ++i)
+
+	// is decimal
+	if (c > '9' || (c < '0' && c !== '-' && c !== '+'))
+	{
+		return intErr(s);
+	}
+	for (var i = 1; i < len; ++i)
 	{
 		var c = s[i];
 		if (c < '0' || '9' < c)
 		{
-			return _elm_lang$core$Result$Err("could not convert string '" + s + "' to an Int" );
+			return intErr(s);
 		}
 	}
+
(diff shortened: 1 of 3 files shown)


src/Native/String.js
```diff
@@ -203,75 +203,83 @@ function indexes(sub, str)
...
-	var start = 0;
-	if (s[0] === '-')
+
+	// if hex
+	var c = s[0];
+	if (c === '0')
 	{
-		if (len === 1)
+		// must be hex
+		if (s[1] !== 'x')
```
brian-carroll: If this is a number of minutes that I've extracted from a time-formatted string, like "09", it will fail. Is that desired behaviour?

mbylstra:
Any chance you could also accept binary literals? eg `0b1010101` or `0B1010101`. These are really helpful if you have to deal with parsing or generating binary file formats. Eg: If I just want the last 5 bits of a byte:

```
Bitwise.and myByte (String.toInt "0b11111")
```

With hex it's 
```
Bitwise.and myByte 0x1F
```
which is far less readable. Parsing binary formats is really hard - this makes it a little easier!

tests/Test/String.elm
```diff
@@ -40,5 +41,70 @@ tests =
...
+                , goodFloat "-3.14" -3.14
+                , goodFloat "0.12" 0.12
+                , goodFloat ".12" 0.12
+                , goodFloat "1e-42" 1e-42
+                , goodFloat "6.022e23" 6.022e23
+                , goodFloat "6.022E23" 6.022e23
+                , goodFloat "6.022e+23" 6.022e23
+                , badFloat "6.022e"
+                , badFloat "6.022n"
+                , badFloat "6.022.31"
```
eeue56: this is a good place for having fuzz tests, e.g

```elm

fuzz int "All ints are correctly converted to a string and back" (\int ->
  Expect.equal (Ok int) (String.toInt (toString int))

```

src/Native/String.js
```diff
@@ -203,75 +203,83 @@ function indexes(sub, str)
...
-			{
-				continue;
-			}
-		}
-		return _elm_lang$core$Result$Err("could not convert string '" + s + "' to a Float" );
+		return floatErr(s);
 	}
-	return _elm_lang$core$Result$Ok(parseFloat(s));
+	var n = +s;
+	return n === n ? _elm_lang$core$Result$Ok(n) : floatErr(s);
```
eeue56: :thought_balloon: Why use this over `isNaN`? is it meant to be faster? If so, a comment might be worthwhile here since it's not obvious

src/Native/String.js
```diff
@@ -203,75 +203,83 @@ function indexes(sub, str)
...
-	{
-		if (len === 1)
-		{
-			return _elm_lang$core$Result$Err("could not convert string '" + s + "' to a Float" );
-		}
-		start = 1;
-	}
-	var dotCount = 0;
-	for (var i = start; i < len; ++i)
+	if (s.length === 0 || /[\sxbo]/.test(s))
```
eeue56: :art: might be worth having a comment stating what the regex test does here (as it's not obvious unless you think about it/are familiar with non-decimal number strings)

justinmimbs:
The `toInt` function could be simplified with regexes. For example, the version below performs the same checks (but allows strings with leading zeros), and is a bit faster in some [brief timing tests](https://jsfiddle.net/2nzaLo6f/).

I think @brian-carroll brings up a good point. I would want `"09"` to parse without error. :thumbsup:

```js
function toInt(s)
{
  // decimal
  if (/^[+-]?\d+$/.test(s))
  {
    return _elm_lang$core$Result$Ok(parseInt(s, 10));
  }

  // hex
  if (/^0x[\dA-Fa-f]+$/.test(s))
  {
    return _elm_lang$core$Result$Ok(parseInt(s, 16));
  }

  return intErr(s);
}
```

mbylstra:
Something to keep in mind with parsing numbers like `09` is that leading zero is often used as syntax for Octal representations. I'm not saying `09` _should_ be parsed as an Octal number, it's just something to consider.

src/Native/String.js
crazymykl: Also, many languages (e. g. JavaScript, Ruby) allow capital letters for indicating base, `0XF` and similar.

evancz:
The alternate `toInt` implementation is not faster by my tests. The results reverse by [swapping the order of the tests](https://jsfiddle.net/2nzaLo6f/1/), so I suspect this has more to do with "VM warm up" than the actual details.

About parsing `0b1111`, the code you write does not type check. `toInt` produces a `Result`. It'd be better to just allow binary digit literals, but that's a separate discussion that I do not want to have right now.

About leading zeros, I only know of JS making that octal, and it's getting phased out. I would prefer not to allow "too many" zeros at the front, but to be backwards compatible and release this as a core patch, I need to keep allowing it for now.

I've responded to other things with some code changes.

evancz:
Thanks for the reviews everyone!

PR #820, opened 2017-01-23
Fix #819 

### LINKED ISSUE #819: 5.1.0 Ports variable 'result' not defined/in scope?
Upgraded to 5.1.0, and now on any port send I'm getting the following error:
`ReferenceError: Can't find variable: result`

looks like there is no `result` variable in scope at the line `var value = result._0;` in the following compiled JS:

```
function setupIncomingPort(name, callback)
	{
		var sentBeforeInit = [];
		var subs = _elm_lang$core$Native_List.Nil;
		var converter = effectManagers[name].converter;
		var currentOnEffects = preInitOnEffects;
		var currentSend = preInitSend;

		// CREATE MANAGER

		var init = _elm_lang$core$Native_Scheduler.succeed(null);

		function preInitOnEffects(router, subList, state)
		{
			var postInitResult = postInitOnEffects(router, subList, state);

			for(var i = 0; i < sentBeforeInit.length; i++)
			{
				postInitSend(sentBeforeInit[i]);
			}

			sentBeforeInit = null; // to release objects held in queue
			currentSend = postInitSend;
			currentOnEffects = postInitOnEffects;
			return postInitResult;
		}

		function postInitOnEffects(router, subList, state)
		{
			subs = subList;
			return init;
		}

		function onEffects(router, subList, state)
		{
			return currentOnEffects(router, subList, state);
		}

		effectManagers[name].init = init;
		effectManagers[name].onEffects = F3(onEffects);

		// PUBLIC API

		function preInitSend(value)
		{
			sentBeforeInit.push(value);
		}

		function postInitSend(incomingValue)
		{
			var value = result._0;
			var temp = subs;
			while (temp.ctor !== '[]')
			{
				callback(temp._0(value));
				temp = temp._1;
			}
		}
```

lukewestby:
If the result value is extracted and guarded by a throw in `send` then the value should be passed directly to `postInitSend`.

I think it would be good to set up some fixtures that can run in a separate node.js test suite to verify that ports are working okay to avoid regressions like this in the future, but that's a discussion for the ML.

#819

### DIFF
--- a/src/Native/Platform.js
+++ b/src/Native/Platform.js
@@ -511,9 +511,8 @@ function setupIncomingPort(name, callback)
 		sentBeforeInit.push(value);
 	}
 
-	function postInitSend(incomingValue)
+	function postInitSend(value)
 	{
-		var value = result._0;
 		var temp = subs;
 		while (temp.ctor !== '[]')
 		{
@@ -530,7 +529,7 @@ function setupIncomingPort(name, callback)
 			throw new Error('Trying to send an unexpected type of value through port `' + name + '`:\n' + result._0);
 		}
 
-		currentSend(incomingValue);
+		currentSend(result._0);
 	}
 
 	return { send: send };


evancz:
Thanks for the quick fix @lukewestby! This is released as 5.1.1 now. Sorry to folk who ran into this!

lukewestby:
👍 any time! `setupIncomingPort()` is near to my heart

PR #825, opened 2017-01-28
[failing] add tests for equality of Json.Encode.Value

soenkehahn:
Equality (elm's `==`) is not commutative for `Json.Encode.Value`s created with `object`. I believe this is a bug. This PR adds failing test cases but not a fix.

### DIFF
--- a/tests/Test/Json.elm
+++ b/tests/Test/Json.elm
@@ -3,16 +3,21 @@ module Test.Json exposing (tests)
 import Basics exposing (..)
 import Result exposing (..)
 import Json.Decode as Json
+import Json.Encode as Encode
 import String
 import Test exposing (..)
 import Expect
+import List
 
 
 tests : Test
 tests =
-    describe "Json decode"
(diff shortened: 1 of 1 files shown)


evancz:
To restrict equality to only work on sensible values, it requires changes in the compiler. Those changes are pretty involved, and the problem is known. Overall, I do not think having a failing test will bring us closer to the fix, and it will make other things harder in the meantime.

soenkehahn:
Sorry, this was not meant to be merged as is, only to provide a very concrete bug report. (Therefore it's completely fair to close this PR.)

> Those changes are pretty involved, and the problem is known.

Is this problem tracked somewhere? Could you point me to it, please?

evancz:
There's not a public issue. The path is not "4 clear steps" that just have to be implemented by anyone, so I don't think having it tracked on GitHub really helps fix things. Design choices need to be made, making those choices needs to be high priority, etc.

Issue #831, opened 2017-02-02
String.toInt returns a Float in certain circumstances


ChrisWellsWood:
I was doing some validation on input, and noticed an odd situation when parsing strings:

```Elm
> String.toInt "1"
Ok 1 : Result.Result String Int
> String.toInt "dfsf"
Err "could not convert string 'dfsf' to an Int" : Result.Result String Int
> String.toInt "-"
Ok NaN : Result.Result String Int
> String.toInt "+"
Ok NaN : Result.Result String Int
```
But the type of `NaN` is `Float`, not `Int`. I've not found any other circumstances where this occurs yet. I have a work around where I catch it manually, but I have to use `toFloat` on it first in order to use `isNaN`.

I'm using Elm 0.18, on FireFox and Elm repl. Thanks!

jvoigtlaender:
There do exist other operations in current Elm that produce `NaN` despite being typed `Int`. See https://github.com/elm-lang/core/issues/590, covered in https://github.com/elm-lang/core/issues/721.

Also, it would be pretty important that you not only specify which version of Elm you use, but also which version of the `core` library you use. Because the implementation of `String.toInt` was recently changed, see https://github.com/elm-lang/core/releases/tag/5.1.0.

mgold:
This issue reproduces in 0.18.0/5.1.1.

ChrisWellsWood:
Yeah, I'm using Elm 0.18 and elm-core 5.1.1.

evancz:
Fixed by OP in #834

PR #834, opened 2017-02-03
Added extra condition to stop '-' or '+' returning a `Float` from `String.toInt`.

ChrisWellsWood:
## Current Issue
Currently `String.toInt` can return a `Float`, as the JavaScript function `parseInt` will return `NaN` (which is type `Float` in Elm) if you supply it with '+' or '-'. This will get around the type checker, so the compiler will think that the 'NaN' is an `Int`. See #831 for more details of the bug.

## Solution
I've added a new condition so that an error will be raised if the input string is a single '-' or '+' character.

If you'd like to check the logic, I've mocked it up on [repl.it](https://repl.it/F5ol/0).

### DIFF
--- a/src/Native/String.js
+++ b/src/Native/String.js
@@ -236,7 +236,7 @@ function toInt(s)
 	}
 
 	// is decimal
-	if (c > '9' || (c < '0' && c !== '-' && c !== '+'))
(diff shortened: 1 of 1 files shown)


evancz:
Nice find! That boolean expression is getting out of control, but I'm not really sure how to simplify it in a nice way. Anyway, thanks for the fix!

PR #835, opened 2017-02-06
Broken test with runtime error for equality of Values

ArcherWheeler:
The added test fails with a runtime exception in the compiled Javascript.

### DIFF
--- a/tests/Test/Json.elm
+++ b/tests/Test/Json.elm
@@ -3,6 +3,7 @@ module Test.Json exposing (tests)
 import Basics exposing (..)
 import Result exposing (..)
 import Json.Decode as Json
+import Json.Encode exposing (object, null)
 import String
 import Test exposing (..)
 import Expect
@@ -13,6 +14,7 @@ tests =
     describe "Json decode"
         [ intTests
         , customTests
+        , equalityTests
         ]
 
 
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Isn't this essentially a duplicate of https://github.com/elm-lang/core/pull/825?

ArcherWheeler:
Not exactly, but it's similar. #825 is a failing test on equality of values. For this, the the test results in a runtime exception if the Value encodes a nested object.

jvoigtlaender:
Ah, right. I'll delete the noise from this issue here. Maybe you want to edit your original comment to point out that the Travis CI success masks the runtime error.

evancz:
I do not think it makes sense to add failing tests in hopes of fixing the underlying issue. It will require some pretty involved design and compiler changes, and in the meantime, we will have to struggle through constantly failing tests. So the timeline is not right to merge this in.

soenkehahn:
@evancz: Sorry, we should have mentioned this, but this PR was never meant to be merged as is, but just as a bug report. We thought it'd be easiest to understand what's going wrong by providing a test-case.

(I've worked with @ArcherWheeler on this.)

evancz:
Equality is "undefined" on JSON values. The fix will _probably_ be disallowing things like functions and JSON values with something like `comparable`, but there is not a certain design or timeline on anything like that.

It is an issue with a known workaround (do not use equality like that) and a relatively small portion of users will ever run into it at all (maybe <5%?), so it is not top priority. We know about it though, and a coherent fix is quite complicated, so in the same way there is no public issue for "make the parser faster", there is not a public issue for this large project.

soenkehahn:
Interesting, thanks for the explanation. I knew that functions are not `comparable`, but I'm very surprised that `Json.Encode.Value` is not, since it's just data. Also for `Json.Encode.Value` comparisons don't '[...] crash as quickly as possible' (quoted from the documentation you linked to). So for example `object [] == object []` doesn't crash. If it would crash it'd be easier to figure out that `Json.Encode.Value` isn't `comparable`.

I'm not working in elm currently, so for me it's not urgent to do anything about this, for what it's worth.

Issue #833, opened 2017-02-03
Better documentation for Date.fromString


jinjor:
The document for [Date.fromString](http://package.elm-lang.org/packages/elm-lang/core/latest/Date#fromString) is just saying

```
Attempt to read a date from a string.
```

, but does not describe what kind of string is valid.
Giving some examples or saying it is just the same as `new Date()` may be more helpful.

evancz:
It makes sense to either go with ISO 8601 and rename this to `fromIso8601` or to do something more serious to solve this in a better way.

PR #836, opened 2017-02-07
Add specification for valid date string

albronca:
reference #833

It's probably not worth it to have various examples, as the date string can vary vastly, but the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) is pretty clear about what's accepted.


evancz:
The plan is to do a more comprehensive fix that makes it work nicer.

PR #838, opened 2017-02-11
Add tests for Set

j-panasiuk:
Unit tests for each function exposed by Set module
If you feel there is too much redundancy in the tests, please let me know and I will trim them a bit.
(on the other hand I feel that many of these should be fuzz tests, but I didn't see fuzz testing used anywhere else in core, so decided to stick with unit testing for now)


rtfeldman:
Lovely, thank you @j-panasiuk!

Issue #721, opened 2016-09-22
Ideas for how Int and Float should work differently


evancz:
The goal of this issue is to organize all of the math related issues in one place and categorize them so something coherent can be done.

### API Suggestions

  - #238 - Add equivalent of `Number.EPSILON` to `Basics`
  - #190 - Change `toFloat` to `number -> Float`
  - #4 - change `(%)` to act on `number` types instead of `Int`

### Observations that math is weird because of JavaScript

  - #590 - integer division by zero
  - #92 - integer division that overflows 32-bits
  - #194 - exponentiation with negative integers

PR #842, opened 2017-03-07
Fixes Basics.truncate for values larger than 32 bit ints

Grannath:
The current native JavaScript implementation uses the `n | 0` bit-wise operation to truncate the value. This is restricted by the definition of bit-wise operations. They are defined to operate on 32 bit signed integers.

In general, JavaScript allows 64 bit floating point numbers. With these, integers can be represented up to `2^53 - 1` aka [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER). By using a bit-wise operation, the usable range of Int in Elm is restricted artificially.

The implementation is replaced here with `Math.ceil` and `Math.floor` depending on the sign. Testing in a REPL now successfully evaluates e.g. `truncate <| 2.45 * (2 ^ 33)`, which would have been impossible before.

Such a restriction is a) unnecessary and b) unexpected. I highly suggest removing it.

### DIFF
--- a/src/Native/Basics.js
+++ b/src/Native/Basics.js
@@ -73,7 +73,7 @@ function isInfinite(n)
 
 function truncate(n)
 {
-	return n | 0;
+	return n < 0 ? Math.ceil(n) : Math.floor(n);
 }
 
 function degrees(d)


witoldsz:
@evancz could you consider merging it "as is"? Today I had a nasty bug because of the `return n | 0` implementation of `truncate` (the doc says nothing about cutting the number to 32 bits). The #721 umbrella is a nice idea, but this single bug could be fixed just now, what would you say?

harfangk:
I ran into this issue myself today. I hope this would get merged soon!
By the way, there's also `Math.trunc()` function in ES5, which corresponds to this function, but of course it's supported on relatively newer browsers. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc

PR #844, opened 2017-03-15
Dict.first, Dict.last, Set.first, Set.last

billstclair:
I have an application that needs to process the elements of a Set one at a time. Currently, the only way to get an element from a set is to convert it to a list and take an element of the list. It's quicker and conses less to just get the first or last element out of the internal storage, and use that.

Since Set is based on Dict, I added first and last to both.

My current code:

```
popSet : Set comparable -> Maybe (comparable, Set comparable)
popSet set =
    case Set.toList set of
        [] ->
            Nothing
        res :: _ ->
            Just (res, Set.remove res set)
```

With these changes:

```
popSet : Set comparable -> Maybe (comparable, Set comparable)
popSet set =
    case Set.first set of
        Nothing ->
            Nothing
        Just res ->
            Just (res, Set.remove res set)
```

### DIFF
--- a/src/Dict.elm
+++ b/src/Dict.elm
@@ -1,7 +1,7 @@
 module Dict exposing
   ( Dict
   , empty, singleton, insert, update
-  , isEmpty, get, remove, member, size
+  , isEmpty, get, remove, member, size, first, last
   , filter
   , partition
   , foldl, foldr, map
@@ -23,7 +23,7 @@ Insert, remove, and query operations all take *O(log n)* time.
 @docs empty, singleton, insert, update, remove
 
 # Query
-@docs isEmpty, member, get, size
+@docs isEmpty, member, get, size, first, last
 
 # Lists
 @docs keys, values, toList, fromList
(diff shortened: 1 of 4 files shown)


Skinney:
You'll have more luck trying to add this to `elm-community/dict-extra`.

billstclair:
@Skinney I currently use in my code an implementation based on Set.toList and Dict.toList. The purpose of this patch is to make that faster with no memory allocation by using the internals of the opaque Set and Dict types. That can't be done in `elm-community/dict-extra`.

myrho:
+1

The `elm-community/intdict` also has this already. It's called [findMin](http://package.elm-lang.org/packages/elm-community/intdict/2.0.0/IntDict#findMin) and findMax there.

evancz:
At some point we will do a review of the `*-extras` and bring things in. These seem like good candidates because of their asymptotics, but there are other `Dict` changes that mean it will be easier to just do the changes by hand, rather than through PR.

PR #847, opened 2017-03-24
Add the compose function to the List package.

tibastral:
The compose function is useful when we want to apply a chain of functions to a single argument.

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -6,7 +6,7 @@ module List exposing
   , map, map2, map3, map4, map5
   , filterMap, concatMap, indexedMap
   , foldr, foldl
-  , sum, product, maximum, minimum, all, any, scanl
+  , sum, product, compose, maximum, minimum, all, any, scanl
   , sort, sortBy, sortWith
   )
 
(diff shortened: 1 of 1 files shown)


mgold:
This should start in the [list-extra](https://github.com/elm-community/list-extra) package.

jvoigtlaender:
Or in a function-extra package. Even in core, I doubt that List would be the right module for this function.

tibastral:
Thank you for your feedback. I keep using this function everywhere, simple, generic and I would say it also can give ideas to people, and is kind of a "functional programming only" concept, that's why I think it could be a good function to have here.

jvoigtlaender:
But have you read the following paragraph?

https://github.com/elm-lang/core/blob/master/CONTRIBUTING.md#adding-new-functions

It clearly indicates that you should close your pull request here and instead target it at another package. It even points you to `elm-community`.

And as said above, having this function be called `List.compose` seems wrong. It is not about composing list(s).

evancz:
Better in `*-extra` and we can bring in things later in a coherent way.

PR #849, opened 2017-04-01
Json.Decode.keyValuePairs need not reverse order

zwilias:
Reversing the entries does not break any guarantees: an object is an unordered collection of key-value pairs as per the spec. 

However, browsers generally respect order of definition when dealing with objects, and reverting that order when converting to an ordered data-structure violates the principle of least surprise.


evancz:
I was going through this library the other day and wasn't sure if I should put things in order. Based on seeing this PR and #863, I figured I'd go for it.

PR #853, opened 2017-04-12
Default imports are slightly wrong in the README file

norpan:
Module `List` does not expose the `List` type, it's a built-in type.
Module `Platform.Cmd` is imported `as Cmd`.
Module `Platform.Sub` is imported `as Sub`.


evancz:
It should be updated for what will be true in `6.0.0`

Issue #799, opened 2016-12-30
Port interactions do not preserve the order of Cmd.batch


lguminski:
I have noticed that Elm 0.18 does not respect the order of port interactions defined by `Cmd.batch`. This is a problem for me as I am writing an Elm web client which receives data with a series of events (log file), parses it, and dispatches a series of event to a Javascript module through Elm ports. `Cmd.batch` like this

```elm
  {} ! [ portA 5, portA 4, portB 3, portB 2, portA 1 ]
```

in Javascript console produces
```
Event [1] received through portA
Event [4] received through portA
Event [5] received through portA
Event [2] received through portB
Event [3] received through portB
```
As you can see first Javascript subscription handlers of Port A have been called, and then Port B, which broke the original ordering `portA 5, portA 4, portB 3, portB 2, portA 1 `. **Note**: the test has been started after all subscriptions are in place (see `elm.ports.triggerTest.send(null);` line in `index.html` below). So the order of subscribing should not affect the observed behavior.

**How to reproduce**
All files are in https://github.com/lguminski/ElmOrderingPortEventsTest repository. Clone them
```
git clone https://github.com/lguminski/ElmOrderingPortEventsTest
cd ElmOrderingPortEventsTest
make
```
and open http://localhost:8000 in your browser and check logs in Javascript console.

**Files**

OrderingPortEventsTest.elm
```elm
port module Main exposing (..)
import Html exposing (..)
import Html.Attributes exposing (..)

type alias Model =
  {}

type Msg
  = NoOp
  | Test ()


main =
  Html.program
    { init = init
    , update = update
    , view = view
    , subscriptions = subscriptions
    }

view : Model -> Html Msg
view model =
  div [] []

subscriptions : Model -> Sub Msg
subscriptions model =
  Sub.batch [ triggerTest Test ]

port portA : Int -> Cmd msg
port portB : Int -> Cmd msg
port triggerTest : (() -> msg) -> Sub msg

init : ( Model, Cmd Msg )
init =
  {} ! []

update : Msg -> Model -> ( Model, Cmd Msg )
update msg model =
  case msg of

    NoOp ->
      model ! []

    Test _ ->
      model  !  [ portA 5, portA 4, portB 3, portB 2, portA 1 ]
```

index.html
```html
<!DOCTYPE html>
<html lang="en">
    <body>
        <script src="elm.js"></script>
        <script>
            const elm = Elm.Main.fullscreen();
            elm.ports.portA.subscribe((number) => {
[...]

Issue #730, opened 2016-10-13
Ports are not published in the specified order


DavidDTA:
When using a `Cmd.batch` to send data over two ports, the ports receive an event in the same order regardless of the order in which they appear in the batch.

Working example:
https://daviddta.github.io/elm-port-order-bug/

Code:
https://github.com/DavidDTA/elm-port-order-bug

Looking at the console, we see that port one always receives the event before port two.

DavidDTA:
In answer to the robot's questions, I'm using Chrome and Elm 0.17.1.

jvoigtlaender:
What makes you think this is a bug? Is there any documentation out there that claims any guarantees about the order of stuff happening that is put into `Cmd.batch`? (Hint: I don't think there is. And there should not be.)

If there is nothing like that to be found, then this is not a bug, but instead some feature you want (order preservation on stuff put into `Cmd.batch`). In that case, the issue tracker is not the appropriate venue for such a feature request. Such discussions happen on Slack or on the mailing list.

http://elm-lang.org/community

DavidDTA:
I tried posting to the elm discuss mailing list following your suggesion, but it has not gone through yet (pending approval).

jvoigtlaender:
Okay, I'm sure the message will get through on the mailing list, so am closing the issue here.

Meanwhile, maybe this helps you as well: http://faq.elm-community.org/17.html#what-is-the-difference-between-cmd-and-task

Note that it talks about multiset in relation to Cmd. Multisets have no order.

DavidDTA:
A few questions about the linked article:

1. The article claims that `Cmd` is not monadic, but I believe that the existence of `Cmd.Extra.message` and `Cmd.map` indicates that it is.
1. The article mentions that APIs generally expose `Task` in favor of `Cmd`.  Why is the port API `a -> Cmd msg` instead of  `a -> Task Never ()` or something like that?
1. Is there a recommended way to pass data to ports in order?  I've come up with the workaround of sending over only one port per update and using `Cmd.Extra.message` to trigger additional updates immediately, but I don't think it's very clean.

jvoigtlaender:
All these are good questions. For the mailing list.

[...]

PR #856, opened 2017-04-24
Document `Cmd.batch`

fosskers:
From #799 and #730 , it has been noticed by multiple users that the execution order of a `List Cmd` folded by `Cmd.batch` is not always the same as the order they were defined in the List. This might come from an unconscious assumption that `Cmd.batch` works like [`sequence :: [IO a] -> IO [a]`](https://www.stackage.org/haddock/lts-8.12/base-4.9.1.0/Prelude.html#v:sequence)  instead of like its async cousin `Task`.

This PR documents the `Cmd.batch` function to help avoid confusion for future users.

### DIFF
--- a/src/Platform/Cmd.elm
+++ b/src/Platform/Cmd.elm
@@ -48,7 +48,26 @@ map =
   Native.Platform.map
 
 
-{-|-}
+{-| Fold many `Cmd` together. Useful for combining things like HTTP
+requests:
+
+    import Http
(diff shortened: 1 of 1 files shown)


evancz:
This is documented in the `dev` branch. See https://github.com/elm-lang/core/blob/dev/src/Platform/Cmd.elm for details.

Thank you for the PR!

PR #861, opened 2017-05-04
Remove duplicate function definition

Janiczek:
This eg. crashes `prepack` tool from working on Elm-compiled JS files.

(There is another `_Json_badIndex` two definitions above the deleted one.)


evancz:
This is resolved in the `dev` branch.

PR #863, opened 2017-05-06
Documentation: Specifies that Json.Decode.keyValuePairs garantees no ordering.

astlouisf:
I was surprised today to find that `Json.Decode.keyValuePairs` doesn't enforce any order to elements of the resulting list. I wrongly assumed that it would considering `Dict.toList` does.

A simple notice clears any confusion.


evancz:
I just did a change in d9d2c18b9e274ea4680134da6f09de7c2ec4048d so that order is kept from the JSON.

PR #864, opened 2017-05-08
remove duplicate of badIndex

JorisM:
badIndex is defined twice


dawehner:
Note: This PR enables elm to be used with prepack.

evancz:
This got fixed in the `dev` branch already. Thanks for the PR nonetheless!

PR #865, opened 2017-05-12
Dev: Fixes runtime crash when slicing a large array so it becomes small (< 32 elements)

Skinney:
Includes test to reproduce issue.

The first problem was that depth calculation when slicing down to a one element array (slice 0 1 array) would return `NaN`, which is a setup for failure.

The second problem was the assumption that `hoistTree` always would be called with a tree with a size of > 1.

### DIFF
--- a/src/Array.elm
+++ b/src/Array.elm
@@ -754,6 +754,7 @@ sliceRight end ((Array length startShift tree tail) as arr) =
 
             depth =
                 (endIdx - 1)
+                    |> max 1
                     |> toFloat
                     |> logBase (toFloat branchFactor)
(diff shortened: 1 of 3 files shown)


Skinney:
Hmm. That second commit was supposed to be a different PR. It's a harmless change though.

It doubles performance of `_JsArray_push` in Chrome and Safari (no change in Firefox). A similar change of the `_JsArray_set` function also had some performance increase in Chrome and Safari, but had the opposite effect (way worse performance) in Firefox.

evancz:
Thanks! :)

PR #871, opened 2017-06-10
Fix example in filterMap documentation.

soveran:


jvoigtlaender:
Why is that a "fix"?

Someone of age 19 is a teenager. It's nine*teen*.

soveran:
I think you are absolutely right. I was reading the Elm Guide and its section on [Maybe](https://guide.elm-lang.org/error_handling/maybe.html). In the paragraph about "Partial Functions", the `getTeenAge` function uses `if 13 <= age && age <= 18`, then refers to [List.filterMap](http://package.elm-lang.org/packages/elm-lang/core/latest/List#filterMap) for further reading. As I followed the link, I encountered this mismatching example and I thought it was an error. Others my also find it confusing, but I think you are correct and we can close this issue now :-)

jvoigtlaender:
Might want to open at https://github.com/evancz/guide.elm-lang.org/issues, then.

PR #872, opened 2017-06-12
Improve performance of List.foldr

Skinney:
This PR takes the same idea presented in #707 and applies it to `List.foldr`. I see performance improvements of 30-40% on my machine (Safari and Chrome). Since `foldr` is used by many other list functions, this gives a nice performance boost across the board.

I tested each level of "unrollment" individually. 9 seemed to be the magic number.

I tried the same thing with `foldl`. It gave a modest improvement in Chrome (8-10%), but reduced performance in Safari (10-14%).

I have not tested if #707 still provides a performance improvement over a map fn using this foldr implementation.

### DIFF
--- a/src/Elm/Kernel/List.js
+++ b/src/Elm/Kernel/List.js
@@ -35,17 +35,6 @@ function _List_toArray(xs)
 	return out;
 }
 
-var _List_foldr = F3(function(f, b, xs)
-{
-	var arr = _List_toArray(xs);
-	var acc = b;
-	for (var i = arr.length; i--; )
-	{
-		acc = A2(f, arr[i], acc);
-	}
-	return acc;
-});
-
 var _List_map2 = F3(function(f, xs, ys)
 {
 	var arr = [];
--- a/src/List.elm
+++ b/src/List.elm
@@ -160,8 +160,96 @@ foldl func acc list =
     foldr (+) 0 [1,2,3] == 6
 -}
 foldr : (a -> b -> b) -> b -> List a -> b
-foldr =
-  Elm.Kernel.List.foldr
+foldr fn acc ls =
+    foldrHelper fn acc 0 ls
+
+
+foldrHelper : (a -> b -> b) -> b -> Int -> List a -> b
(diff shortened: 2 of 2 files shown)


evancz:
Two questions:

  - Instead of having branches for every single scenario, what if you only had 0, 1, and 9? Or 0, 1, 2, 4, 8? Or some other pattern for reducing the number of branches? How much does that reduce the size of generated code? How much perf does it cost?

  - Can you use `f a (f b (f c ...))` instead of using `|>`? No reason to require a compiler optimization to trigger if we do not need to.

Skinney:
* Won't compile without branches for every single scenario.
* Sure
* Not only did nested case-statements reduce generated code size with about 550 bytes (2500bytes -> 1950bytes). It also increased performance. We're now in the 40-45% range of performance improvements (Chrome & Safari). Good call :)

Skinney:
Another interesting thing, for future reference, is that I consistently get 50+% performance improvement when inlining the `ctr > 500` line in the compiled code. By that I mean changing the compiled code from `...Util.cmp(ctr, 500) > 0` -> `ctr > 500`.

evancz:
Neat! I want to do some thinking about whether this has implications on how pattern matching code is generated.

About it not compiling, you need to change it from `a :: b :: []` to `a :: b :: _` and do a recursive call.

Can you share the code that is doing the performance comparison? I would like to run it myself and see the numbers. With that ability, I can take a look at this on my own in the next few days.

And yeah, this is one of the first optimizations we can do if type information is kept all the way until code generation. Swapping in `===` and `<` for `String`, `Int`, etc. will surely make things faster.

Skinney:
Sure: https://github.com/Skinney/elm-list-exploration

evancz:
I merged it in, but it did not register here for some reason.

I ended up reducing the unrolling to 4 to make things a bit smaller without huge perf losses. Not sure if it's the right trade, but it seemed like a reasonable compromise.

Anyway, thanks for looking into this!

PR #876, opened 2017-06-20
Eslint/fix Native/Utils.js

owanturist:
Hello! This PR includes changes of code for passing of eslint checking. The PR based on #875 and can be merged only after that.

**Before**: 157 problems (132 errors, 25 warnings)
**After**: 22 problems (0 errors, 22 warnings)

_I have another branches with changes for every file with native code and if the PR will merged or if the idea with eslint integration will approved I'll continue creation of PRs._


src/Native/Utils.js
```diff
@@ -1,7 +1,12 @@
-//import //
-
-var _elm_lang$core$Native_Utils = function() {
-
+/* global
+	_elm_lang$core$Native_Array
+	_elm_lang$core$Array$toList
+	_elm_lang$core$Dict$toList
+	_elm_lang$core$Set$toList
+*/
```
owanturist: It's more legal and pure way for indicate of globals

evancz:
I don't think I want to do this. I'm okay with the process, and I sometimes do odd things for performance or asset size that I would not want to be "against the rules"

PR #878, opened 2017-06-21
Improved information about format used for function fromString in Date.elm

Bernardoow:
Improved information about format used for function fromString in Date.elm


evancz:
I am planning to do a more comprehensive fix. I wrote some code to verify that it is an ISO 8601 string, but it may make sense to go much further than that.

PR #881, opened 2017-07-02
`sequence` is not stack-safe

zwilias:
This probably should not come up in any realistic scenario, but then again, it's just one less thing that can go wrong.


evancz:
Having the new `foldr` implementation be entirely in Elm makes me more comfortable with this change. Thanks!

zwilias:
Absolutely! The more of core we can have in Elm, the more comfortable I
feel with this type of change :)

PR #883, opened 2017-07-08
Remove |> and <| from the Random module.

mgold:
As requested. Also includes some very minor revisions to docs and removes the `mul32` function that was never called.

I copied the file to a separate directory and it compiled, but extra eyes to make sure I didn't do anything stupid would be great.

And thank you so much for merging the previous PR!


evancz:
Looks good to me! Thank you!

Also, of course, very glad to have the better implementation :) Just needed to be going through a bunch of core stuff to do the review!

PR #887, opened 2017-07-19
Make it possible for elm-test to use core's Random

mgold:
This adds `Random.independentSeed`, which allows the creation of random
seeds that are completely independent of each other. No amount of taking
random numbers from one seed will "wind" it to the same place as
another, such that they produce the same values.

This feature is essential to elm-test's ability to reproduce failures
because it allows each fuzz test to consume as many random values as it
needs without affecting tests that run after it. It means we can run
failing tests alone without disturbing the failure. It also makes it
possible to parallelize test runs.

If this PR is merged, elm-test will be able to use core's Random
implementation instead of my third party library. cc @rtfeldman

### DIFF
--- a/src/Random.elm
+++ b/src/Random.elm
@@ -6,7 +6,7 @@ effect module Random where { command = MyCmd } exposing
   , andThen
   , minInt, maxInt
   , generate
-  , step, initialSeed
+  , step, initialSeed, independentSeed
   )
 
 {-| This library helps you generate pseudo-random values.
@@ -51,7 +51,7 @@ it separate feels like a nice balance for now:
 @docs maxInt, minInt
 
 # Generate Values Manually
-@docs Seed, step, initialSeed
+@docs Seed, step, initialSeed, independentSeed
 
 -}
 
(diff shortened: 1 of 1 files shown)


src/Random.elm
```diff
@@ -522,13 +525,45 @@ you use a seed, you'll get another one back.
...
+The independent seeds are extremely likely to be distinct for all practical
+purposes. However, it is not proven that there are no pathological cases.
+-}
+independentSeed : Generator Seed
+independentSeed =
+    Generator <|
+        \seed0 ->
+            let
+                gen =
+                    int 0 0xFFFFFFFF
```
basile-henry: As far as I know `int` uses two's complement signed integers. Did you mean for the range to go from `0` down to `minInt`? Otherwise, if you meant to generate any 32-bit integer, how about `int minInt maxInt`?

mgold: `Int` is a JavaScript number and the fact that it is a whole number is enforced by the type system, not the representation. If you examine a list of numbers made from this generator, the range is exactly as expected: zero to 2^32 - 1. Note that when a value is passed to `initialSeed`, it undergoes an *unsigned* rightshift which forces it to be positive. But, it appears no such check happens for the increment, so it can be negative. Which, according to the bounds [here](https://en.wikipedia.org/wiki/Linear_congruential_generator), is wrong. So you found a bug!

mgold:
Force-pushed a commit that resolves both of @basile-henry's comments.

evancz:
Is this on top of the increased randomness and addition of `constant`?

@mgold, maybe you can explain to me in Slack what needs to be done. I do not recall what this is all about.

PR #901, opened 2017-08-23
Improve String.toInt docs: "+" and "-" return Ok NaN

fdbeirao:
This caught me by surprise today, so I hope this PR helps future me.


evancz:
That's a bug, not a feature! I forget if it is fixed in the upcoming release, but I'd like to fix this in 0.19

zwilias:
Fixed in #834 indeed, @fdbeirao

PR #905, opened 2017-09-02
Fix equality check of oneOf decoder

### LINKED ISSUE #904: Equality check on Json.Decode.oneOf always returns true
Currently any pair of `oneOf` decoders are equal in the equality check that is [used for diffing in virtual-dom](https://github.com/elm-lang/virtual-dom/blob/temp-3.0.0/src/Elm/Kernel/VirtualDom.js#L604). Because of this, changes in `oneOf` decoder is not reflected in event handler. ([An example on Ellie](https://ellie-app.com/4cb2MQ2GnGXa1/5))

Equality of two `oneOf` decoders are checked with `_Json_listEquality()`, which expects JavaScript arrays of decoders. However, a `oneOf` decoder has an Elm `List` of decoders instead of a JavaScript `Array`. Because of this, the `len` variable below becomes `undefined` given `oneOf` decoders, and the function always returns `true`.

https://github.com/elm-lang/core/blob/bc067c8e3db20235fc275b1d825bb76f365a1509/src/Elm/Kernel/Json.js#L378-L398

shuhei:
Fixes #904 

- Renaming the existing `_Json_listEquality` as `_Json_arrayEquality` because it actually works on JavaScript arrays instead of `List`s
- Adding a new `_Json_listEquality` function that works on Elm `List`s

### DIFF
--- a/src/Elm/Kernel/Json.js
+++ b/src/Elm/Kernel/Json.js
@@ -370,7 +370,7 @@ function _Json_equality(x, y)
 			return x.index === y.index && _Json_equality(x.decoder, y.decoder);
 
 		case __1_MAP:
-			return x.func === y.func && _Json_listEquality(x.decoders, y.decoders);
+			return x.func === y.func && _Json_arrayEquality(x.decoders, y.decoders);
 
 		case __1_AND_THEN:
 			return x.callback === y.callback && _Json_equality(x.decoder, y.decoder);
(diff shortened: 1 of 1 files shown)


rupertlssmith:
This is a good patch and fixes a subtle bug that people could spend hours figuring out.

This patch is old enough that in the mean time JSON stuff got moved to a separate package from core. Would you be able to open a new PR against `elm/json` for this?

https://github.com/elm/json/issues/15

shuhei:
Hi, thanks for looking into this old PR! I'm on vacation right now and also have been out of touch with Elm for a while. If you need this fix soon, please feel free to create a PR to the elm/json repo.

rupertlssmith:
If you are out of touch with Elm, then I will create the PR in elm/json - your fix applies just the same in the new package.

PR #907, opened 2017-09-04
wrap Dict.fromList into a Result (in the documentation)

bacey:
decodeString returns a Result, so Dict.fromList needs to be wrapped into a Result too


evancz:
Thank you for pointing this out! Some modules got moved around, so I made the changes in the appropriate places.

PR #908, opened 2017-09-15
Add Tuple.mapBoth function

minond:
Helper function when you need to apply a function to both elements in a Tuple. Like a combination of `mapFirst` and `mapSecond`:

```elm
mapBoth ((+) 1) (6, 8) == (7, 9)
```

### DIFF
--- a/src/Tuple.elm
+++ b/src/Tuple.elm
@@ -1,6 +1,6 @@
 module Tuple exposing
   ( first, second
-  , mapFirst, mapSecond
+  , mapFirst, mapSecond, mapBoth
   )
 
 {-| Some helpers for working with 2-tuples.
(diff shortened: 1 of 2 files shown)


minond:
Ummm... I think I'm missing something. Build is failing with an `elm-make: elm-package.json: openBinaryFile: does not exist (No such file or directory)` error, which I'm getting locally as well. I'm looking around for some technical documentation on contributing to `core` but haven't been able to find it or anything that is helpful with dealing with this error. Could someone point me in the right direction?

jvoigtlaender:
Maybe the most important information to you, wanting to contribute to `core`, is this document: https://github.com/elm-lang/core/blob/master/.github/CONTRIBUTING.md

Note in particular what it says in the "Adding New Functions" section.

minond:
@jvoigtlaender thanks for sending me that. What's documented there totally makes sense. If this seems like it belongs in an *-extra package instead, then I can move it.

jvoigtlaender:
Contributing it to https://github.com/elm-community/basics-extra would make sense. That package already has a section of tuple functions (http://package.elm-lang.org/packages/elm-community/basics-extra/2.1.2/Basics-Extra#=%3E).

evancz:
This is already in the `dev` branch I believe!

PR #910, opened 2017-09-26
Replace andThen with flatMap.

Josef-Vonasek:
I find the name `flatMap` very natural fit for the signature `(a -> (Box a)) -> Box a -> Box b`. After all the first thing that comes onto my mind, when I see **flatMap** is **flatten** and **map**.

```elm
flatten : Box (Box a) -> Box a
map : (a -> b) -> Box a -> Box b
```

It is then not hard to figure out that it should probably behave like 
```elm
flatMap function box == flatten (map function box)
```

`andThen` on the other hand feels very opaque and / or vague (e.g. `|>` is very often called andThen) and without documentation I would have no idea would does it do.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -105,7 +105,7 @@ import Signal ( Signal )
         withDefault : a -> Maybe a -> a
         oneOf : List (Maybe a) -> Maybe a
         map : (a -> b) -> Maybe a -> Maybe b
-        andThen : Maybe a -> (a -> Maybe b) -> Maybe b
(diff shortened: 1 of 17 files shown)


Josef-Vonasek:
The tests are currently failing, but my time is spare. 

Is this pull request acceptable? If **yes**, I will fix the code, so that all test will pass. If **not**, I am just going to close this pull request.

evancz:
I considered this name alongside other options when I chose `andThen`. I do not think it makes sense to make this change.

PR #915, opened 2017-10-21
Documentation update for Random (map3)

dstreet26:
The example shows map3 being used to combine 3 generators to produce random colors.

[Link to Doc](http://package.elm-lang.org/packages/elm-lang/core/latest/Random#map3)

Here is more or less the error that someone would get if they tried to use the example as-is.

```
-- TYPE MISMATCH --------

Function `map` is expecting the 2nd argument to be:

    Generator Float

But it is:

    Generator Int
```

### DIFF
--- a/src/Random.elm
+++ b/src/Random.elm
@@ -297,7 +297,7 @@ map2 func (Generator genA) (Generator genB) =
 
     hsl : Generator Color.Color
     hsl =
-      map3 Color.hsl (map degrees (int 0 360)) (float 0 1) (float 0 1)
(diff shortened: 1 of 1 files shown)


evancz:
The examples got updated, so the new thing there is https://github.com/elm-lang/core/blob/master/src/Random.elm#L377-L403

Longer term, we'd like to add an `examples/` folder to all packages such that snippets can come with an Elm file that is compiled before publication. The idea would be that you can then refer to lines from those files.

PR #918, opened 2017-11-01
Move allocation out of _Utils_compare (improving performance)

Skinney:
In my work on a new `Dict` implementation, I discovered that one of the functions being called the most is `_Utils_compare`. By moving the object allocation out of the function (pre-allocating) we can improve performance of the function.

In my benchmark, `Dict.get` goes from a mean runtime of ~30ns per call (for a `Dict` of 100 key-value pairs) to ~26ns. If my math is correct (it rarely is) that is a 13% performance increase.

I also tried the following, but it did not offer better performance and is more code:

```
var lt = { $: 'LT' };
var eq = { $: 'EQ' };
var gt = { $: 'GT' };

var _Utils_compare = F2(function(x, y)
{
    var ord = _Utils_cmp(x, y);
    return ord < 0 ? lt : ord > 0 ? gt : eq;
});
```

(The actual benchmarks were performed by modifying a 0.18 compiled output by hand. Even though the code looks slightly different in this 0.19-branch, I don't see why the results would be any different).

### DIFF
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -152,11 +152,11 @@ var _Utils_le = F2(function(a, b) { return _Utils_cmp(a, b) !== _Utils_GT; });
 var _Utils_gt = F2(function(a, b) { return _Utils_cmp(a, b) === _Utils_GT; });
 var _Utils_ge = F2(function(a, b) { return _Utils_cmp(a, b) !== _Utils_LT; });
 
-var _Utils_ordTable = ['LT', 'EQ', 'GT'];
+var _Utils_ordTable = [{ $: 'LT' }, { $: 'EQ' }, { $: 'GT' }];
 
 var _Utils_compare = F2(function(x, y)
 {
-	return { $: _Utils_ordTable[_Utils_cmp(x, y) + 1] };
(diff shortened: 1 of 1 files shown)


justinmimbs:
Regarding the idea of using primitives at runtime to represent argument-less union types: I hand-modified the JS output of Robin's `Dict.LLRB.get` function to use `cmp` instead of `compare`, and it saw a 77% increase in ops/sec in Chrome (using dicts of 100 elements with ints as keys).

i.e. from

```js
var _p16 = A2(_elm_lang$core$Basics$compare, targetKey, _p15._1);
switch (_p16.ctor) {
    case 'LT':
```

to

```js
var _p16 = _elm_lang$core$Native_Utils.cmp(targetKey, _p15._1);
switch (_p16) {
    case -1:
```

evancz:
The final version looks like this: https://github.com/elm-lang/core/blob/master/src/Elm/Kernel/Utils.js#L127-L131

There are bunches of inlining tricks we can do to avoid allocations, but the compiler infrastructure is not sophisticated to do them right now. So it's not that I don't think of them. It's that they cannot be done without a great deal of infrastructure work to get type information on all expressions. That is something I have my eye on as well.

PR #920, opened 2017-11-19
Add more docs to `Time` module

dbrgn:
Elm beginner here.

Initially I did not really understand how `inSeconds` etc work, so I tried to add some docs.

I hope the changes are correct. I did not manage to generate a preview of the docs. This is what I tried:

```
$ cd path/to/core
$ elm-make --docs=docs.json
```

...but the `docs.json` file is empty, so it can't be fed to http://package.elm-lang.org/help/docs-preview.

### DIFF
--- a/src/Time.elm
+++ b/src/Time.elm
@@ -101,25 +101,25 @@ hour =
   60 * minute
 
 
-{-|-}
+{-| The number of milliseconds since January 1st, 1970. -}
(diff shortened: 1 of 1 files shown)


evancz:
This will be moving to https://github.com/elm-lang/time which does things in a much clearer way. It will be out with the next release of Elm, and I hope it clarifies things with a bunch of new docs in the README and such.

PR #925, opened 2017-12-06
New dict implementation

Skinney:
A new Dict implementation based on Left-Leaning Red-Black trees (LLRB). Performance numbers can be found here: https://discourse.elm-lang.org/t/new-dict-implementation-for-elm-now-done/174/3

The code that is commented out helps validating the invariants of a LLRB tree. It's not something that should be built in production builds, or even used by regular users, but it's invaluable when making changes to the data structure.

### DIFF
--- a/src/Basics.elm
+++ b/src/Basics.elm
@@ -614,6 +614,8 @@ e =
 
 
 {-| Convert radians to standard Elm angles (radians).
+
+    radians pi == 3.141592653589793
 -}
 radians : Float -> Float
 radians angleInRadians =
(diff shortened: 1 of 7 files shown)


Skinney:
Just wanted to mention @justinmimbs, who did some impressive work on this code.

evancz:
Do you mind getting the conflicts resolved? I think we should try to see the impact on code size in applications that use dictionaries!

evancz:
I think messy commits are kind of lame, so if it is easier, just close and open a new PR with just the changes in a commit or something. Or not. Whatever works!

PR #926, opened 2017-12-08
Add AnimationFrame package link and update docs for Time.every

rl-king:
Updated the docs for Time.every, referring/linking to the AnimationFrame package in the note. Someone might interpret this as `requestAnimationFrame` not being available in Elm without a port (https://github.com/furrary/react-fiber-demo/pull/3#issuecomment-347727613).

_note: I wasn't able to test this as elm make docs didn't seem to work._

### DIFF
--- a/src/Time.elm
+++ b/src/Time.elm
@@ -61,10 +61,12 @@ subscriptions work.
 
 [arch]: https://github.com/evancz/elm-architecture-tutorial/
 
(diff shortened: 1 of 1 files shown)


evancz:
It should be clarified now https://package.elm-lang.org/packages/elm/time/latest/Time#every though with a little mistake!

Issue #329, opened 2015-08-04
Bool values cannot be the member of Set


jinjor:
I'm facing a problem with making a Set that contains Bool values.
And also, lists or tuples that contains Bool cause the same trouble.

Here is the simple example and the error message.
```elm
set = Set.fromList [True]
```
```
The 1st argument to function `fromList` has an unexpected type.

4| set = Set.fromList [True]
Note: comparable types include Int, Float, Char, String, lists, and tuples.

As I infer the type of values flowing through your program, I see a conflict
between these two types:

    comparable

    Bool
```
I understand what the compiler is saying, but not sure why Set or Dict keys need to be `comparable`. Should they always be sorted?

I think the possible solutions are...
- Bool (or other union types) should be `comparable`
- there should be another type instead of `comparable`

Any ideas? Or is there any workaround?

Thanks.

jvoigtlaender:
`Set` is implemented via `Dict`, and `Dict` is implemented via some sort of balanced trees, operations on which do in fact require a total order on entries/keys. So the implementation indeed requires `comparable` for types that should be put into a `Set` or used as key in a `Dict`.

So probably the right feature request here is that `Bool` should become `comparable`. Making all union types comparable would be more complicated, but special-casing for `Bool` could make sense.

evancz:
Why do you want a set of booleans? Any case I can think of can also be handled just by booleans.

jvoigtlaender:
Well, what about a set of tuples, some tuple component of which is a boolean? Certainly that's a valid use case that is not covered by just booleans alone?

evancz:
Ah, I was just looking at the example given. Nested stuff makes sense. Let's hear @jinjor's specific case though.

jinjor:
Yeah, exactly the case @jvoigtlaender said. (Sorry, the example above is not real one. )

Here is the specific case in my WebRTC app.
```elm
type alias Peer = Int
type alias MediaType = String
type alias Upstream = Bool
type alias Connection = (Peer, MediaType, Upstream)

type alias Model =
  {  connections : Set Connection
  ,  videoUrls : Dict Connection String

  ...

  }
```
Oh, now I found `MediaType` can also be written as an union type, like `type MediaType = Audio | Video`.

evancz:
[...]

PR #929, opened 2017-12-20
Allow ADTs and records in Sets and as Dict keys

brianhempel:
It's a pain to serialize and deserialize ADTs and records to and from comparable types when you want to use them in a set. This patch allows ADTs and records to be stored in sets and used as dictionary keys. However, they are still _not_ `comparable`. The comparison operators `> < >= <=` continue to only operate on the previous `comparable` types.

This PR implements part (2) of [this proposal which observes that you might allow any type in sets without making all types `comparable`](https://github.com/elm-lang/elm-compiler/issues/774#issuecomment-274577632). Here's how it works:

1. The `Set` and `Dict` implementation need the key type to have an order, but the order itself is not important because order is not part of the mathematical contract for what a `Set` or `Dict` is. This patch augments the Elm internal `cmp` function to also assign an order to values of any non-function type. (ADT values are ordered first alphabetically by constructor name, then by each argument; for records, fields are compared in alphabetical order.)
2. As this ordering is somewhat arbitrary, ADTs and records are _not_ publicly considered `comparable`. The typechecker continues to ensure that the `> < >= <=` etc. functions are only applied to the normal `comparable` types.

This patch allows us to reduce the Sketch-n-Sketch codebase by [~225 hairy and opaque lines of serialization and deserialization](https://github.com/ravichugh/sketch-n-sketch/commit/0a7ad48222ef75523e845bb25d5191e68ccde956#diff-1695b0c613bbb07d5f6a446026de3e41).

As is, this PR is based off of Core 5.1.1 so we could use it in Sketch-n-Sketch right away. If the approach is acceptable I'm happy to rework it against the current `dev` branch.

### DIFF
--- a/src/Dict.elm
+++ b/src/Dict.elm
@@ -10,9 +10,8 @@ module Dict exposing
   , toList, fromList
   )
 
-{-| A dictionary mapping unique keys to values. The keys can be any comparable
-type. This includes `Int`, `Float`, `Time`, `Char`, `String`, and tuples or
-lists of comparable types.
+{-| A dictionary mapping unique keys to values. The keys can be any type
+other than functions.
 
 Insert, remove, and query operations all take *O(log n)* time.
 
@@ -40,6 +39,7 @@ Insert, remove, and query operations all take *O(log n)* time.
 import Basics exposing (..)
 import Maybe exposing (..)
 import List exposing (..)
+import Native.Basics
 import Native.Debug
 import String
 
@@ -98,14 +98,14 @@ dictionary.
     get "Spike" animals == Nothing
 
 -}
-get : comparable -> Dict comparable v -> Maybe v
+get : k -> Dict k v -> Maybe v
 get targetKey dict =
   case dict of
     RBEmpty_elm_builtin _ ->
       Nothing
 
     RBNode_elm_builtin _ key value left right ->
-      case compare targetKey key of
+      case keyCompare targetKey key of
         LT ->
           get targetKey left
 
@@ -117,7 +117,7 @@ get targetKey dict =
 
 
 {-| Determine if a key is in a dictionary. -}
-member : comparable -> Dict comparable v -> Bool
+member : k -> Dict k v -> Bool
 member key dict =
   case get key dict of
     Just _ ->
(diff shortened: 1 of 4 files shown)


evancz:
Based on reading the description and some links, it seems like this gives short-term benefits for your project, but it is not a coherent overall design that is well-integrated with the language.

My general policy is that "making changes" is not default good. It may be that everyone agrees that the current thing is bad. It may also be that some people think that a certain thing would be better, but also not what we'd want to do ultimately. I think it is best to err on the side of conservative changes in the language and core libraries, especially when work arounds exist.

So I don't think it makes sense to go for this now. I do think your idea of "internally comparable" vs comparable with normal functions is interesting. It is 100% observable via `Map.foldr` and `Map.foldl` though, so I don't know if that distinction is real in practice. (I.e. I can just define my own `lt` function using a `Dict` underneath.)

Do you mind writing up the particular thing you need to use as keys? No proposals, just a description of your situation. I think having real world data like this will be really valuable in finding a path that we'll be happy with in 20 years.

evancz:
As a side note, making relatively serious changes to the language and core libraries is always preceded by conversation. I know other communities do this "PR or GTFO" insanity, as if it is in any way reasonable to have people write tons of code before having any conversations about expectations, but we tend to avoid that process. I think "PR or GTFO" also leads to negative interactions where folks feel "I did all this work, and your _not_ merging it?!" because they see themselves as giving and the maintainer as being unappreciative. My analysis is that every "contribution" is better framed as a "collaboration" and you have to talk and know each others plans to collaborate effectively.

brianhempel:
In the SnS we allow the user to select various features in the output. We store the selected features in a set, as it doesn't make sense and could cause bugs if a feature was somehow double-selected. The hierarchy of selectable features is a fairly complex ADT which, previous to this PR, we were serializing and deserializing into strings. If we had anticipated how unwieldy our serialization/deserialization code would get, we would have simply used a plain list to store the selected features and used special functions on that list to treat it like a set. In other words, the best workaround for us would be to not use the built-in Set for our sets—somewhat unsatisfying.

Skimming all the requests for adding comparability to everything, most people don't really want comparability, they want to be able to put anything into a set (c.f. [1](https://github.com/elm-lang/elm-compiler/issues/774#issuecomment-53653149), [2](https://github.com/elm-lang/elm-compiler/issues/774#issuecomment-66926825), [3](https://github.com/elm-lang/elm-compiler/issues/774#issuecomment-143837907), [4](https://github.com/elm-lang/elm-compiler/issues/774#issuecomment-152291312), [5](https://github.com/elm-lang/elm-compiler/issues/774#issuecomment-252454653), [6](https://github.com/elm-lang/elm-compiler/issues/774#issuecomment-278622198), [7](https://groups.google.com/forum/#!topic/elm-discuss/CexFZT9Mqa8), [8](https://github.com/elm-lang/elm-compiler/issues/310#issue-21649726), [9](https://github.com/elm-lang/elm-compiler/issues/310#issuecomment-283075462), [10](https://github.com/elm-lang/core/issues/329#issuecomment-127793585), and [11-last message in thread](https://groups.google.com/forum/#!msg/elm-discuss/1M3uA5w6R5Q/NbzkHQ_LCwAJ)) which is why this PR eschews a broad public definition of comparability and opts for an internal one, much like the current implementation of `==`. If you're trying to avoid typeclasses or functor-modules (you may not be—I don't know) I don't see how else sets can be appropriately generalized.

PR #931, opened 2017-12-28
Document sin/cos/tan functions

dbrgn:
Previously it was not clear whether the argument should be passed in
degrees or radians.


evancz:
Decided to add much more elaborate comments. It contains the info you want along with some examples of how to use them. Thanks for the push to add these!

PR #937, opened 2018-01-15
Maybe: Add eliminator for Maybe.

MostAwesomeDude:
This is an essential helper.

### DIFF
--- a/src/Maybe.elm
+++ b/src/Maybe.elm
@@ -36,6 +36,14 @@ type Maybe a
     | Nothing
 
 
+{-| Eliminate an optional value. -}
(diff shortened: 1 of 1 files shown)


mkwiatkowski:
You can already find this helper in the exts package: http://package.elm-lang.org/packages/krisajenkins/elm-exts/27.4.0/Exts-Maybe#maybe Their implementation is a bit shorter: https://github.com/krisajenkins/elm-exts/blob/27.4.0/src/Exts/Maybe.elm#L75 Indeed would be nice to see this in core.

evancz:
The policy is to make API requests to the `maybe-extra` packages. There's probably something like this. If not, they will decide. That's the feeder for core.

PR #944, opened 2018-02-24
[0.19] Char.fromCode - typo

zwilias:
`c` isn't defined here


evancz:
Nice find! Thank you!

PR #948, opened 2018-03-07
New dict implementation

Skinney:
A new Dict implementation based on Left-Leaning Red-Black trees (LLRB). Performance numbers can be found here: https://discourse.elm-lang.org/t/new-dict-implementation-for-elm-now-done/174/3

The code that is commented out helps validating the invariants of a LLRB tree. It's not something that should be built in production builds, or even used by regular users, but it's invaluable when making changes to the data structure.

Also includes code from justinmimbs (fromSortedList and friends).

### DIFF
--- a/src/Dict.elm
+++ b/src/Dict.elm
@@ -50,7 +50,7 @@ Insert, remove, and query operations all take *O(log n)* time.
 
 # Transform
 
-@docs map, filter, foldl, foldr, partition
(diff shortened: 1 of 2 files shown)


Skinney:
New commit strips out the "sorted optimization" to save on compiled asset size.
The new Dict implementation is still faster than the current implementation, and without sorted opts it should also be smaller.

PR #959, opened 2018-05-14
New Dict implementation using Left-Leaning RB-Trees.

Skinney:
This new implementation is a drop-in replacement, offering faster inserts (~170% increase) and faster removals (~30% faster).

The code is also compiles to less js. Recent testing shows that this implementation produces 488 less bytes after minification and gzip.

I've made as few changes to the original code as possible.


evancz:
Just got a chance to do the review. Looks great to me! Excited to get it out to folks ❤️

Issue #942, opened 2018-02-13
Char.isUpper only handles ascii a-z, unlike Char.toUpper


drathier:
`Char.toUpper` handles non-ascii letters, such as `åäö` -> `ÅÄÖ`, but `Char.isUpper` only handles ascii a-z. I expect them to agree on what characters are upper case.

Same with `Char.toLower` and `Char.isLower`.

ufocoder:
@drathier BTW [documentation](http://package.elm-lang.org/packages/elm-lang/core/5.1.1/Char) said us that all `Classification` functions are working only with ASCII. So it's a normal function behavior

drathier:
The functions that convert to upper/lower case modify more than just ascii a-z. I'm fine with either, but the combination is troublesome. 

Personally, I'd like to drop these 4 functions because not all scripts have upper/lower case characters, so relying on them being different like in English is a problem. This is a request to fix an api inconsitency, not a bug report.

drathier:
Apparently which lowercase characters correspond to what uppercase characters is locale-dependent, so doing the non-ascii version properly will be hard. https://stackoverflow.com/questions/12537377/in-haskell-how-can-i-uppercase-a-unicode-character-with-respect-to-current-local

Also, ß is a single German lowercase character, but it maps to two uppercase characters.

A-z only (which breaks any non-english app with user input, like given names), or dropping support entirely seem to be the two options here.

PR #970, opened 2018-07-24
Make Char.isLower and Char.isUpper Unicode-aware

Janiczek:
This allows people with non-ASCII alphabets work with `Char.isLower` and `Char.isUpper`. Uses `toUpper` and `toLower` underneath, which use Javascript's `String.prototype.toLower/UpperCase()`.

The second condition in the functions is there to distinguish between characters that have an upper/lower-case pairing, and those that don't (`'0' == Char.toLower '0'` but we don't want `isLower '0'` to be true).

_EDIT: I can't update code of this PR anymore; there is #1138 with a fix for the `==` and `/=` import._

### DIFF
--- a/src/Char.elm
+++ b/src/Char.elm
@@ -66,46 +66,42 @@ type Char = Char -- NOTE: The compiler provides the real implementation.
 -- CLASSIFICATION
 
 
-{-| Detect upper case ASCII characters.
+{-| Detect upper case Unicode characters.
 
     isUpper 'A' == True
     isUpper 'B' == True
     ...
     isUpper 'Z' == True
+    
+    isUpper 'Σ' == True
 
     isUpper '0' == False
     isUpper 'a' == False
     isUpper '-' == False
-    isUpper 'Σ' == False
 -}
 isUpper : Char -> Bool
 isUpper char =
-  let
-    code =
-      toCode char
-  in
-    code <= 0x5A && 0x41 <= code
+  (char == Char.toUpper char)
+    && (char /= Char.toLower char)
 
 
-{-| Detect lower case ASCII characters.
+{-| Detect lower case Unicode characters.
 
     isLower 'a' == True
     isLower 'b' == True
     ...
     isLower 'z' == True
+    
+    isLower 'π' == True
 
     isLower '0' == False
     isLower 'A' == False
     isLower '-' == False
-    isLower 'π' == False
 -}
 isLower : Char -> Bool
 isLower char =
-  let
-    code =
-      toCode char
-  in
-    0x61 <= code && code <= 0x7A
+  (char == Char.toLower char)
+    && (char /= Char.toUpper char)
 
 
 {-| Detect upper case and lower case ASCII characters.


Janiczek:
Is related to #385.

drathier:
What's considered an uppercase character depends on your locale. This PR is still an improvement, though.

Related to #942.

evancz:
For future reference, the [`toLocaleUpperCase`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleUpperCase) function talks about cases where this will break:

> The `toLocaleUpperCase()` method returns the value of the string converted to upper case according to any locale-specific case mappings. `toLocaleUpperCase()` does not affect the value of the string itself. In most cases, this will produce the same result as `toUpperCase()`, but for some locales, such as Turkish, whose case mappings do not follow the default case mappings in Unicode, there may be a different result.

So it seems that `toUpperCase()` is a pure function, but `toLocaleUpperCase()` is not. My instinct is that the "correct" version of this function takes `Language` as an argument. (Not sure if the idea of "locale" is better. Maybe it is geographical? Maybe there is some standards body that defines locals?) 

Next step is to find nice links that describe:

1. How "upper case" is defined by unicode. Is there a big table somewhere?
2. How a "locale" is defined and who manages that. Are there "new locales" if human culture changes? Who captures that, and how do browsers know about it?

I would prefer to understand the problem more completely before changing things.

Janiczek:
From my cursory googling and research:


## 1. How is "upper case" defined?

I think [this FAQ](http://unicode.org/faq/casemap_charprop.html) is the link you want.

In short, yes, there is a big table. *Three*, in fact.

1. ftp://ftp.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt
2. ftp://ftp.unicode.org/Public/UCD/latest/ucd/SpecialCasing.txt
3. ftp://ftp.unicode.org/Public/UCD/latest/ucd/CaseFolding.txt (GitHub or Markdown doesn't support FTP links :man_shrugging: )

Here is the relevant [section of standard](http://www.unicode.org/versions/Unicode11.0.0/ch05.pdf#G21180).

It has some sense of inter-version stability between the Unicode versions.


## 2. How is "locale" defined?

Again, [an Unicode FAQ](http://cldr.unicode.org/locale_faq-html); and this time there's a whole [homepage](http://cldr.unicode.org/).

You can [download the current version](http://cldr.unicode.org/index/downloads), there are a lot of XML files inside with various data (casing of dates / languages / ..., etc.), to be interpreted according to [LDML](http://unicode.org/reports/tr35/).

They are also transformed from the XML into [JSON](https://github.com/unicode-cldr/cldr-json), which might be a better fit for Elm?

gormonn:
I ran into the same issue when doing the exercise in the forms section, namely checking the uppercase password.

At first glance, I thought this was a serious omission.

However, then I wondered if it was worth letting users set their passwords to Unicode.
On the one hand, this password be more resistant to brute force.
On the other hand, it may be inconvenient for the user if the device does not have a specific locale.

However, in any case, this is not decided at the stage of front-end approval, but much earlier. Thus, this limitation may cause frustration for developers from regions other than English. And negatively affects the use of Elm as the main front-end stack in the Enterprise environment. This means about the popularity and development of the language.

But I believe that such an annoying flaw will still not be a problem for Elm.

Meanwhile.
I’m very curious if I’m stuck in mind trap just because Elm doesn’t support my native language well. There may be other uses for Unicode and Char.isUpper that we are not aware of. So write if you know this.

avh4:
FYI, this package can currently be used to deal with unicode strings: https://package.elm-lang.org/packages/BrianHicks/elm-string-graphemes/latest/

sagehane:
https://github.com/elm/core/blob/e47edebc0240dc0b840875d7714f9393e0b3f45d/src/Char.elm#L84-L85

Wouldn't `char /= Char.toLower char` suffice? If a character does not equal its lowercase counterpart, it must be uppercase.

`A` would become `a`, `True`
`a` would become `a`, `False`
`0` would become `0`, `False`

Same applies to `isLower`.

(I tried using code comments, didn't work, idk why)

Edit: The one scenario where this might make a difference is if there's a "middle case" character that has both an upper and a lower case variant. But I don't think such a character exists, and even if it does, should `isUpper` return a `True` or `False` in such a scenario?

miniBill:
@sagehane 
```
> let s = Char.fromCode 453 in (s, Char.toUpper s, Char.toLower s)
('ǅ','Ǆ','ǆ') : ( Char, Char, Char )
```

sagehane:
@miniBill, good to know. So, would you argue that `'ǅ'.isLower()` should return `true`, or `false`? I feel like if a character has an uppercase form, it should be considered a lowercase character. That is, I disagree with the current code.

miniBill:
`ǅ` is neither lowercase nor uppercase, according to Unicode

https://ellie-app.com/gBmgVVFzhbRa1 this contains a table of all the 1441 codepoints that give wrong results with the current proposal

I personally think the proposal is the best compromise between accuracy and size/speed. Getting better results belongs in external packages (like elm-unicode)

rupertlssmith:
This PR causes a compile error when it is used: https://github.com/elm-janitor/apply-patches/issues/1

Janiczek:
@rupertlssmith I can't edit this PR's code anymore, see #1138 for compilable code.

PR #975, opened 2018-08-25
Fix optimized `compare` when first arg is Nil

brian-carroll:
Fix a "truthiness" bug in the PROD version of `_Utils_cmp`

[This line](https://github.com/elm/core/blob/master/src/Elm/Kernel/Utils.js#L104) wants to check if the first argument is a Tuple. In PROD mode it uses `if (!x.$)` to check for falsiness, since tuples have no `$` property in PROD mode.

Unfortunately `_List_Nil` can also be passed in here, which is represented as `{ $: 0 }`. When that happens, `x.$` _exists_, but since it is falsy it passes the condition and is misidentified as a tuple! The tuple-related code then runs even though the inputs are lists. This can produce incorrect results, such as in the SSCCE below.

This PR fixes it by using `hasOwnProperty` instead of a truthiness check.

(There are, of course, lots of ways to check for missing properties. Not sure what your preference is. I figured that checking the existence of a property should be more efficient than doing anything with its value. Dunno if that's right or not.)

```elm
module Main exposing (main)

import Browser
import Html exposing (Html, text)

main = Browser.sandbox
    { init = ()
    , view = view
    , update = \() () -> ()
    }

view : () -> Html msg
view () =
    if [] < [0] then
        text "Default compile shows this"
    else
        text "--optimize shows this"
```

### DIFF
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -101,7 +101,7 @@ function _Utils_cmp(x, y, ord)
 	//*/
 
 	/**__PROD/
-	if (!x.$)
+	if (!x.hasOwnProperty('$'))
 	//*/
 	/**__DEBUG/
 	if (x.$[0] === '#')
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -101,7 +101,7 @@ function _Utils_cmp(x, y, ord)
 	//*/
 
 	/**__PROD/
-	if (!x.hasOwnProperty('$'))
+	if (typeof x.$ === 'undefined'))
 	//*/
 	/**__DEBUG/
 	if (x.$[0] === '#')


evancz:
Can you benchmark it against something like `x.$ < 2` and `typeof x.$ == 'number'`?

brian-carroll:
Sure, I'll have a look at that today

brian-carroll:
OK I tried all 3 of those on latest Chrome, Firefox and Edge, on a Windows 10 machine.

`hasOwnProperty` turns out to be the worst on all browsers. So a good suggestion to benchmark it!
`typeof` and `<` are similar to each other on Chrome and Firefox (6x faster than `hasOwnProperty`)
On Edge, `typeof` is 5x faster than both of the others.

So `typeof` seems to be the one they're all optimizing the hardest. Updated the PR to use that instead.

https://jsperf.com/elm-19-compare

evancz:
Thanks! That matches my mental model of what would be faster. It is a shame to need all those extra bits, but I don't see another path really :P

PR #981, opened 2018-09-05
Keep relative ordering of dispatches in _Platform_dispatchEffects

### LINKED ISSUE #980: Timer subscription doesn't fire in response to URL change
While porting my app to 0.19, I found that my page loading bar timer was working only on initial load—after changing the "page", the timer subscription would not fire again.

# SSCCE

Here's the best I could do to recreate it in an SSCCE:

```elm
import Browser exposing (UrlRequest)
import Browser.Events
import Browser.Navigation as Nav
import Html exposing (a, div, h1, text)
import Html.Attributes exposing (href)
import Time
import Url exposing (Url)


main =
    Browser.application
        { init = init
        , view = view
        , update = update
        , subscriptions = subscriptions
        , onUrlRequest = ClickedLink
        , onUrlChange = UrlChanged
        }


type alias Model =
    { count : Int
    , navKey : Nav.Key
    , isRunning : Bool
    }


init : () -> Url -> Nav.Key -> ( Model, Cmd Msg )
init _ url key =
    ( loadPage url
        { count = 0
        , isRunning = False
        , navKey = key
        }
    , Cmd.none
    )


type Msg
    = Tick Time.Posix
    | ClickedLink UrlRequest
    | UrlChanged Url


update msg model =
    case msg of
        Tick _ ->
            ( { model | count = model.count + 1 }, Cmd.none )

        ClickedLink req ->
            case req of
                Browser.Internal url ->
                    ( model, Nav.pushUrl model.navKey (Url.toString url) )

                Browser.External href ->
                    ( model, Nav.load href )

        UrlChanged url ->
            ( loadPage url model, Cmd.none )


loadPage url model =
    { model | isRunning = url.fragment == Just "/timer" }


subscriptions : Model -> Sub Msg
subscriptions model =
    if model.isRunning then
        Browser.Events.onAnimationFrame Tick

    else
        Sub.none


view model =
    { title = ""
    , body =
        [ text (String.fromInt model.count)
        , if model.isRunning then
            div []
                [ h1 [] [ text "Running" ]
                , a [ href "/" ] [ text "stop" ]
                ]

          else
            div []
                [ h1 [] [ text "Idle" ]
                , a [ href "#/timer" ] [ text "start" ]
                ]
        ]
    }
```

I would have made it smaller, but the bug only appears to occur when the `subscription` function is called in response to an `update` triggered by `Browser.application`'s `onUrlChange` event.

## Expected Behavior

When you click the "start" link in the above example app, the counter should start increasing 60 times per second.

## Actual Behavior

After clicking "start", the counter is stuck at 0.

## Analysis

As best I can tell, the `subscription` is getting called at the right time, but the Elm runtime seems to be ignoring the subscription value returned by my code. I suspect the bug is similar to https://github.com/elm/compiler/issues/1776 which also reports subscriptions getting ignored.

# Additional Notes

Replacing `Browser.Events.onAnimationFrame Tick` with `Time.every 16 Tick` appears to fix the bug. Although in my real application which I first experienced the bug in, it doesn't make any difference whether I use the animation-frame timer vs `Time.every`—they both get stuck.

JaSpa:
While the for-in loop is executing, the call to `__Scheduler_rawSend` can
lead to a reentrant call to `_Platform_dispatchEffects`. Because of this the inner call dispatches some effects before the first call even finished dispatching all effects leading to messages delivered in the wrong order, e.g. dropping subscriptions.

This fixes #980 and elm/compiler#1776.

### DIFF
--- a/src/Elm/Kernel/Platform.js
+++ b/src/Elm/Kernel/Platform.js
@@ -208,20 +208,40 @@ var _Platform_map = F2(function(tagger, bag)
 
 // PIPE BAGS INTO EFFECT MANAGERS
 
+var _Platform_dispatchQueue = [];
+var _Platform_dispatching = false;
 
 function _Platform_dispatchEffects(managers, cmdBag, subBag)
 {
-	var effectsDict = {};
-	_Platform_gatherEffects(true, cmdBag, effectsDict, null);
-	_Platform_gatherEffects(false, subBag, effectsDict, null);
+  _Platform_dispatchQueue.push({
+    __managers: managers,
+    __cmds: cmdBag,
+    __subs: subBag
+  });
 
-	for (var home in managers)
-	{
-		__Scheduler_rawSend(managers[home], {
-			$: 'fx',
-			a: effectsDict[home] || { __cmds: __List_Nil, __subs: __List_Nil }
-		});
-	}
+  if (_Platform_dispatching)
+  {
+    return;
+  }
+  _Platform_dispatching = true;
+
+  var effect;
+  while (effect = _Platform_dispatchQueue.shift())
+  {
+    var effectsDict = {};
+    _Platform_gatherEffects(true, effect.__cmds, effectsDict, null);
+    _Platform_gatherEffects(false, effect.__subs, effectsDict, null);
+
+    for (var home in effect.__managers)
+    {
+      __Scheduler_rawSend(managers[home], {
+        $: 'fx',
+        a: effectsDict[home] || { __cmds: __List_Nil, __subs: __List_Nil }
+      });
+    }
+  }
+
+  _Platform_dispatching = false;
 }
 
 


JaSpa:
This is a potential fix for elm/compiler#1776 by which a lot of people seem to be affected. @evancz (or someone else from the core team, sorry I am not really sure who to ping) can you take a look?

evancz:
Thank you for looking into this! Robin got a chance to look into this for a Friday recently, and I think we get the root issue now. I believe there is a very subtle problem with the code in this PR though:

It refers to `managers[home]` but I think it needs to be `effects.__managers[home]` instead.

As is, you could have issues if someone has multiple Elm applications running on the same page. Maybe one application has some set of effects, but another has a different set. I _think_ the code in this PR could get confused between them.

The overall approach seems to be the best way to fix this though, so my attempt is in https://github.com/elm/core/commit/704dcc08d4617d45c8b0fce7a68729335c880756

PR #999, opened 2018-10-23
Updated Documention: Task examples for Browser.Dom.focus were pointing to the wrong location

wking-io:
The Task examples that used `Browser.Dom.focus` pointed to the wrong location. They also used the wrong Error type so I updated all of these to point in the correct location and use the correct modules, functions, and types.

### DIFF
--- a/src/Task.elm
+++ b/src/Task.elm
@@ -38,16 +38,16 @@ import Result exposing (Result(..))
 {-| Here are some common tasks:
 
 - [`Time.now : Task x Posix`][now]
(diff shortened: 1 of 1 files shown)


evancz:
It looks like this got fixed by https://github.com/elm/core/commit/e2b4e514cc3eab1bfeafa75b75acc60398076819 in response to https://github.com/elm/core/issues/997

So if it's not already good on the website, it will be when 1.0.3 comes out with some miscellaneous bug fixes soon.

PR #1000, opened 2018-10-24
Improve overall iteration performance

austinshenk:
Multiple modules in core loop through list, arrays, and objects in various ways. Some can't be improved like the while cons, but there are a number of them that could be restructured to gain quite a bit of performance for not much, or any, file size increase.

A performance test bench detailed [here](https://web.archive.org/web/20141205235948/https://blogs.oracle.com/greimer/entry/best_way_to_code_a) shows many cases where a decrementing while loop with a simple conditional is the fastest looping structure.

```
var i=arr.length; while (i--) {}
```

I know this can't always be achieved given the functionality of the loop, but in the worst case we can ensure small improvements. For example, I've found various cases where a for-loop's length could be cached so it is not having to be calculated every iteration. It sounds small, but it even has better performance.

I'm planning to implement and test many of the areas I've found and report on the results for performance and also verifying functionality was not altered.

### DIFF
--- a/src/Elm/Kernel/JsArray.js
+++ b/src/Elm/Kernel/JsArray.js
@@ -21,7 +21,7 @@ var _JsArray_initialize = F3(function(size, offset, func)
 {
     var result = new Array(size);
 
-    for (var i = 0; i < size; i++)
+    for (var i = 0; size - i; i++)
     {
         result[i] = func(offset + i);
     }
@@ -33,7 +33,7 @@ var _JsArray_initializeFromList = F2(function (max, ls)
 {
     var result = new Array(max);
 
-    for (var i = 0; i < max && ls.b; i++)
+    for (var i = 0; max - i && ls.b; i++)
     {
         result[i] = ls.a;
         ls = ls.b;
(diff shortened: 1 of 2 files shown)


austinshenk:
Did similar testing to `String.map` as I did with reverse. The following screen is the timing results of mapping through a 1 million character string.

image

Overall about a 2x increase in performance. I also verified its functionality was not altered.

austinshenk:
I tried out a variation of the decrementing while loop by instead making it an incrementing one with a similar conditional. Hadn't really thought of doing it this way before.

```
while (len-i) {}
```

This allows most functions to maintain their body functionality while also easily gaining the performance boost.

Skinney:
Nice! Which browsers have you tested this on?

austinshenk:
I've been testing this just on Chrome. I want to setup some test cases on an online performance tester so I can test it in multiple browsers and multiple platforms.

evancz:
It is really risky to change a bunch of things like this, so I think the next step should be:

1. Benchmark individual functions
2. Get results in many browsers
3. Post results on https://discourse.elm-lang.org/

If the results show good results on certain specific functions, they may make sense as _individual_ pull requests. That way they can be considered in isolation from other changes which may not have as good results, or may be different for some other reason. One aspect to consider is that JS VMs change over time, so it may be reasonable to just leave things as is knowing that whatever moment we optimize things, it may be different down the line.

Anyway, in the meantime, I think it makes sense to close here and try to do those three steps.

PR #1006, opened 2018-11-24
Improve List.sortBy performance

go1t:
Instead of calling the transform function on each comparison, we can call it once before the sorting to reduce the number of calls.


go1t:
that's a derp

PR #1007, opened 2018-12-01
Added Debug.logTime function for easier performance debugging

albertdahlin:
This lets you "tag" any function for easy performance debugging. More info in [this discourse thread](https://discourse.elm-lang.org/t/very-useful-function-when-optimizing-render/1517).

Wrap any function in your code:
image

This will log the execution time in the console:
image

It is also visible in the performance timeline:
image

### DIFF
--- a/src/Debug.elm
+++ b/src/Debug.elm
@@ -2,13 +2,14 @@ module Debug exposing
   ( toString
   , log
   , todo
+  , logTime
   )
 
 {-| This module can be useful while _developing_ an application. It is not
 available for use in packages or production.
 
 # Debugging
-@docs toString, log, todo
+@docs toString, log, todo, logTime
 -}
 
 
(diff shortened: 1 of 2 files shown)


albertdahlin:
I was thinking about renaming this function to `Debug.benchmark` instead, is that better?

Tails:
Considering that this addition will take a long time to be included in an official package; how do you build this core package and include it in your environment?

albertdahlin:
The way I use this locally is
- Replace the `Debug.elm` module somewhere in `~/.elm/packages` or wherever they are located on your system (I use Linux)
- Remove the `*.dat` files in the core module (found in the parent dir to where Debug.elm is located)
- Remove `elm-stuff` in my project.

Then you should be able to use `Debug.logTime` in your project. Probably there is a better way to do this but it works on my computer ;-)

Tails:
It compiles! That's excellent. Lifesaver!

So someone should at some point be making an "illegal" package manager for Elm that applies and unapplies core patches such as your excellent one.

PR #1024, opened 2019-03-27
Fix small error in Task docs

mainlymortal:
I was trying to set focus on an element and found some relevant docs; however, I noticed that there was a mistake/typo with a referenced type.


mainlymortal:
I've moved on.

PR #1026, opened 2019-04-23
Improve performance of List.append

jhrcek:
I did some [benchmarks](https://github.com/jhrcek/faster-list-concat) and it turns out that thanks to this improvement the implementation of `List.concat` about 2x (in Google Chrome) to 3x (in Firefox) faster than the current implementation.

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -363,14 +363,8 @@ product numbers =
 You can also use [the `(++)` operator](Basics#++) to append lists.
 -}
 append : List a -> List a -> List a
-append xs ys =
-  case ys of
-    [] ->
-      xs
-
-    _ ->
-      foldr cons ys xs
-
+append =
+    (++)
 
 {-| Concatenate a bunch of lists into a single list:
 


src/List.elm
```diff
@@ -378,7 +378,7 @@ append xs ys =
 -}
 concat : List (List a) -> List a
 concat lists =
-  foldr append [] lists
+  foldr (++) [] lists
```
turboMaCk: I would rather suggest making `append` an alias for `(++)` so the performance is consistent

jhrcek: `++` is currently defined in terms of `append`: https://github.com/elm/core/blob/master/src/Basics.elm#L81 so I can't do that without further changes to the codebase.

turboMaCk: I think that ` Elm.Kernel.Utils.append` and `List.append` are two independent things but I haven't check it to be 100% sure.

jhrcek: Oh, you're right. Let's wait for opinion from some of the maintainers before doing larger changes.

turboMaCk: this will have a similar fate as https://github.com/elm/core/pull/631

jhrcek: I modified the PR to use `Basics.++` in `List.append` instead of in `List.concat`

Skinney: I’ll take a look at this during the weekend.

Skinney:
So `++` is faster than `List.append` because `++` compiles down to a native javascript function which first checks if the arguments are strings (`++` works on both strings and lists) and then performs the correct appending operation depending on what is being appended.

In theory, this should mean that `++` is slower than `List.append` which doesn't need to perform reflection to figure out what types its arguments are. However, `++` uses a low-level JS code to perform the List appending, which is much faster than the written-in-elm implementation of `List.append` since JS doesn't need to ensure immutability at each step.

This PR proposes to increase performance of `List.append` by aliasing it to `++`. While this does improve performance, it's not the best solution, as the `typeof "string"` check that `++` performs is unnecessary, and makes batch operations like `List.concat` slower than it needs to be.

There are several different ways to solve this, all of which requires careful benchmarking and perhaps some design discussion with Evan to make sure the actual fix has long life (it would be bad to implement a fix which is negated by future code generation or language changes).

I'm planning on taking a closer look at the implementation of List functions during the next week or two, and I'll be sure to figure out the best way to improve the performance of `List.append` in the process of doing that.

If I reach the conclusion that this PR is as good a fix as any other, I'll be sure to say so and recommend the merging of this PR (credit were credit is due). I do, however, believe there is a better fix to this and I'm planning on writing that up *soon*.

Skinney:
Fixed in #1027

jhrcek:
Closing in favor of https://github.com/elm/core/pull/1027

PR #1027, opened 2019-05-03
List module performance improvements

robinheghan:
This PR improves performance of the following functions:

* append
* filterMap
* indexedMap
* concatMap
* concat
* foldr (for large lists)

The improvements range from 20% to 100% depending on the browser and function.

More specifically, this is what has been done:

* indexedMapped, concatMap and filterMap has been re-implemented.
* append is now implemented directly in javascript, and the (++) operator now yields to this function. It's now expected that List.append is faster than (++), while appending via the operator is as fast as before.
* foldrs threshold for reverting to a slower, but stack safe, implementation is now doubled to match the threshold of take (which has a similar implementation).

### DIFF
--- a/src/Elm/Kernel/List.js
+++ b/src/Elm/Kernel/List.js
@@ -70,6 +70,23 @@ var _List_map5 = F6(function(f, vs, ws, xs, ys, zs)
 	return _List_fromArray(arr);
 });
 
+var _List_append = F2(_List_ap);
+
+function _List_ap(xs, ys)
+{
+	if (!xs.b)
+	{
+		return ys;
+	}
+	var root = _List_Cons(xs.a, ys);
+	xs = xs.b
+	for (var curr = root; xs.b; xs = xs.b) // WHILE_CONS
+	{
+		curr = curr.b = _List_Cons(xs.a, ys);
+	}
+	return root;
+}
+
 var _List_sortBy = F2(function(f, xs)
 {
 	return _List_fromArray(_List_toArray(xs).sort(function(a, b) {
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -179,5 +179,5 @@ function _Utils_ap(xs, ys)
 	}
 
         // append Lists
-        __List_ap(xs, ys);
+        return __List_ap(xs, ys);
 }
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -4,7 +4,7 @@ import Array exposing (toList)
 import Basics exposing (LT, EQ, GT)
 import Dict exposing (toList)
 import Elm.Kernel.Debug exposing (crash)
-import Elm.Kernel.List exposing (Cons, Nil)
+import Elm.Kernel.List exposing (Cons, Nil, ap)
 import Set exposing (toList)
 
 */
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -178,16 +178,6 @@ function _Utils_ap(xs, ys)
 		return xs + ys;
 	}
 
-	// append Lists
-	if (!xs.b)
-	{
-		return ys;
-	}
-	var root = __List_Cons(xs.a, ys);
-	xs = xs.b
-	for (var curr = root; xs.b; xs = xs.b) // WHILE_CONS
-	{
-		curr = curr.b = __List_Cons(xs.a, ys);
-	}
-	return root;
+        // append Lists
+        __List_ap(xs, ys);
 }
--- a/src/List.elm
+++ b/src/List.elm
@@ -132,10 +132,10 @@ element (starting at zero).
 -}
 indexedMap : (Int -> a -> b) -> List a -> List b
 indexedMap f xs =
-  reverse (indexedMapHelper 0 f xs [])
+  reverse (indexedMapHelper f 0 xs [])
 
 
-indexedMapHelper :  (Int -> a -> b)->Int -> List a -> List b -> List b
+indexedMapHelper :  (Int -> a -> b) -> Int -> List a -> List b -> List b
 indexedMapHelper fn index list result =
   case list of
     [] ->
--- a/src/List.elm
+++ b/src/List.elm
@@ -372,13 +372,8 @@ product numbers =
 You can also use [the `(++)` operator](Basics#++) to append lists.
 -}
 append : List a -> List a -> List a
-append xs ys =
-  case ys of
-    [] ->
-      xs
-
-    _ ->
-      foldr cons ys xs
+append =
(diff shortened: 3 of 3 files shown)


src/Elm/Kernel/Utils.js
```diff
@@ -178,16 +178,6 @@ function _Utils_ap(xs, ys)
...
-	}
-	var root = __List_Cons(xs.a, ys);
-	xs = xs.b
-	for (var curr = root; xs.b; xs = xs.b) // WHILE_CONS
-	{
-		curr = curr.b = __List_Cons(xs.a, ys);
-	}
-	return root;
+        // append Lists
+        __List_ap(xs, ys);
```
harrysarson: Unless I am missing soemthing  (which I may well be), a return is needed here otherwise this is a nip. Unless this function appends in place? 
```suggestion
        return __List_ap(xs, ys);
```

Skinney:
Thank you, @harrysarson.

src/List.elm
```diff
@@ -132,7 +132,17 @@ element (starting at zero).
 -}
 indexedMap : (Int -> a -> b) -> List a -> List b
 indexedMap f xs =
-  map2 f (range 0 (length xs - 1)) xs
+  reverse (indexedMapHelper 0 f xs [])
+
+
+indexedMapHelper :  (Int -> a -> b)->Int -> List a -> List b -> List b
```
turboMaCk: nitpick - missing spaces around `->`

src/List.elm
```diff
@@ -198,7 +208,7 @@ foldrHelper fn acc ctr ls =
                                 d :: r4 ->
                                     let
                                         res =
-                                            if ctr > 500 then
+                                            if ctr > 1000 then
```
turboMaCk: I wonder why did you change this magic number but I assume some measurements suggested to do so.

src/Elm/Kernel/List.js
```diff
@@ -70,6 +70,23 @@ var _List_map5 = F6(function(f, vs, ws, xs, ys, zs)
 	return _List_fromArray(arr);
 });
 
+var _List_append = F2(_List_ap);
+
+function _List_ap(xs, ys)
```
turboMaCk: nitpick - I wonder about naming here. For instance cons uses this convention `var _List_cons = F2(_List_Cons);` (note same naming but different cases) https://github.com/elm/core/blob/4fbcb15dfcdb3038030a744b77c230b1e3665036/src/Elm/Kernel/List.js#L16

turboMaCk (commented):
thanks for looking into this

harrysarson:
I tried running the test suite (using @jerith666's #1017) for this pull and got:

```
elm: Bad kernel symbol: List_ap
CallStack (from HasCallStack):
  error, called at compiler/src/Elm/Kernel.hs:188:19 in main:Elm.Kernel
elm: thread blocked indefinitely in an MVar operation
Compilation failed while attempting to build /home/travis/build/harrysarson/core/tests/tests/Main.elm
```

This may be a quirk of the hacks used to run the test suite but I thought it worth flagging in case there is an actual problem here.

turboMaCk:
@harrysarson try to `rm --rf elm-stuff` and start over. I just tried it as well and all tests passed for me without and error. I think this error is unrelated to the change (MVar is a Haskell concurrency primitive used in the compiler and the error indicates dead lock in the compiler). I'm using NixOS on unstable-branch and manual compilation of elmi-to-json and elm-test (both from master)

turboMaCk:
@harrysarson sorry for the miss information! when I use `run-tests.sh` I also get the same error and even after cleaning `elm-stuff`.

src/List.elm
Skinney: It’s to match the similar number for take. The number should be as big as possible, without risking that people run into stack overflows. Since take has had the number at 1000 for a long while now, and people don’t complain about stack overflows, it seemed sensible to increase it for foldr as well.

src/Elm/Kernel/List.js
Skinney: `_List_Cons` is a constructor (the function returns a type). `_List_cons` is a function, which calls the constructor. You’ll see it defined in the `List` module as a function.

harrysarson:
@turboMaCk I have no confidence that `run-tests.sh` is not causing this error by the weird things it has to do. How did you run the tests?

turboMaCk:
@harrysarson My first attempt was just to run `elm-test` but later I realized this might not running expected thing for elm-core as there might be some hacks needed to really run tests against the code in the repository. I haven't investigated deeper but my feeling now is that it's a legit issue.

My **guess** is that it's this call https://github.com/elm/core/pull/1027/files#diff-99e42729736fe9132da4bba9d117b0e0R182 to `__List_ap` defined in `List.js` used in `Utils.js` which is problematic and that probably this function needs to be imported somehow here https://github.com/elm/core/blob/c9975bf73eefdaa5b7d5b41d1bd1b15dec854d41/src/Elm/Kernel/Utils.js#L7 but I think @Skinney would know those things better.

Skinney:
I actually know very little about how these things work. I doubt anyone but Evan knows.

harrysarson:
Now #1017 has landed, if you re run the CI it should go green. (Best way to re run CI is to open and  close this pull request)

src/List.elm
```diff
@@ -387,7 +391,12 @@ concat lists =
 -}
 concatMap : (a -> List b) -> List a -> List b
 concatMap f list =
-  concat (map f list)
+  let
+    helper val acc =
+      append (f val) acc
+  in
+    foldr helper [] list
```
Janiczek, line 398: @sparksp figured out a faster `concatMap`:
```elm
fasterConcatMap : (a -> List b) -> List a -> List b
fasterConcatMap f =
    List.foldr (f >> (++)) []
```
Granted, this doesn't take the JS changes into account, so maybe overall your version is faster, but in current `elm/core` the above snippet wins the benchmark.

sparksp: Luckily this is the same as the proposed concatMap 😄 `(f >> (++))` is functionally the same as `helper` just less expressive.

Janiczek: beware: `append` vs `(++)` does introduce a difference, IIRC.

Skinney: The JS changes in this PR makes `++` and `List.append` equally fast. So if this PR was to be merged, I don't think there would be a difference between these two implementations.

turboMaCk: exactly. The point of this PR is to make the non-obvious performance difference between `List.append` and `(++)` to go away which will implicitly improve all the other functions defined in terms of `append`.

I wonder if you have any idea how @evancz feels about the change @Skinney. It's been a while since this was opened.

Skinney: I remember that me and Evan looked through this, but I forget what the conclusion of that meeting was. I don't think he was opposed to any of the changes, though.

PR #1031, opened 2019-05-27
Fixes runtime exception when comparing inequal lists with over 100 elements

Skinney:
This has the same solution as #1018, but uses fuzz testing instead of unit tests. This PR is also not conflated with upgrading the testing code for 0.19.

### DIFF
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -31,16 +31,16 @@ function _Utils_eqHelp(x, y, depth, stack)
 		return true;
 	}
 
-	if (depth > 100)
-	{
-		stack.push(_Utils_Tuple2(x,y));
-		return true;
-	}
-
 	if (typeof x !== 'object' || x === null || y === null)
 	{
 		typeof x === 'function' && __Debug_crash(5);
 		return false;
+	}
+
+	if (depth > 100)
+	{
+		stack.push(_Utils_Tuple2(x,y));
+		return true;
 	}
 
 	/**__DEBUG/
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -26,14 +26,14 @@ function _Utils_eq(x, y)
 
 function _Utils_eqHelp(x, y, depth, stack)
 {
-	if (depth > 100)
+	if (x === y)
 	{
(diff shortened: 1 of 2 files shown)


jerith666:
Sorry for the long delay in responding here, but I finally had a chance to run the tests at this commit (with the 0.19.1 updates from #1059), and they don't pass for me:

```
elm-test 0.19.1
---------------

Running 1268 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 193469607216503 /home/matt/git/elm/core/tests/tests/Main.elm

↓ Test.Equality
↓ Equality Tests
↓ List equality
✗ Simple comparison

    This test failed because it threw an exception: "TypeError: Cannot read property '$' of undefined"



TEST RUN FAILED

Duration: 65658 ms
Passed:   1267
Failed:   1
```

jerith666:
I also tried running in the tests added by your 3520419 commit with the fix from my 6eff09b commit, and the tests pass.

jerith666:
Finally, I also tried running the tests added by my 6eff09b against the fix from your 3520419; the tests also fail in that case.

jerith666:
I think what this all means is that it's the `typeof x !== 'object' || x === null || y === null` check that prevents `Tuple2`s containing `undefined` from being pushed onto the stack, *not* the `x === y` check.

Skinney:
Hmm. So the `typeof` check should be moved up to before the `depth` check?

jerith666:
I think so, yes.  My fix moves both the `x === y` and the `typeof ...` checks above the `depth` check.

I should say that I don't have any deep knowledge of how Elm objects get represented as JS objects that's informing me here, I'm just going by what I saw in the browser debugger when I was exploring the original failures from the SSCCE I posted in #1011.  That's maybe why I got a bit more elaborate in my test cases than you, as well?

I'm happy to open a new PR with my fix rebased onto the current tip of `master` if that helps.

jerith666:
Yep, that leaves `Utils.js` identical to what I had in #1018, so it should work!

harrysarson:
Should we add the extra tests to prevent regressions?

jerith666:
I'm always a fan of more tests -- esp. for something as critical as `elm/core`!  :)

I can make a PR with the combined set of tests if you'd like.

jerith666:
My new PR is #1062

PR #1033, opened 2019-06-04
Faster insert, remove and update for Dict

robinheghan:
This PR introduces a re-write of the remove algorithm, speeding it up considerably. This new implementation has also allowed us to inline the balancing steps of insertion, speeding up inserts as well. Finally, there is now a specialized implementation of `update`, merging the code of `insert` and `remove` for great benefit.

According to my benchmarks, on the latest version of Safari on Mac:
Insertion is 20% faster
Remove is 140% faster
Update matches the increase of insert and remove.

### DIFF
--- a/src/Dict.elm
+++ b/src/Dict.elm
@@ -1,49 +1,60 @@
 module Dict exposing
-  ( Dict
-  , empty, singleton, insert, update, remove
-  , isEmpty, member, get, size
-  , keys, values, toList, fromList
-  , map, foldl, foldr, filter, partition
-  , union, intersect, diff, merge
-  )
+    ( Dict
+    , empty, singleton, insert, update, remove
+    , isEmpty, member, get, size
+    , keys, values, toList, fromList
+    , map, foldl, foldr, filter, partition
+    , union, intersect, diff, merge
+    )
 
 {-| A dictionary mapping unique keys to values. The keys can be any comparable
 type. This includes `Int`, `Float`, `Time`, `Char`, `String`, and tuples or
 lists of comparable types.
 
-Insert, remove, and query operations all take *O(log n)* time.
+Insert, remove, and query operations all take _O(log n)_ time.
+
 
 # Dictionaries
+
 @docs Dict
 
+
 # Build
+
 @docs empty, singleton, insert, update, remove
 
+
 # Query
+
 @docs isEmpty, member, get, size
 
+
 # Lists
+
 @docs keys, values, toList, fromList
 
+
 # Transform
+
 @docs map, foldl, foldr, filter, partition
 
+
 # Combine
+
 @docs union, intersect, diff, merge
 
 -}
 
-
 import Basics exposing (..)
-import Maybe exposing (..)
 import List exposing (..)
+import Maybe exposing (..)
 
 
 
 -- DICTIONARIES
-
-
 -- The color of a node. Leaves are considered Black.
+
+
 type NColor
     = Red
     | Black
@@ -57,27 +68,39 @@ that lets you look up a `String` (such as user names) and find the associated
 
     users : Dict String User
     users =
-      Dict.fromList
-        [ ("Alice", User "Alice" 28 1.65)
-        , ("Bob"  , User "Bob"   19 1.82)
-        , ("Chuck", User "Chuck" 33 1.75)
-        ]
+        Dict.fromList
+            [ ( "Alice", User "Alice" 28 1.65 )
+            , ( "Bob", User "Bob" 19 1.82 )
+            , ( "Chuck", User "Chuck" 33 1.75 )
+            ]
 
     type alias User =
-      { name : String
-      , age : Int
-      , height : Float
-      }
+        { name : String
+        , age : Int
+        , height : Float
+        }
+
 -}
 type Dict k v
     = RBNode_elm_builtin NColor k v (Dict k v) (Dict k v)
     | RBEmpty_elm_builtin
+      -- Temporary state used when removing elements
+    | RBBlackMissing_elm_builtin (Dict k v)
 
 
-{-| Create an empty dictionary. -}
+{-| Create an empty dictionary.
+-}
 empty : Dict k v
 empty =
-  RBEmpty_elm_builtin
+    RBEmpty_elm_builtin
+
+
+{-| Create a dictionary with one key-value pair.
+-}
+singleton : comparable -> v -> Dict comparable v
+singleton key value =
+    -- Root node is always Black
+    RBNode_elm_builtin Black key value RBEmpty_elm_builtin RBEmpty_elm_builtin
 
 
 {-| Get the value associated with a key. If the key is not found, return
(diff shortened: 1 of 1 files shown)


showell:
I am trying to read `insertHelp`..  The overall structure of `LT`, `EQ`, and `GT` is obviously quite clear.  And of course the `EQ` case requires no explanation.  But I don't understand why `LT` and `GT` are structured so differently.

For the `LT` case you get a subtree with top, left, and right nodes, which I'll notate as (top, left, right).  The eight scenarios are:

- (Red, Red, Red) : impossible, I'm guessing
- (Red, Red, Black): rotated in your patch
- (Red, Black, Red): not handled?
- (Red, Black, Black): already fine
- (Black, Red, Red): already fine
- (Black, Red, Black): already fine
- (Black, Black, Red): already fine
- (Black, Black, Black): already fine

Sorry for the tedious enumeration of all eight states. :)  The only state I'm confused about is the theoretical (Red, Black, Red) state.  Is this state somehow impossible by virtue of what happens in the `GT` side of the same helper?  Is that the secret sauce that I'm not gleaning somehow?

Skinney:
It's difficult for me to answer this, because the implementation is hard to read because of several optimizations, and because red-black trees in general are tricky.

This implementation has an invariant where no red right child is allowed, and so having a red right child should trigger a rotation, but in some cases we can be certain that this cannot be the case.  Then there are certain operations which have been inlined, and so all rebalancing does not occur in the helper functions.

It would be easier for me to give you a good answer if you were to point to some lines in the code, and ask what would happen if `node == R B R` or something along those lines.

showell:
If (R, B, R) was ever produced by `insertHelper` when `key < nKey`, then you'd fall through to the second case of this code:

```
case insertHelp key value nLeft of
    RBNode_elm_builtin Red lK lV (RBNode_elm_builtin Red llK llV llLeft llRight) lRight ->
        RBNode_elm_builtin Red lK lV (RBNode_elm_builtin Black llK llV llLeft llRight) (RBNode_elm_builtin Black nKey nValue lRight nRight)

    newLeft ->
        RBNode_elm_builtin nColor nKey nValue newLeft nRight
```

But that code only preserves the invariant of black-on-top if `nColor` is `Black`.  But I think you actually can rely that `nColor` is black, because `nColor` comes from an actually valid tree, as opposed to an intermediate subtree from `insertHelper`.

So I think we're good.

showell:
Ok, I wrote some code that brute forces through more than 8! (40,320) `fromList` calls to verify that `Dict` works correctly for n <= 8.  I highly doubt that there's some strange flaw in the code that only manifests for n >= 9, but I'd love to go deeper for 100% assurance.   The algorithm doesn't really kick into all its complexity until about n=6, when you start to have nontrivial redhead subtrees

Once you hit 10! cases, my current testing mechanisms start overworking the browser.  I have some ideas on how to make this analysis go deeper without combinatorial explosion (basically, eliminate clearly isomorphic cases from each generation), but that will have to wait until at least tomorrow.

I'll show the results of my analysis first.  Here's a list of the unique tree structures that Skinney's `insert` algorithm produces:

```

==============
2 nodes:

T20 = (B R _)

==============
3 nodes:

T30 = (B B B)

==============
4 nodes:

T40 = (B B T20)
T41 = (B T20 B)

==============
5 nodes:

T50 = (B (R B B) B)
T51 = (B T20 T20)

==============
6 nodes:

T60 = (B (R B B) T20)
T61 = (B (R B T20) B)
T62 = (B (R T20 B) B)

==============
7 nodes:

T70 = (B (R B T20) T20)
T71 = (B (R T20 B) T20)
T72 = (B (R T20 T20) B)
T73 = (B T30 T30)

==============
8 nodes:

T80 = (B (R T20 T20) T20)
T81 = (B T30 T40)
T82 = (B T30 T41)
T83 = (B T40 T30)
T84 = (B T41 T30)
```

And below is the code that I wrote to produce the above list.  Note that you need to hack `Dict.elm` to export `Dict(..)` and `NColor(..)`, and then wire up a view to call `show`.  You'll observe that my code tries to enforce some invariants by emitting "wtf" when things go awry.  Nothing went awry.  (I don't enforce everything, though.)

```
dictString : Dict String String -> Dict String v -> String
dictString aliases d =
    let
        c color =
            case color of
                Red ->
                    "R"

                Black ->
                    "B"


        remap s =
            case get s aliases of
                Nothing ->
                    s

                Just t ->
                    t

        helper color left right =
            let
                l =
                    dictString aliases left |> remap

                r =
                    dictString aliases right |> remap
            in
            "(" ++ c color ++ " " ++ l ++ " " ++ r ++ ")"
    in
    case d of
        RBNode_elm_builtin color k v left right ->
            case ( left, right ) of
                ( RBEmpty_elm_builtin, RBEmpty_elm_builtin ) ->
                    c color

                ( RBEmpty_elm_builtin, RBNode_elm_builtin Red _ _ _ _ ) ->
                    -- Red is never a right child
                    "wtf"

                ( RBNode_elm_builtin Red _ _ _ _, _ ) ->
                    if color == Red then
                        "wtf"

                    else
                        helper color left right

                _ ->
                    helper color left right

        RBEmpty_elm_builtin ->
            "_"

        _ ->
            "wtf"


show =
    let
        f aliases lst =
            let
                lst1 =
                    lst |> List.map String.fromInt

                lst2 =
                    lst1 |> List.map (\k -> ( k, "" ))

                dictRepr =
                    fromList lst2
                        |> dictString aliases
            in
            ( lst, dictRepr )

        makeTrees : Dict String String -> Int -> ( String, Dict String String )
        makeTrees aliases n =
            let
                lsts =
                    List.range 1 n
                        |> List.Extra.permutations

                prefix =
                    "T" ++ String.fromInt n

                uniqueStrings =
                    lsts
                        |> List.map (f aliases)
                        |> List.map (\( _, repr ) -> repr)
                        |> Set.fromList
                        |> Set.toList

                makeId idx s =
                    prefix ++ String.fromInt idx

                makeTup idx s =
                    ( s, makeId idx s )

                newAliasesTuples =
                    uniqueStrings
                        |> List.indexedMap makeTup

                updatedAliases =
                    union
                        aliases
                        (fromList newAliasesTuples)

                humanOutput =
                    "\n\n==============\n"
                        ++ String.fromInt n
                        ++ " nodes:\n"
                        ++ (newAliasesTuples
                                |> List.map (\( val, name ) -> "\n" ++ name ++ " = " ++ val)
                                |> List.foldr (++) ""
                           )
            in
            ( humanOutput, updatedAliases )

        startingAliases =
            empty

        accum : ( String, Dict String String ) -> Int -> ( String, Dict String String )
        accum ( humanOutput, aliases ) n =
            let
                ( newHumanOutput, newAliases ) =
                    makeTrees aliases n
            in
            ( humanOutput ++ newHumanOutput, newAliases )

        buildOutRanges lst =
            case lst |> List.reverse of
                [] ->
                    ( "", empty )

                _ ->
                    let
                        n =
                            List.length lst

                        front =
                            List.take (n - 1) lst

                        tail =
                            List.head (List.drop (n - 1) lst)
                                |> Maybe.withDefault 0
                    in
                    accum (buildOutRanges front) tail

        result =
            buildOutRanges (List.range 2 8)
    in
    result
        |> (\( output, _ ) -> output)
        |> (\s -> Html.pre [] [ Html.text s])
```

showell:
It's worth cleaning up most of these errors from `elm-analyse`, I'm guessing:

```  > Unused variable `minKey` inside pattern at ((285,30),(285,36))
  > Unused variable `minValue` inside pattern at ((285,37),(285,45))
  > Unused variable `key` inside pattern at ((295,32),(295,35))
  > Unused variable `value` inside pattern at ((295,36),(295,41))
  > Unused variable `key` inside pattern at ((298,34),(298,37))
  > Unused variable `value` inside pattern at ((298,38),(298,43))
  > Unused variable `v` at ((513,15),(513,16))
  > Unused variable `value` at ((655,17),(655,22))
  > Unused variable `key` at ((663,13),(663,16))
```

showell:
@Skinney I believe this is a failing test for `remove`:

~~~ elm
            Dict.empty
                |> Dict.insert 0 Nothing
                |> Dict.insert 5 Nothing
                |> Dict.insert 6 Nothing
                |> Dict.insert 4 Nothing
                |> Dict.insert 2 Nothing
                |> Dict.insert 3 Nothing
                |> Dict.insert 1 Nothing
                |> Dict.insert 7 Nothing
                |> Dict.remove 0
                |> Dict.remove 5
                |> Dict.remove 6
                |> Dict.remove 4
                |> Dict.remove 2
                |> Dict.remove 3
~~~

It seems to produce this tree:

~~~
[{ color = "B", k = 1, path = "l", v = Nothing },
 { color = "B", k = 7, path = "", v = Nothing },
 { color = "B", k = 7, path = "r", v = Nothing }]

Skinney:
@showell interesting! will take a look :)

showell:
It seems that this leaves a "missing" node after removing the 2 key.  I tried to make a simpler test case, but I think you need to insert all 8 nodes to get the tree into a state that leads to the bug:

~~~
            Dict.empty
                |> Dict.insert 0 Nothing
                |> Dict.insert 5 Nothing
                |> Dict.insert 6 Nothing
                |> Dict.insert 4 Nothing
                |> Dict.insert 2 Nothing
                |> Dict.insert 3 Nothing
                |> Dict.insert 1 Nothing
                |> Dict.insert 7 Nothing
                |> Dict.remove 0
                |> Dict.remove 5
                |> Dict.remove 6
                |> Dict.remove 4
                |> Dict.remove 2

PR #1054, opened 2019-10-23
Use native log10 for logBase with base of 10

### LINKED ISSUE #1048: logBase implementation is insufficiently precise for base 10
https://github.com/elm/core/blob/e2b18dec1a265eddd73c32062df2c3f9bf23885c/tests/tests/Test/Basics.elm#L136

As above, a unit test exists checking `logBase 10 100`, however if this had checked `logBase 10 1000`, it would have caught the fact that `logBase 10 x` implemented as log x / log 10 (using JavaScript's natural logarithm) can yield results slightly *less* than an integer even for small power of 10 inputs.  This unfortunately complicates a common use case for `logBase 10`: finding it's `floor` in order to determine a exponent when formatting a number in scientific notation.

JavaScript's `log10` is provided separately specifically to address this imprecision, and indeed it returns an integer for all powers of 10 ranging from -311 to 308 (i.e., all those representable as normal IEEE 754 binary64 floats and even a few of the subnormals).  Accordingly, I respectfully recommend either adding a corresponding `log10` function in Elm or a special case for `logBase 10` that instead calls JavaScript's `log10`.

richyliu:
Current logBase implementation is not precise enough. For example: `logBase 10 100 == 2` but `logBase 10 1000 /= 3`.

The proposed change is to use a separate native Javascript log10 function for logs with base 10.

This would close #1048

### DIFF
--- a/src/Basics.elm
+++ b/src/Basics.elm
@@ -614,9 +614,12 @@ sqrt =
 -}
 logBase : Float -> Float -> Float
 logBase base number =
-  fdiv
-    (Elm.Kernel.Basics.log number)
-    (Elm.Kernel.Basics.log base)
+  if base == 10 then
+    Elm.Kernel.Basics.log10 number
+  else
+    fdiv
+      (Elm.Kernel.Basics.log number)
+      (Elm.Kernel.Basics.log base)
 
 
 {-| An approximation of e.
--- a/src/Elm/Kernel/Basics.js
+++ b/src/Elm/Kernel/Basics.js
@@ -53,6 +53,7 @@ var _Basics_floor = Math.floor;
 var _Basics_round = Math.round;
 var _Basics_sqrt = Math.sqrt;
 var _Basics_log = Math.log;
+var _Basics_log10 = Math.log10;
 var _Basics_isNaN = isNaN;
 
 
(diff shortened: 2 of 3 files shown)


gampleman:
Would it make sense to special case also `Math.log2` and `Math.log1p`?

rupertlssmith:
This seems to fail to compile, due to using the `==` operator in `Base.elm`, where it is defined:

```
> elm make
Detected problems in 1 module.
-- UNKNOWN OPERATOR --------------------------------------------- src/Basics.elm

I do not recognize the (==) operator.

621|   if base == 10 then
               ^^
Is there an `import` and `exposing` entry for it?
```

It needs to be done like this:


```
if eq base 10 then
```

rupertlssmith:
I would like to add this PR to elm-janitor: https://github.com/elm-janitor/manifesto/blob/master/README.md

Would you be able to fix the compile error to get this patch over the line?

rupertlssmith:
I have another suggestion, which is to extend the test over a larger range of powers of 10. This will test the entire range of a double precision float:

```
        log10Tests =
          describe "Check accuracy of log base 10"
            (List.map
              (\i -> test ("logBase 10 " ++ (10^i |> toString)) <|
                \() -> Expect.equal (toFloat i) (logBase 10 (10^(toFloat i)))
              )
              (List.range -308 308)
            )
```

PR #1055, opened 2019-10-23
Make docs more clear that string is list of utf-16 code points

### LINKED ISSUE #1047: String.length should mention it counts UTF-16 code units
Documentation of `String.length` should mention that it returns length as number of UTF-16 code units.

Or alternatively perhaps the note at `String` type which says "Strings are not lists of characters." should be expanded to say "Strings are not lists of characters, but lists of UTF-16 code points." or something like that.

richyliu:
This would close #1047

### DIFF
--- a/src/String.elm
+++ b/src/String.elm
@@ -14,7 +14,7 @@ module String exposing
 
 {-| A built-in representation for efficient string manipulation. String literals
 are enclosed in `"double quotes"`. Strings are *not* lists of characters, but
-lists of UTF-16 code points.
+lists of UTF-16 code units.
 
 # Strings
 @docs String, isEmpty, length, reverse, repeat, replace
@@ -105,7 +105,7 @@ isEmpty string =
 
 
 {-| Get the length of a string, or more specifically, the number of UTF-16 code
-points
+units
 
     length "innumerable" == 11
     length "" == 0
--- a/src/String.elm
+++ b/src/String.elm
@@ -13,7 +13,8 @@ module String exposing
   )
 
 {-| A built-in representation for efficient string manipulation. String literals
-are enclosed in `"double quotes"`. Strings are *not* lists of characters.
+are enclosed in `"double quotes"`. Strings are *not* lists of characters, but
+lists of UTF-16 code points.
 
 # Strings
 @docs String, isEmpty, length, reverse, repeat, replace
@@ -103,7 +104,8 @@ isEmpty string =
   string == ""
 
 
-{-| Get the length of a string.
+{-| Get the length of a string, or more specifically, the number of UTF-16 code
+points
 
     length "innumerable" == 11
     length "" == 0


showell:
I think the correct terminology is "UTF-16 code units", not "UTF-16 code points." Or you could say "Unicode code points encoded with UTF-16".

src/String.elm
```diff
@@ -81,7 +81,7 @@ import Result exposing (Result)
 
 A `String` can represent any sequence of [unicode characters][u]. You can use
 the unicode escapes from `\u{0000}` to `\u{10FFFF}` to represent characters
-by their code point. You can also include the unicode characters directly.
```
showell: I believe that in this place you actually want to say "code point," since this is specifically referring to a Unicode code point (as opposed to an encoding of the point).

showell:
To get this merged, it probably makes sense to address my most recent comment and then squash your two commits together.

I think we want to be really precise in this documentation, since there are subtle differences between Unicode code points and the encoding of Unicode characters.  It's also possible that we want to underplay the UTF-16 aspect, since that seems like mostly an internal implementation detail.

@richyliu You may find this useful:

https://www.quora.com/In-the-Unicode-standard-what-is-the-difference-between-a-code-unit-and-a-code-point

I'm not super clear on Unicode terminology myself, but I know enough to spot some things that are ambiguous.

It's possible that Evan or somebody else in the core team will just want to revisit the docs holistically here.

I appreciate

richyliu:
Ok, I changed line 84 back to `code point` and squashed the commits

src/String.elm
```diff
@@ -13,7 +13,8 @@ module String exposing
   )
 
 {-| A built-in representation for efficient string manipulation. String literals
-are enclosed in `"double quotes"`. Strings are *not* lists of characters.
+are enclosed in `"double quotes"`. Strings are *not* lists of characters, but
+lists of UTF-16 code units.
```
LiberalArtist, line 17: I think the existing wording ("Strings are *not* lists of characters.") was meant to clarify that a `String` is not `List` (unlike, e.g., [in Haskell](https://hackage.haskell.org/package/base-4.16.1.0/docs/Prelude.html#t:String), where `type String = [Char]`).

I think the reference to UTF-16 is fine (but maybe could be explained in terms of `Char`), but I don't think we should use the word "list". For example, the Revised^6 Report on the Algorithmic Language Scheme—another language where "list" has a very specific meaning—defines the character and string types like this:
> [**Characters**](http://www.r6rs.org/final/html/r6rs/r6rs-Z-H-2.html#node_toc_node_sec_Temp_8)
> Scheme characters mostly correspond to textual characters. More precisely, they are isomorphic to the *scalar values* of the Unicode standard.
> [**Strings**](http://www.r6rs.org/final/html/r6rs/r6rs-Z-H-2.html#node_toc_node_sec_Temp_9)
> Strings are finite sequences of characters with fixed length and thus represent arbitrary Unicode texts.

PR #1056, opened 2019-10-23
Fixes String.uncons returning Nothing on null byte

### LINKED ISSUE #1035: String.uncons returns Nothing whenever there is a null byte in the string's first position
SSCCE:

```elm
String.uncons "\u{0000}"
```

should equal `Just ( '\u{0000}', "" )`, but it equals `Nothing`.

Further, any string starting with the null byte (such as `String.uncons "\u{0000}X"`) also returns `Nothing`.

It looks like this ternary statement is doing this: https://github.com/elm/core/blob/master/src/Elm/Kernel/String.js#L18-L24

`"\u0000".charCodeAt(0)` returns `0`, which means we take the `__Maybe_Nothing` branch.

richyliu:
`String.uncons "\u{0000}"` should equal `Just ( '\u{0000}', "" )`, but it equals `Nothing`.

`string.chartCodeAt(0)` returns `0` on a null byte and `NaN` on an empty string. Instead of coercing to boolean in the ternary, it's better to use `isNaN`.

This would fix #1035

### DIFF
--- a/src/Elm/Kernel/String.js
+++ b/src/Elm/Kernel/String.js
@@ -15,7 +15,7 @@ var _String_cons = F2(function(chr, str)
 function _String_uncons(string)
 {
 	var word = string.charCodeAt(0);
-	return word
+	return !isNaN(word)
 		? __Maybe_Just(
 			0xD800 <= word && word <= 0xDBFF
 				? __Utils_Tuple2(__Utils_chr(string[0] + string[1]), string.slice(2))


evancz:
Ah, I was going through issues and saw #1035 before I saw this one! I did a commit that fixes it as well https://github.com/elm/core/commit/1f85f5edbc7f357fe79673f395beab63e2eb754f. I believe they'll both work just as well. The slow thing here should be allocating a `Just` so I doubt the way of checking on the number is a huge deal in terms of perf. The version I did should compress a bit better with gzip though, so I think it makes sense to go that way as long as it's equally correct!

PR #1065, opened 2019-11-14
Added edge case examples for sum and product

JoshuaHall:
These examples will save a few seconds for people, instead of someone having to lookup what the behavior of these functions is when given an empty List or trying it on `elm repl` (like me 😃 ).

### DIFF
--- a/src/List.elm
+++ b/src/List.elm
@@ -334,7 +334,7 @@ minimum list =
 
 
 {-| Get the sum of the list elements.
-
+    sum [] == 0
     sum [1,2,3,4] == 10
 -}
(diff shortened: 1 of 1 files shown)


evancz:
Thank you for the suggestion! I added these examples (and a few others) in https://github.com/elm/core/commit/fdfa57ec5db388a0cb63242cb3bf0dda1a1e3ae7

I believe there needs to be a totally open newline before the indented code block. Otherwise it will show up as normal text and look weird. So that is why I did it as a separate commit rather than using this PR directly. (The `sum` change was missing a newline I think.)

Issue #1072, opened 2019-12-14
Non Standard behavior on division by zero


CSDUMMI:
There is no answer to the equation `x = a/0` where a ∈ R.
Not even `x = +∞` or `x = - ∞` are valid.
So programming languages have to have a way of dealing with this
mathematical undefined.
Some languages crash (Python 3) others return Infinity (JS)
and Elm does all of it.

There are 4 functions in the Basics Module in elm-core,
that use division in some kind:
1. `(/) : Float -> Float -> Float`
2. `(//) : Int -> Int -> Int`
3. `modBy : Int -> Int -> Int`
4. `remainderBy : Int -> Int -> Int`
There are others, but those that I found used these functions under the hood (degrees for example).
But how do they behave?
```elm
> 1/0
Infinity : Float
> 1//0
0 : Int
> modBy 0 5
Error: Cannot perform mod 0. Division by zero error.
> remainderBy 0 5
NaN : Int
```
What first caught my eye about this was that `modBy 0 5` threw a runtime error,
something Elm isn't supposed to rely upon as anything but last resort.
But more important for me was, that none of these functions returned the same
value, when they had to say "I can't compute this".
Even more problematic is that this behavior isn't explained in the documentation.
And because most of these values are still `number`s and some (0) even reasonable
numbers, programmers mightn't even notice that they are from now on using an "undefined"
value to compute real results.
Because in Elm `1//0 == 0//1 == True`, users of `(//)` will not have a way
of identify wheter they just calculated a zero division, without keeping the divisor.
And for those who use `(/)` or `remainderBy`, they'll have to use `>` and `<`
to identify `Infinity` and `NaN`, because the Elm compiler doesn't compile
a program using these values as functions.

I don't know what to do about this,
whether there should be a value (like NaN in JS) that
is part of the `number` class and can be matched for with `==` or `case .. of`
statements, that signifies "I can't compute this".
Or that all these functions return a Maybe, where Nothing == "I can't compute this"?
What I do know is that this behavior should be known to the programmer and thus
added in a note to the documentation of the functions in question.

I have written an Elm Module, that I will attach, with all the examples I gave in
a separate function.
**It compiles!**

By the way, there have been many issues about this and I will reference some them here:
#909  #1034 #565 #590 #932

CSDUMMI:
[DivisionByZero.elm.txt](https://github.com/elm/core/files/3963890/DivisionByZero.elm.txt)

CSDUMMI:
[...]

PR #1073, opened 2019-12-16
Added documentation to all the affected functions of Issue

CSDUMMI:
In #1072 I explained the behavior and my intention to at least create a documentation
notice for the users of the affected functions, which I am hereby doing.


CSDUMMI:
I have opened this pull request over a month ago and would like to know your thoughts about it.
Why is nobody responding anymore to this or my issue on the subject?
I am now going to implement my more advanced proposal on CSDUMMI/core,
as detailed in #1072

PR #1077, opened 2020-02-10
Add workflows to set expectations about issues and pull requests

supermario:


evancz:
Thank you!

.github/workflows/set-pull-expectations.yml
```diff
@@ -0,0 +1,19 @@
+name: Set Pull Expectations
+on:
+  pull_request:
```
Razzeee: The action doesn't seem to support pull_requests? 
https://github.com/actions/github/blob/master/entrypoint.js

PR #1084, opened 2020-04-23
Add String.split example

dullbananas:
to show that it always returns at least one list item even if the string is empty


lue-bird:
`String.split` could even return a non-empty list if that's the case.

PR #1091, opened 2020-07-26
Improve List.length performance

dullbananas:


harrysarson:
Out of interest: do you have a benchmark?

OvermindDL1:
Yeah I'm curious about a benchmark too, because a closure that doesn't wrap any local variables just compiles to a local function anyway...

dullbananas:
> Out of interest: do you have a benchmark?

I made a benchmark [here](http://dlangs.pythonanywhere.com/s/ectest.html). It tests a list with 50,000 items.

harrysarson:
Wow, I see a huge improvement. Looks like  by removing the `F2` wrappers we get a massive performance gain.

image

```js
var $author$project$Main$currentlength = function (list) {
	return A3(
		$elm$core$List$foldl,
		F2(
			function (_v0, i) {
				return i + 1;
			}),
		0,
		list);
};
var $author$project$Main$lengthHelp0 = F2(
	function (list, acc) {
		lengthHelp0:
		while (true) {
			if (!list.b) {
				return acc;
			} else {
				var xs = list.b;
				var $temp$list = xs,
					$temp$acc = acc + 1;
				list = $temp$list;
				acc = $temp$acc;
				continue lengthHelp0;
			}
		}
	});
var $author$project$Main$newlength = function (list) {
	return A2($author$project$Main$lengthHelp0, list, 0);
};
```

PR #1092, opened 2020-08-27
correctly compare bytes in ==

harrysarson:
`Bytes` equality requires a special case in `_Utils_eqHelp`. This PR adds that special case.

Fixes https://github.com/elm/bytes/issues/15

### DIFF
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -63,6 +63,20 @@ function _Utils_eqHelp(x, y, depth, stack)
 		y = __Dict_toList(y);
 	}
 	//*/
+	
+	if (typeof DataView === "function" && x instanceof DataView) {
+		let length = x.byteLength;
+		
+		if (y.byteLength !== length) {
+			return false;
+		}
+
+		for (let i = 0; i < length; ++i) {
+			if (x.getUint8(i) !== y.getUint8(i)) {
+				return false;
+			}
+		}
+	}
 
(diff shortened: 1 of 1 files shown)


harrysarson:
```
 ✖  fatal     HttpError: Resource not accessible by integration 
    at /node_modules/@octokit/request/lib/request.js:56:27
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Object.doComment [as comment] (/entrypoint.js:47:5)
```

The GH action does not seem to work. Pinging @supermario :)

harrysarson:
CI does pass: https://travis-ci.org/github/elm/core. Not sure why GH does not display it 🤷

src/Elm/Kernel/Utils.js
```diff
@@ -63,6 +63,20 @@ function _Utils_eqHelp(x, y, depth, stack)
 		y = __Dict_toList(y);
 	}
 	//*/
+	
+	if (typeof DataView === "function" && x instanceof DataView) {
+		let length = x.byteLength;
```
lydell, line 68: `let` would cause a syntax error in Internet Explorer, right? So should be `var`? Same thing in the for-loop below.

harrysarson: Yes of course! I will fix this. IE 😠

harrysarson: Taken me 6 months!

PR #1102, opened 2021-04-05
Add forgotten test to the test suite

jfmengels:
I found a suite of tests that were created but not included in the tests to run.

Note: I have not been able to re-run tests, as tests break for a different reason (already on `master`). I tried running with the latest version of `elm` and `elm-test` for 0.19.0 and for 0.19.1 without success.


harrysarson:
Here is the travis failure: https://travis-ci.org/github/elm/core/builds/765919223. Travis also fails on master :worried:

harrysarson:
See #1103

Issue #1103, opened 2021-04-05
Recent updates to the node-test-runner have broken the run-tests.sh script


harrysarson:
To reproduce:

```bash
./tests/run-tests.sh
```

The error is:

```
Failed to run `elm-json solve`:

-- NO VALID PACKAGE VERSION ----------------------------------------------------

Because elm/core does not appear to exist and this project depends on elm/core
at any version, no valid set of package versions could be found.



# clear out the copy of elm-core fetched by the above and replace it
# with the local source code we want to actually test

VERSION_DIR="$(ls ${ELM_HOME}/0.19.1/packages/elm/core/)"
ls: cannot access '/home/harry/repo/elm/core/tests/.elm/0.19.1/packages/elm/core/': No such file or directory
```

Looks like the upgrade to `elm-test@0.19.1-revision5` (where we added elm-json amongst other things) caused this script to fail.

harrysarson:
As the one who released the offending version of elm-test, sorry!

jfmengels:
@harrysarson What is your opinion on what should be done? Is this an issue you are tracking in `node-test-runner`?

harrysarson:
I put a PR up at #1104 which is one way to fix this. @mpizenberg had a good idea on discord for how to keep using node-test-runner to run tests but I got distracted and never looked at it properly.

Given that tests.sh used all sorts of hacks to get this working I am pretty confident that the fix should be here.

mpizenberg:
Yep, here is the fix: https://github.com/elm/core/compare/master...mpizenberg:tests-alternative

Works with both elm-test and elm-test-rs, and I think is simpler than the current setup

harrysarson:
#1104 is now ready (from my end) to land. It is the minimal PR that fixes `./tests/run-tests.sh`. A follow up PR to migrate the CI to github actions (or to fix travis so that the status appears on PRs) would be nice but I decided to keep #1104 as simple as possible in the hope that is can be merged with minimum friction and unblock other PRs here. :)

PR #1104, opened 2021-04-05
fix tests by making tests a package instead of an application

harrysarson:
Fix tests by making the "fake application" in ./tests instead a "fake package" which helps work around the fact that elm-json does not like the fact that does not like the hacks we use to test this repo.

### Old commit message

~the node test runner does not like the hacks we do in
./tests/run-tests.sh. However elm-test-rs does not mind, it seems easier
to switch test runner than to update the hacks for the node test runner.~

~As an added extra we migrate CI from travis to github actions.~


harrysarson:
Should be ready for review now :)

harrysarson:
Marking as draft because we have found a better fix in #1103

harrysarson:
CI passes now: https://travis-ci.org/github/harrysarson/core

tests/run-tests.sh
```diff
@@ -3,68 +3,41 @@
...
-# elm/core hosted on package.elm-lang.org
-"${ELM_TEST}" tests/Main.elm --fuzz=1 > /dev/null || true;
-
-# clear out the copy of elm-core fetched by the above and replace it
-# with the local source code we want to actually test
-
-VERSION_DIR="$(ls ${ELM_HOME}/0.19.1/packages/elm/core/)"
-CORE_PACKAGE_DIR="${ELM_HOME}/0.19.1/packages/elm/core/$VERSION_DIR"
+# Create a link to the git package
+CORE_LINK="${ELM_HOME}/0.19.1/packages/elm/core/1.0.5"
```
mpizenberg: You might want to add a comment here saying that setting this to `1.0.5` is fine as long as the strategy used by the dependency solver in elm-test (here `elm-json solve`) prioritizes installed versions. In such case, it will use the linked `1.0.5` to the git repo even when a new 1.0.6 version of core exists. This of course only holds true as long as the build-time and run-time tests dependencies do not require a version of `elm/core` > `1.0.5`.

Otherwise, we need to bump this `1.0.5` once there is a new release of `elm/core`.

harrysarson: How about this: https://github.com/elm/core/pull/1104/commits/a888bfedfddd7d505808d017e8e6519017c65dfd hopefully this will keep us in sync regardless of the dependency resolution

mpizenberg: I suppose this does the trick in 99% of the cases. The only exception I can think of is for the commit that changes the package version. It will try to use the new version that does not exist in the package server yet and thus will make the elm compiler fail I think. Might be annoying if this shows a red failure for CI on the commit corresponding to new package push.
What do you think?

PS: except if your `elm publish` is faster than your CI but that's a bit playing the daredevil ^^

harrysarson: hmm, this is a problem. Maybe we need to go back to the old approach of running elm-test once to prepopulate elm_home and then simlink

mpizenberg: Well you could also just ask the package server which version of elm-core is the latest.

mpizenberg: **edited**
```sh
curl -L https://package.elm-lang.org/packages/elm/core/releases.json | jq -r 'keys | .[-1]'
# 1.0.5
```

rupertlssmith:
I would like to add this PR to https://github.com/elm-janitor/core.

I think this would be better squashed to a single commit with `git rebase -i`, would it be possible to do that?

harrysarson:
I have given you commit rights on my fork, please feel free to squash these commits on the [harrysarson:fix-tests](https://github.com/harrysarson/core/tree/fix-tests) branch.

PR #1122, opened 2021-11-03
Remove ANSI handling inside Debug.log implementation

jfmengels:
I noticed that `Debug.toString` calls `_Debug_toAnsiString(false, value)`, where `false` corresponds the `anis` parameter, which determines whether the printed value should be ANSI colored or not.

I have looked for other places where `_Debug_toAnsiString` is used in projects from either `elm` and `elm-explorations` organizations, and could not find any. Therefore, my assumption is that the `ansi` parameter is not used, and that it can therefore be removed. I **may** be missing something here though.

The proposed changes here are removing the `ansi` parameter from the `_Debug_toAnsiString` function, and then also removing and simplifying the places where the parameter was used.

There was also `_Debug_toHexDigit` that was not used (I could not find a reference to it in any of the aforementioned organizations).

The pull request can be removed commit by commit if you so like.

### DIFF
--- a/src/Elm/Kernel/Debug.js
+++ b/src/Elm/Kernel/Debug.js
@@ -52,7 +52,6 @@ function _Debug_toString__DEBUG(value)
 
 function _Debug_toStringHelper(value)
 {
-	var ansi = false;
 	if (typeof value === 'function')
 	{
 		return '<function>';
--- a/src/Elm/Kernel/Debug.js
+++ b/src/Elm/Kernel/Debug.js
@@ -75,7 +75,7 @@ function _Debug_toStringHelper(value)
 
 	if (typeof value === 'string')
 	{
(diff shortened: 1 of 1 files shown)


harrysarson:
My guess is there is something special that happens in the elm repl here meaning this function is used after all.

jfmengels:
@harrysarson You're absolutely right. Looks like it is used in https://github.com/elm/compiler/blob/master/compiler/src/Generate/JavaScript.hs#L99 through code like `JsName.fromKernel Name.debug "toAnsiString"`. I did not expect this to be constructed this way, so I learned something. Thanks for the learning opportunity and sorry for the disturbance 😅 

From a quick search, I think `_Debug_toHexDigit` is still not used though, but I might be missing something else again.

PR #1125, opened 2022-02-26
WIP - Implement app.shutdown() function.

### LINKED ISSUE #886: Leaking Event handlers when using multiple elm apps with routing on page
## TL;DR

When having multiple elm apps using `Navigation.program` in an SPA you can't stop an app from listening to url changes even when they aren't currently displayed. This can lead to weird behavior forces workarounds inside the apps and could lead to memory leaks and performance issues.

All workarounds target "ignoring" the incoming events instead of actually unbinding their handles. Having an `app.kill()` function would be the cleanest solution for the issue at hand, but also be beneficial in other unrelated situations.
- no workaround code needed inside the apps.
- more control over the lifetime of the app and avoid possible memory leaks. This could be a biggie. Leaking memory and/or having handlers running in the background is something that might lead to hard-to-debug behavior in single-page-apps.
- `.embed()` would get a complementary destroy function and not feel so lonely anymore :-)

## The problem

We have an app shell architecture, meaning a host app that loads several other apps on our page. This lets us maintain and deploy each app separately. The app shell routes to `/app1`  and App 1 will route to everything after `/app1/...`. Some of the loaded apps are using `Navigation.program`. When the path changes from `/app1` to `/app2`, App 1 still listens to popstate events even though it should be idle.

## Possible workarounds

There are a couple of ways to circumvent this from being a problem.
1. Removing all subscriptions. => Won't work. There is no way to tell the `Navigation.program` to stop listening to this event since the subscription is untouchable and always there https://github.com/elm-lang/navigation/blob/master/src/Navigation.elm#L73.
2. "Turning off" the App1 when routing away (with something like this `update = (\model msg -> if model.isMounted then update model msg else (model, Cmd.none))`) => would work but bloats the model and update function a little bit.
3. Making sure each app only parses routes that are within it's namespace (only parse stuff starting with `/app1`)

All these methods will just elimate the reaction to the event, rather than removing the actual event listener.

## Proposed solutions

I think 2 solutions would solve this problem, of which the second is preferable.
1. Having some way of telling the Navigation.program to stop listening to popstate(would require ports to notify the program to stop/start listening)
2. Actually having an `app.kill()` function which destroys the elm app and unbinds all handlers and maybe cleans up the DOM.

## Also ...
Having `app.kill()` would also be benficial in a different scenario (correct me if I am wrong):
Say I am using `react-elm-components` and unmount and remount my component a couple of times by navigating away and back to it. Every time an elm app is created and possibly attaches handlers to the page. Simply calling `app.kill` in `componentWillUnmount` would solve that issue.

janwirth:
Closes #886.

### DIFF
--- a/src/Elm/Kernel/Platform.js
+++ b/src/Elm/Kernel/Platform.js
@@ -51,7 +51,14 @@ function _Platform_initialize(flagDecoder, args, init, update, subscriptions, st
 
 	_Platform_enqueueEffects(managers, initPair.b, subscriptions(model));
 
-	return ports ? { ports: ports } : {};
+	const shutdown = function() {
+		managers = null
+		model = null
+		stepper = null
+		ports = null
+	}
+
+	return ports ? { ports: ports, shutdown: shutdown } : { shutdown: shutdown };
 }
 
 


ChristophP:
@janwirth great to have some contribution in this direction.
Two questions:
1. How would subscriptions be unbound and event handlers removed? (I see you still have a WIP in the title so maybe my comment is a bit early)
2. I noticed you used `const`, but the rest of the Elm kernel code uses `var`.

janwirth:
1. I am relying on the implementation shared by @supermario, I'm not the author so perhaps he can offer an explanation
2. fixed.

supermario:
The [gist I shared on unmounting](https://gist.github.com/supermario/4c2615806c6c561a16edf5dd7208a759) covers port unsubscribes, which are manual in that context.

If this `shutdown()` is intended to be holistic, it should probably be extended to do the port unsubscribes automatically as well. I could look into this in a few weeks from now if nobody does so till then.

There's a tangential issue here of `unsubscribeAll()` also being missing, i.e. without manually retaining references to all the subscribed handlers, I believe there's no way to clear out all subscribed handlers at once. Though I can't think of any use-cases for that now where one would want to do that but not `shutdown()`.

PR #1135, opened 2023-01-12
src/Elm/Kernel/Utils.js handles Json.Encode.Value object values.

billstclair:
I have an application that stores the `Json.Encode.Value` from which its objects are decoded. When comparing two of these, one encoded raw from the wire, and one encoded from a saved version, `_Utils_eqHelp` crashes, because the wire version has fields that are missing in the decoded version. This pull request fixes that, by noticing that it has a `Value` in its hand, and using `==` right then, instead of descending recursively.

There's an Ellie example for this crash at https://ellie-app.com/kHr9dynry86a1

### DIFF
--- a/src/Elm/Kernel/Utils.js
+++ b/src/Elm/Kernel/Utils.js
@@ -37,6 +37,14 @@ function _Utils_eqHelp(x, y, depth, stack)
 		return false;
 	}
 
+        // Handle Json.Encode.Value
+        if (x.$ === 0) {
+                if (! (y.$ === 0)) {
+                        return false;
+                } else {
+                        return x.a == y.a
+                }
+
 	if (depth > 100)
 	{
 		stack.push(_Utils_Tuple2(x,y));


src/Elm/Kernel/Utils.js
```diff
@@ -37,6 +37,14 @@ function _Utils_eqHelp(x, y, depth, stack)
 		return false;
 	}
 
+        // Handle Json.Encode.Value
+        if (x.$ === 0) {
```
jfmengels, line 41: I think you have a missing closed bracket.

Also, more importantly, this is not a good enough distinction for JSON. A lot of data (custom types notably) has `{ $: 0, a: ... }` (with or without additional fields).

I'm not saying that this problem should not be fixed, but the [docs do say not to compare JSON directly](https://package.elm-lang.org/packages/elm/core/latest/Basics#==), and this is not the way to fix the issue.

Additionally, `x.a == y.a` is not a correct/sufficient check, since it doesn't compare JSON recursively. Meaning that `Json.Encode.string "abc" == Json.Encode.string "abc"` would return True, but `Json.Decode.list Json.Encode.string ["abc"] == Json.Decode.list Json.Encode.string ["abc"]` for instance would return False.

You would need a deep equal function written for JS objects, with all the problems and inconsistencies that they may have (do you compare prototypes, do you compare functions, etc?)

PR #1155, opened 2025-11-12
Add hot reloading and make it possible to stop Elm apps

### LINKED ISSUE #122: Developer experience: Not always clear how to init from JS
I appreciate how much effort and focus the Elm community devotes to clear & thorough documentation. In an effort to continue improving the docs, I'll go over an experience I recently had trying to use `Browser.sandbox` for a SSCCE.

I found that it was not very clear from the `Browser.element` or `Browser.sandbox` docs how to initialize them within JS.

I spent quite a bit of time pouring over the package docs, elm guide, and elm guide examples to no avail.

I eventually happened upon https://guide.elm-lang.org/interop/ , which shows the use of the `node` field, but it's not clear when this field is relevant wrt `sandbox, element, document, application`. I also found https://guide.elm-lang.org/webapps/, but this only shows how to init a `document`.

Potential ways to ameliorate this confusion:
1. Include js snipptes in the `Browser` docs.
2. Include js/html snippets in the guides examples instead of only pure Elm.
3. Include something like the table below as a quick and clear reference in the Browser docs and Elm guide (with help from @Janiczek).
```
+----------------------+-------------------+------------------------+
| entry point js param | uses `node` field | supports `flags` field |
+----------------------+-------------------+------------------------+
| Platform.worker      | no                | yes                    |
+----------------------+-------------------+------------------------+
| Browser.sandbox      | yes               | no                     |
+----------------------+-------------------+------------------------+
| Browser.element      | yes               | yes                    |
+----------------------+-------------------+------------------------+
| Browser.document     | no                | yes                    |
+----------------------+-------------------+------------------------+
| Browser.application  | no                | yes                    |
+----------------------+-------------------+------------------------+
```
4. All of the above
5. None of the above and something better?

Thanks!
Coury

### LINKED ISSUE #886: Leaking Event handlers when using multiple elm apps with routing on page

lydell:
This takes all the learnings from [elm-watch](https://github.com/lydell/elm-watch/) and implements hot reloading directly in Elm. I’m really happy with how it turned out, it’s pretty simple and clean. “Implements hot reloading” means that there’s now a `Elm.hot.reload()` JavaScript function that tools can use. It’s up to the tool to do file watching and stuff like that.

This also makes it possible to shut down an app, which is useful when embedding an Elm app in a web component or React component. Closes https://github.com/elm/core/issues/886 and closes https://github.com/elm/core/pull/1125

Those two are batched together since they touch on the same parts.

See the new javascript-interface.md file for all the details.

Companion PR:s:

- https://github.com/lydell/browser/pull/1
- https://github.com/lydell/virtual-dom/pull/1

Closes https://github.com/elm/browser/issues/122 (since I’ve documented how to init apps).

### DIFF
--- a/javascript-interface.md
+++ b/javascript-interface.md
@@ -1,0 +1,248 @@
+# JavaScript interface
+
+When you compile Elm to JavaScript using `elm make MyModule.elm --output my-file.js`, how do you actually use that JavaScript file?
+
+## Loading the file
+
+### Browser
+
+```html
+<script src="my-file.js"></script>
+```
+
+(Note: The modern `type="module"` attribute or JavaScript `import` syntax won’t work. It has to be a plain script tag like above.)
+
+The script creates a global variable called `Elm` where all the Elm things are:
+
+```js
+var app = Elm.MyModule.init();
+```
+
+If the `Elm` global variable already exists, it’s augmented. This allows you to load multiple Elm compiled JavaScript files on the same page.
+
+### Node.js
+
+`Platform.worker` programs can be used in [Node.js](https://nodejs.org/).
+
+(The other program types can be used in Node.js, too, if you use the [jsdom](https://github.com/jsdom/jsdom) package.)
+
+Node.js has two module systems: The old “CommonJS” system, and the newer “ESM” system.
+
+For CommonJS:
+
+```js
+var { Elm } = require('./my-file.js');
+var app = Elm.MyModule.init();
+```
+
+For ESM:
+
+```js
+import { createRequire } from 'node:module';
+var require = createRequire(import.meta.url);
+var { Elm } = require('./my-file.js');
+var app = Elm.MyModule.init();
+```
+
+Alternative way:
+
+```js
+import fs from 'node:fs'; // or: var fs = require('fs');
+var code = fs.readFileSync('./my-file.js', 'utf-8');
+var f = new Function(code);
+var scope = {};
+f.call(scope);
+var { Elm } = scope;
+var app = Elm.MyModule.init();
+```
+
+## `Elm.MyModule.init()`
+
+If your Elm module is called `MyModule`, you can initialize it like so:
+
+```js
+var app = Elm.MyModule.init({
+	flags: myFlags,
+	node: myDomNode
+});
+```
+
+If the Elm module is `Blog.Home` it’s `Elm.Blog.Home.init()`, and so on.
+
+`init` takes an object with up to two properties:
+
+- `flags`: If your program takes flags (the first parameter of your Elm `init` function), this is where you pass them.
+- `node`: Pass a DOM node to let Elm take over here, if your program requires one.
+
+You can leave out properties that aren’t needed for your program, and the entire object if none of them are needed. Which ones can be used depend on the program type:
+
+| Program               | `flags`  | `node`                   |
+| --------------------- | -------- | ------------------------ |
+| `Html`                | ignored  | required                 |
+| `Browser.sandbox`     | ignored  | required                 |
+| `Browser.element`     | optional | required                 |
+| `Browser.document`    | optional | ignored, always `<body>` |
+| `Browser.application` | optional | ignored, always `<body>` |
+| `Platform.worker`     | optional | ignored, no `view`       |
+
+The return value is an object with a few properties and functions described below. It’s typically called `app`.
+
+## `app.ports`
+
+If you define ports – _and use them_ – they’ll be available at `app.ports`.
+
+For `myOutgoingPort : String -> Cmd msg`, use `app.ports.myOutgoingPort.subscribe(function(string) { console.log(string) })`. There’s also `unsubscribe` – put your listener function in a variable and pass it to `unsubscribe` to stop listening.
+
+For `myIncomingPort : (String -> msg) -> Sub msg`, use `app.ports.myIncomingPort.send("hello")`.
+
+If you don’t use any ports in your program, `app.ports` does not exist. (It’s _not_ an empty object!) If `app.ports` is unexpectedly missing, double check that your port actually ends up being used – all the way from where you call it, until it’s return value reaches `main`.
+
+## `app.stop()`
+
+This function stops your Elm app. This is useful for example when embedding your Elm app in a web component. When the web component unmounts, stop the Elm app.
+
+`view`, `update` and `subscriptions` are stopped immediately, but pending `Cmd`s such as HTTP requests aren’t cancelled. When they finish, the messages they produce are silently ignored. At that point, if you no longer keep a reference to `app` it will be garbage collected.
+
+The return value is the root DOM node for the app (or `null` for `Platform.worker`).
+
+The first thing `app.stop()` does is call `app.detachView()` (unless you already did that yourself), so read on about that function below for more details.
+
+## `app.detachView()`
+
+This is for advanced use cases. You probably want to use `app.stop()`.
+
+`app.detachView()` makes Elm release control over the DOM: It removes all event listeners and returns the root DOM node for the app to you. This essentially turns the app into a `Platform.worker` – `update` and `subscriptions` will still run, and ports can be used. (If the app was a `Platform.worker` from the beginning, `app.detachView()` doesn’t do anything and returns `null`.)
+
+As mentioned above for `app.stop()`, there might be pending `Cmd`s such as HTTP requests when you stop your Elm app. If it’s important for you to handle the messages they produce as they finish, use `app.detachView()` rather than `app.stop()`. `app.stop()` silently ignores the messages, while with `app.detachView()` they arrive to your `update` function as usual, and you can handle them somehow. Just remember that your `view` function won’t run anymore!
+
+What `Cmd`s might be pending that you might want to care about?
+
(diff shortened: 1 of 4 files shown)


javascript-interface.md
```diff
@@ -0,0 +1,248 @@
...
+
+## `app.ports`
+
+If you define ports – _and use them_ – they’ll be available at `app.ports`.
+
+For `myOutgoingPort : String -> Cmd msg`, use `app.ports.myOutgoingPort.subscribe(function(string) { console.log(string) })`. There’s also `unsubscribe` – put your listener function in a variable and pass it to `unsubscribe` to stop listening.
+
+For `myIncomingPort : (String -> msg) -> Sub msg`, use `app.ports.myIncomingPort.send("hello")`.
+
+If you don’t use any ports in your program, `app.ports` does not exist. (It’s _not_ an empty object!) If `app.ports` is unexpectedly missing, double check that your port actually ends up being used – all the way from where you call it, until it’s return value reaches `main`.
```
jfmengels:
```suggestion
If you don’t use any ports in your program, `app.ports` does not exist. (It’s _not_ an empty object!) If `app.ports` is unexpectedly missing, double check that your port actually ends up being used – all the way from where you call it, until its return value reaches `main`.
```

src/Elm/Kernel/Platform.js
```diff
@@ -15,43 +15,177 @@ import Result exposing (isOk)
...
+
+	/**__DEBUG/
+	init.hotReloadData = {
+		__$impl: impl,
+		__$platform_effectManagers: _Platform_effectManagers,
+		__$scheduler_enqueue: __Scheduler_enqueue
+	};
+	//*/
+
+	return init;
```
lydell, line 41: All functions that wrap `_Platform_initialize` (also in elm/browser) have been changed this way:

- Decrease from `F4` to `F3` and move the last argument to an explicit function (called `init`) inside. That’s equivalent – just another way to express the curriedness.
- Attach `.hotReloadData` on that inner function in `__DEBUG` mode.

The inner function is called `init` because that’s the `init` function that ends up at `Elm.Main.init` for example. So for each Elm entrypoint, there’s `.init()` to initialize an app as usual, and `.init.hotReloadData` for use when hot reloading already initialized apps of that entrypoint.

Code that I’ll comment on later removes `.hotReloadData`, so there’s no risk that people start depending on it.

src/Elm/Kernel/Platform.js
```diff
@@ -15,43 +15,177 @@ import Result exposing (isOk)
...
+			__$update: _update,
+			__$subscriptions: _subscriptions
+		}
+	}
+
 	var result = A2(__Json_run, flagDecoder, __Json_wrap(args ? args['flags'] : undefined));
 	__Result_isOk(result) || __Debug_crash(2 /**__DEBUG/, __Json_errorToString(result.a) /**/);
 	var managers = {};
-	var initPair = init(result.a);
+	var initPair = impl.__$init(result.a);
```
lydell, line 65: You’ll see this a lot in the diff: Instead of popping off stuff from from `impl` (like `var init = impl.__$init; init()`), we pass around the whole `impl` object and read from it all the time (`impl.__$init()`).

This allows us to simply mutate the `impl` object later (`impl.__$init = newInit`) to swap in new versions of the functions. That’s the core of how the hot reloading works!

src/Elm/Kernel/Platform.js
```diff
@@ -15,43 +15,177 @@ import Result exposing (isOk)
...
+		stop: stop
+	};
+
+	if (ports)
+	{
+		app.ports = ports;
+	}
+
+	/**__DEBUG/
+	app.hotReload = function(hotReloadData)
```
lydell, line 158: Just like with `.hotReloadData` that I mentioned earlier, code that I’ll comment on later removes this `.hotReload` method on apps, so there’s no risk that people will depend on it.

In other words, in non-production mode we “smuggle” some more data/functions out to `_Platform_export` so it can create `Elm.hot.reload` – that’s the only exposed thing.

src/Elm/Kernel/Platform.js
```diff
@@ -91,10 +225,82 @@ function _Platform_setupEffects(managers, sendToApp)
 		if (manager.__portSetup)
 		{
 			ports = ports || {};
-			ports[key] = manager.__portSetup(key, sendToApp);
```
lydell, line 94: The reason for these changes is preventing a memory leak, that causes stopped apps never to be garbage collected. See the individual commits of this PR for the full story.

src/Elm/Kernel/Platform.js
```diff
@@ -502,20 +709,110 @@ function _Platform_mergeExportsProd(obj, exports)
...
+		{
+			if (name in obj)
+			{
+				__Debug_crash(6, moduleName)
+			}
+			else
+			{
+				obj[name] = _Platform_wrapInit(moduleName, exp);
+				scope['Elm'].hot.__hotReloadData[moduleName] = exp.hotReloadData;
+				delete exp.hotReloadData;
```
lydell, line 769: As mentioned earlier, this is where we remove `.hotReloadData` (from `Elm.Main.init` for example). Instead it is stored at `Elm.hot.__hotReloadData["Main"]` for example.

(Note that `Elm.hot.__hotReloadData` is a _flat_ object. For `Elm.Blog.Home` the hot reload data is stored at `Elm.hot.__hotReloadData["Blog.Home"]`.)

src/Elm/Kernel/Platform.js
```diff
@@ -502,20 +709,110 @@ function _Platform_mergeExportsProd(obj, exports)
...
+	{
+		var app = init(args);
+
+		var hotReload = app.hotReload;
+		delete app.hotReload;
+		var reloadFunction = function ()
+		{
+			hotReload(scope['Elm'].hot.__hotReloadData[moduleName]);
+		};
+		scope['Elm'].hot.__reloadFunctions.push(reloadFunction);
```
lydell, line 803: As mentioned earlier, this is where `.hotReload` is removed from `app.`

Instead we store it in `Elm.hot.__reloadFunctions`.

Also notice how `app.hotReload` is called with `init.hotReloadData`. In essence, this lets a running app get the `impl` from a new version of the same Elm entrypoint and update itself with it.

Also notice how we read from `scope['Elm'].hot` all the time. That’s because in `_Platform_export__DEBUG` we mutate stuff in `scope['Elm'].hot` when we hot reload.

It’s a bit mind-bend:y, but this allows apps initialized from some earlier version of Elm code have access to the latest stuff from the latest hot reload.

src/Elm/Kernel/List.js
```diff
@@ -81,6 +81,6 @@ var _List_sortWith = F2(function(f, xs)
 {
 	return _List_fromArray(_List_toArray(xs).sort(function(a, b) {
 		var ord = A2(f, a, b);
-		return ord === __Basics_EQ ? 0 : ord === __Basics_LT ? -1 : 1;
+		return ord.$ === __Basics_EQ.$ ? 0 : ord.$ === __Basics_LT.$ ? -1 : 1;
```
lydell, line 84: See https://github.com/lydell/elm-watch/pull/128

